PR #5, opened 2024-10-22
Allow bulk setting edits

karthiknadig:


anthonykim1:
theres one hanging unused variable

PR #7, opened 2024-10-23
Handle bulk setting edits

karthiknadig:


src/features/envCommands.ts
```diff
@@ -157,24 +160,29 @@ export async function handlePackagesCommand(
 }
 
 export interface EnvironmentSetResult {
-    workspace?: PythonProject;
+    project?: PythonProject;
     environment: PythonEnvironment;
 }
 
 export async function setEnvironmentCommand(
```
anthonykim1, line 167: is the environment command for function `setEnvironmentCommand` referring to environment specific activation command?

karthiknadig: No. it is referring to VS code command pallet command `Set Environment`

PR #13, opened 2024-11-07
Add pipelines yml

karthiknadig:


build/test_update_ext_version.py
```diff
@@ -0,0 +1,104 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+import json
+
+import freezegun
+import pytest
+import update_ext_version
+
+TEST_DATETIME = "2022-03-14 01:23:45"
```
eleanorjboyd, line 10: what is this for and why is it in 2022?

karthiknadig: This is to test if and when the year changes and the updater needs to block the build because the year version does not match.

PR #15, opened 2024-11-11
Code execution APIs

karthiknadig:


src/api.ts
```diff
@@ -986,52 +994,110 @@ export interface PythonProjectApi {
...
+        cwd: string | Uri,
+        envVars?: { [key: string]: string },
     ): Promise<Terminal>;
+}
+
+export interface PythonTerminalExecutionOptions {
+    cwd: string | Uri;
+    args?: string[];
+
+    show?: boolean;
```
eleanorjboyd, line 1042: What is the default for show if not included?

src/features/execution/runAsTask.ts
```diff
@@ -30,9 +44,8 @@ export async function runAsTask(
         echo: true,
         panel: TaskPanelKind.Shared,
         close: false,
-        showReuseMessage: false,
+        showReuseMessage: true,
     };
 
-    const execution = await executeTask(task);
-    return execution;
+    return executeTask(task);
```
eleanorjboyd, line 50: does the switch off the `await` not cause problems?

karthiknadig: No, since we are not expecting to catch errors with this here there is no need to block on await.

src/api.ts
karthiknadig: missed this, default is not to show. The task will be there, user has to open it themselves.

Issue #22, opened 2024-11-19
Automatically select env created for a particular workspace


cwebster-99:
Creating an environment for a particular workspace does not automatically select it for that workspace

Issue #21, opened 2024-11-19
File watching not enabled for the environment panel


cwebster-99:
Deleting an environment within the UI refreshes automatically but deleting in the file explorer causes an error on refresh / doesn't auto refresh in the environments view
![Image](https://github.com/user-attachments/assets/4e4706a1-1fd1-4731-acba-954018b7ee70)

cwebster-99:
~Testing this on the new vsix and it correctly updates the environment manager list but it still has the deleted environment selected on the Python project fold. I think it should switch it back to the system default~

jk it was too fast and I didnt realize it was actually working

karthiknadig:
It should switch back to system default. I see this on my machine. There could be a bug.

Issue #20, opened 2024-11-19
Select workspace dependencies when creating new env


cwebster-99:
Create an environment for a particular subfolder did not provide a step to load in workspace dependencies from file.

Issue #18, opened 2024-11-19
Environment icons rendering weird


cwebster-99:
Icons are showing inconsistent coloring and size in the environment panel
![Image](https://github.com/user-attachments/assets/72d5b2cd-3344-4e81-88f9-8aa332f05bc9)

PR #23, opened 2024-11-20
Fix bugs with creation and selection of venv

karthiknadig:
Fixes https://github.com/microsoft/vscode-python-environments/issues/22
Fixes https://github.com/microsoft/vscode-python-environments/issues/21
Fixes https://github.com/microsoft/vscode-python-environments/issues/20
Fixes https://github.com/microsoft/vscode-python-environments/issues/18
Fixes https://github.com/microsoft/vscode-python-environments/issues/17


src/common/pickers/packages.ts
```diff
@@ -170,25 +170,30 @@ function getGroupedItems(items: Installable[]): PackageQuickPickItem[] {
     return result;
 }
 
+async function getInstallables(packageManager: InternalPackageManager, environment: PythonEnvironment) {
+    const installable = await packageManager?.getInstallable(environment);
+    if (installable && installable.length > 0) {
```
eleanorjboyd: is this length check right? If there are no installables wouldn't the length be 0 or less?

PR #66, opened 2024-12-05
Ensure venv creation and packages checks for python2 before attempting creation or package operations

karthiknadig:
Closes https://github.com/microsoft/vscode-python-environments/issues/67


src/managers/sysPython/utils.ts
```diff
@@ -313,6 +313,10 @@ export async function installPackages(
     api: PythonEnvironmentApi,
     manager: PackageManager,
 ): Promise<Package[]> {
+    if (environment.version.startsWith('2.')) {
+        throw new Error('Python 2.* is not supported (deprecated)');
```
anthonykim1, line 317: should we also show some kind of warning via UI too?

Issue #95, opened 2024-12-18
Add support for grouping environments


karthiknadig:
Add an optional `group` field to Python Environments. If group is provided, and a Node with the name of the Group under the manager, and the environments that belong to that group will be put under it.

karthiknadig:
@jjjermiah Do you think this would help with better display of environments to your case?

karthiknadig:
Proposal:
![Image](https://github.com/user-attachments/assets/c317a969-dc86-40bb-91ae-76ed04043694)

jjjermiah:
thanks @karthiknadig i think this is exactly what i was thinking about!

PR #101, opened 2024-12-22
Adds group feature environment manager UI

karthiknadig:
Closes https://github.com/microsoft/vscode-python-environments/issues/95

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -144,6 +144,33 @@ export interface PythonEnvironmentId {
 }
 
 /**
+ * Display information for an environment group.
+ */
+export interface EnvironmentGroupInfo {
+    /**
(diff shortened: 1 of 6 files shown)


jjjermiah:
@karthiknadig would it also be possible to add a contextValue for the group nodes? I think it would be valuable to execute commands on a group, i.e opening a manifest `toml` pertaining to that set of environments

karthiknadig:
@jjjermiah The context for the tree item should include the Group name and manager ID, like this `pythonEnvGroup;ms-python.python:venv:mygroup;`. You should be able to use that to set commands for specific groups.

eleanorjboyd (approved):

PR #135, opened 2025-01-23
Add get package copilot tool

eleanorjboyd:
add new copilot tool `pythonGetPackages` that can be called by the user or other agents / tools


package.json
```diff
@@ -444,6 +444,30 @@
...
+                "toolReferenceName": "pythonGetPackages",
+                "tags": [
+                    "vscode_editing"
+                ],
+                "icon": "$(files)",
+                "inputSchema": {
+                    "type": "object",
+                    "properties": {
+                        "filePath": {
+                            "type": "string"
```
connor4312, line 462: A description here is useful for the model

package.json
```diff
@@ -444,6 +444,30 @@
...
+                        }
+                    },
+                    "required": [
+                        "filePath"
+                    ]
+                },
+                "canBeReferencedInPrompt": true
```
connor4312, line 470: Should there be a `when` clause so this is only available when there's Python in the workspace?

package.json
```diff
@@ -444,6 +444,30 @@
             {
                 "type": "python"
             }
+        ],
+        "languageModelTools": [
+            {
+                "name": "python_get_packages",
+                "displayName": "Get Python Packages",
+                "modelDescription": "Given a file path, finds the workspace and selected python environment for this file, and returns the list of installed packages.",
```
connor4312: "finds the workspace and selected python environment for this file" may be unnecessary, and you may want to tell the model when this should be called. Something like, "Returns the packages available to the given Python file. You should call this when you want to generate Python code in order to determine the packages the user prefers to use."

package.json
eleanorjboyd: The extension itself does not activate if there is no python file open so I think its ok for now but if its too greedy I can add another clause.

karthiknadig (approved):

Issue #116, opened 2025-01-16
Use shell type API from core to detect shell type


anthonykim1:
Instead of trying to detect and figure out shell type from extension side: https://github.com/microsoft/vscode-python-environments/blob/main/src/features/common/shellDetector.ts 

We should eventually switch over to using terminal shell type from core. Related issue: https://github.com/microsoft/vscode/issues/230165

flying-sheep:
in #137, I’m adding some type safety to shell type handling.

if we merge that PR before we switch to core APIs, we need to take care to not break its type safe way to make sure all shells are handled in all pieces of the code that handle shells.

anthonykim1:
Hi @flying-sheep I appreciate you letting me know.
Shell type API (proposed) did get merged just last week, so it may be good idea if we start using it, and hopefully make the shell detection more reliable for activation/deactivation. Though I don't think the list of shell types identically match though. 

The goal is I want to lessen the burden so we do less prediction of shell types and environments on extension side of things.
There is also shell environment APIs that are coming up too!

karthiknadig:
Done via https://github.com/microsoft/vscode-python-environments/pull/153

PR #137, opened 2025-01-24
Handle all shells’ de/activation

### LINKED ISSUE #138: Venv activation/deactivation is missing for many shell types and broken for one of the existing ones
missing:

- gitbash
- wsl
- ksh
- tcshell
- xonsh
- nushell

broken for `cshell` (tries to use `bin/activate`, instead of the cshell compatible `bin/activate.csh`)

flying-sheep:
Fixes #138

This adds shell activation for

- gitbash
- wsl
- ksh
- tcshell
- xonsh
- nushell

and fixes shell activation for C shell.

it also makes sure that it’s impossible to add a terminal to the enum without also adding shell activation/deactivation commands by using a `Record` type that needs to be exhaustive.

### DIFF
--- a/src/managers/builtin/venvUtils.ts
+++ b/src/managers/builtin/venvUtils.ts
@@ -130,13 +130,13 @@ async function getPythonInfo(env: NativeEnvInfo): Promise<PythonEnvironmentInfo>
         const cmdMgr = (suffix = ''): VenvManager => ({
             activate: { executable: path.join(binDir, `activate${suffix}`) },
             deactivate: { executable: path.join(binDir, `deactivate${suffix}`) },
-            supportsStdlib: true,
+            supportsStdlib: ['', '.bat'].includes(suffix),
         });
         /** Venv activation/deactivation for a POSIXy shell */
         const sourceMgr = (suffix = '', executable = 'source'): VenvManager => ({
             activate: { executable, args: [path.join(binDir, `activate${suffix}`)] },
             deactivate: { executable: 'deactivate' },
-            supportsStdlib: ['', '.ps1'].includes(suffix)
+            supportsStdlib: ['', '.ps1'].includes(suffix),
         });
         // satisfies `Record` to make sure all shells are covered
         const venvManagers: Record<TerminalShellType, VenvManager> = {
--- a/src/managers/builtin/venvUtils.ts
+++ b/src/managers/builtin/venvUtils.ts
@@ -31,6 +31,7 @@ import {
 } from '../../common/window.apis';
 import { showErrorMessage } from '../../common/errors/utils';
 import { Common, VenvManagerStrings } from '../../common/localize';
+import unsafeEntries from '../../common/utils/unsafeEntries';
 import { isUvInstalled, runUV, runPython } from './helpers';
 import { getWorkspacePackagesToInstall } from './pipUtils';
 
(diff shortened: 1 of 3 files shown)


src/managers/builtin/venvUtils.ts
```diff
@@ -117,44 +118,47 @@ async function getPythonInfo(env: NativeEnvInfo): Promise<PythonEnvironmentInfo>
...
         const shellActivation: Map<TerminalShellType, PythonCommandRunConfiguration[]> = new Map();
         const shellDeactivation: Map<TerminalShellType, PythonCommandRunConfiguration[]> = new Map();
 
-        // Commands for bash
-        shellActivation.set(TerminalShellType.bash, [{ executable: 'source', args: [path.join(binDir, 'activate')] }]);
-        shellDeactivation.set(TerminalShellType.bash, [{ executable: 'deactivate' }]);
-
-        // Commands for csh
-        shellActivation.set(TerminalShellType.cshell, [
-            { executable: 'source', args: [path.join(binDir, 'activate')] },
```
flying-sheep, line 130: that was broken, it’s `activate.csh`

src/managers/builtin/venvUtils.ts
```diff
@@ -117,44 +118,47 @@ async function getPythonInfo(env: NativeEnvInfo): Promise<PythonEnvironmentInfo>
...
-        // Commands for fish
-        if (await fsapi.pathExists(path.join(binDir, 'activate.fish'))) {
-            shellActivation.set(TerminalShellType.fish, [
-                { executable: 'source', args: [path.join(binDir, 'activate.fish')] },
-            ]);
-            shellDeactivation.set(TerminalShellType.fish, [{ executable: 'deactivate' }]);
+        for (const [shell, {activate, deactivate}] of unsafeEntries(venvManagers)) {
+            if (shell === TerminalShellType.fish && activate.args && !await fsapi.pathExists(activate?.args[0])) {
+                continue;
+            }
```
flying-sheep: no clue why this check only existed for the fish shell and not others.

can someone help here? should this check happen for more shells?

karthiknadig: Older Venvs did not have `.fish`? This is to avoid using the `activate.fish` if it doesn’t exist.  May be we should add the check for all cases.

karthiknadig:
Thanks for the PR! Love the new shell command handling.

src/managers/builtin/venvUtils.ts
karthiknadig: Older venv example python 3.8 created using `python3.9 -m venv .venv`, does not have `.fish` or `.nu`
image

Newer venv created using `uv`:
image

flying-sheep: I don’t think your two examples only differ by age. I think `uv` simply uses the `virtualenv` package, which supports more shell.

So I guess the check should be in place for all shells supported by `virtualenv` and not by the builtin `venv` package.

flying-sheep: OK, done!

anthonykim1:
Thanks for doing this @flying-sheep 
I mentioned https://github.com/microsoft/vscode-python-environments/issues/116, we should eventually start using shell type API. 

The list on https://github.com/microsoft/vscode/blob/e8346324cfb8b5d680ad657bb67153eb838aebc5/src/vscode-dts/vscode.proposed.terminalShellType.d.ts are subject to change, so feel free to let me know if anyone thinks more shell types could be added to make the workload easier on the extension side. As we want to make sure to cover extensive list of shell types for their specific activation.

anthonykim1 (approved):

karthiknadig (approved):

PR #160, opened 2025-02-06
Use activation event and shell integration for activation tracking

karthiknadig:


src/features/terminal/terminalActivationState.ts
```diff
@@ -0,0 +1,323 @@
...
+        if (activationCommands) {
+            try {
+                for (const command of activationCommands) {
+                    const execPromise = createDeferred<void>();
+                    const execution = shellIntegration.executeCommand(command.executable, command.args ?? []);
+                    const disposables: Disposable[] = [];
+                    let timer: NodeJS.Timeout | undefined = setTimeout(() => {
+                        execPromise.resolve();
+                        traceError(`Shell execution timed out: ${command.executable} ${command.args?.join(' ')}`);
+                    }, 2000);
```
anthonykim1, line 238: Instead of just 2 seconds, we could do a race between longer timeout (5 second or other time of your choice) and also watch if we get exitCode. 

Checking for exit code also let us see the result of whether we activated/deactivated successfully too.


From the shell integration doc: 
```
		 * // Execute a command in a terminal immediately after being created
		 * const myTerm = window.createTerminal();
		 * window.onDidChangeTerminalShellIntegration(async ({ terminal, shellIntegration }) => {
		 *   if (terminal === myTerm) {
		 *     const execution = shellIntegration.executeCommand('echo "Hello world"');
		 *     window.onDidEndTerminalShellExecution(event => {
		 *     if (event.execution === execution) {
		 *       console.log(`Command exited with code ${event.exitCode}`);
		 *     }
		 *   }
		 * }));
```

src/features/terminal/terminalActivationState.ts
```diff
@@ -0,0 +1,323 @@
...
+        environment: PythonEnvironment,
+    ): Promise<void> {
+        const deactivationCommands = getDeactivationCommand(terminal, environment);
+        if (deactivationCommands) {
+            try {
+                for (const command of deactivationCommands) {
+                    const execPromise = createDeferred<void>();
+                    const execution = shellIntegration.executeCommand(command.executable, command.args ?? []);
+                    const disposables: Disposable[] = [];
+                    let timer: NodeJS.Timeout | undefined = setTimeout(() => {
```
anthonykim1, line 287: same feedback from activation could apply here (using exitCode)

src/features/terminal/utils.ts
```diff
@@ -17,3 +20,72 @@ export function isTaskTerminal(terminal: Terminal): boolean {
...
+        }
+    }
+
+    if (env) {
+        return env;
+    }
+
+    // This is a heuristic approach to attempt to find the environment for this terminal.
+    // This is not guaranteed to work, but is better than nothing.
+    const terminalCwd = terminal ? getTerminalCwd(terminal) : undefined;
```
anthonykim1, line 80: Would it be a good idea to also add check that looks through PATH variable?

I think $PATH could have good hints such as
`
/Users/anthonykim/.vscode-insiders/extensions/ms-python.python-2024.23.2025012801-darwin-arm64/python_files/deactivate/zsh:/Users/anthonykim/Desktop/vscode-python/.venv/bin:/Users/anthonykim/.pyenv/shims:/Users/anthonykim/.pyenv/plugins/pyenv-virtualenv/shims:
`
.venv, pyenv etc..

src/features/terminal/terminalActivationState.ts
```diff
@@ -0,0 +1,323 @@
...
+    private onTerminalShellExecutionEndEmitter = new EventEmitter<TerminalShellExecutionEndEvent>();
+    private onTerminalShellExecutionEnd = this.onTerminalShellExecutionEndEmitter.event;
+
+    private onDidChangeTerminalActivationStateEmitter = new EventEmitter<DidChangeTerminalActivationStateEvent>();
+    onDidChangeTerminalActivationState = this.onDidChangeTerminalActivationStateEmitter.event;
+
+    private onTerminalClosedEmitter = new EventEmitter<Terminal>();
+    private onTerminalClosed = this.onTerminalClosedEmitter.event;
+
+    private activatedTerminals = new Map<Terminal, PythonEnvironment>();
```
anthonykim1, line 54: I like this idea of having a map with Terminal with the corresponding environment.

TODO: We should make sure to handle/test reloading, reconnecting scenarios as well.
Not too sure if activated terminal is guaranteed to stay activated or needs reactivation upon reloading, reconnection, restart vs code, etc .

src/features/terminal/terminalActivationState.ts
```diff
@@ -0,0 +1,323 @@
...
+        } else {
+            this.activateLegacy(terminal, environment);
+        }
+    }
+
+    private async deactivateInternal(terminal: Terminal, environment: PythonEnvironment): Promise<void> {
+        if (terminal.shellIntegration) {
+            await this.deactivateUsingShellIntegration(terminal.shellIntegration, terminal, environment);
+        } else {
+            this.deactivateLegacy(terminal, environment);
```
anthonykim1, line 195: I think adding some logs of whether it deactivated using shell integration vs. legacy would be good for our debugging.

src/extension.ts
```diff
@@ -221,9 +220,25 @@ export async function activate(context: ExtensionContext): Promise<PythonEnviron
             traceInfo(
                 `Internal: Changed environment from ${e.old?.displayName} to ${e.new?.displayName} for: ${location}`,
             );
-            await updateActivateMenuButtonContext(terminalManager, projectManager, envManagers);
             updateViewsAndStatus(statusBar, workspaceView, managerView, api);
         }),
+        onDidChangeTerminalShellIntegration(async (e) => {
+            const envVar = e.shellIntegration?.env;
+            if (envVar) {
+                if (envVar['VIRTUAL_ENV']) {
```
anthonykim1, line 228: VIRTUAL_ENV would be indicative only for `venv` and `virtualenv`, right?
I assume there are other things we need to look for remaining virtual environments, which may be up to the virtual env providers.

src/features/terminal/utils.ts
karthiknadig: The PATH will not tell us which exact environment to pick if the terminal is not an activated terminal. It can get confusing to extract this just from PATH.

src/features/terminal/terminalActivationState.ts
karthiknadig: The `env` feature in terminal shell integration will tell us if the terminal is activated or not. So there is no need to track across reloads, reconnection, restart. This tracking is so that we can run the right `deactivate` command, and also set contexts when needed.

src/features/terminal/terminalActivationState.ts
karthiknadig: The `deactivate` function has the logs for this already.

src/extension.ts
karthiknadig: That is correct, we plan on support `venv` built-in, we will likely expose an API for env managers to provide this data.

src/features/terminal/terminalActivationState.ts
karthiknadig: I don't think 5 seconds is a reasonable wait time. I actually don't think 2 seconds is a reasonable time for this either. But currently Shell Integration is quite slow on the default powershell. So I decided to keep it at 2. Typically, this is only called once per terminal.

Not sure what you mean by watching `exitCode`. The execution end event is triggered after exit, so I am using that as an indicator. If you have a different approach we can discuss it.

src/features/terminal/terminalActivationState.ts
karthiknadig: see above https://github.com/microsoft/vscode-python-environments/pull/160#discussion_r1946760609

src/features/terminal/terminalActivationState.ts
anthonykim1: > I don't think 5 seconds is a reasonable wait time. I actually don't think 2 seconds is a reasonable time for this either. But currently Shell Integration is quite slow on the default powershell. So I decided to keep it at 2. Typically, this is only called once per terminal.

I see, yeah.

> Not sure what you mean by watching exitCode. The execution end event is triggered after exit, so I am using that as an indicator. If you have a different approach we can discuss it.

I was mainly thinking of a Promise race where we have timeout of 2 seconds and a listener for `.onDidEndTerminalShellExecution.`

And if we see that `.onDidEndTerminalShellExecution` ran with `execution.exitCode ` other than undefined, we can resolve faster than 2 seconds.

anthonykim1 (approved):

karthiknadig:
@anthonykim1 What you suggested is a great idea, can you create an issue for it so we can do this. Feel free to contribute it if you feel up for it.

Issue #169, opened 2025-02-07
Leverage execution.exitCode for faster execution promise resolve


anthonykim1:
https://github.com/microsoft/vscode-python-environments/pull/160#discussion_r1946083357

We have 
```
                    let timer: NodeJS.Timeout | undefined = setTimeout(() => {
                        execPromise.resolve();
                        traceError(`Shell execution timed out: ${command.executable} ${command.args?.join(' ')}`);
                    }, 2000);
``` 
in src/features/terminal/terminalActivationState.ts

We could do a Promise race where we resolve faster, via listening to `.onDidEndTerminalShellExecution` and watching if exitCode returned with values other than `undefined`

PR #171, opened 2025-02-07
Use exit code for activation, deactivation executeCommand

anthonykim1:
Resolves: #169

### DIFF
--- a/src/features/terminal/terminalActivationState.ts
+++ b/src/features/terminal/terminalActivationState.ts
@@ -232,18 +232,21 @@ export class TerminalActivationImpl implements TerminalActivationInternal {
                     const execPromise = createDeferred<void>();
                     const execution = shellIntegration.executeCommand(command.executable, command.args ?? []);
                     const disposables: Disposable[] = [];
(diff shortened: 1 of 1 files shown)


karthiknadig:
I don't think the changes here dispose correctly. There is nothing stopping the timer calling `resolve()` after the first thing calls resolve.

anthonykim1:
For some reason I thought Promise.race provided inherent safety against this, but I was wrong :/
Flag should help so we dont get side effect (wrong trace logs saying shell execution timed out)

karthiknadig:
You don't need a new variable, the `Deferred` type already tracks this.

anthonykim1:
Is https://github.com/microsoft/vscode-python-environments/pull/171/commits/f1dfbd47991c28d1f56fff66b0c7fffc736e2b66 how you do it? @karthiknadig

karthiknadig:
Here is how I wanted to refactor it: https://github.com/microsoft/vscode-python-environments/pull/181/files

Either timer or the end event should cause the other to be disposed correctly. In the #181, I also removed the dependency on deferred, extracted common execution code into a separate method (reducing duplication), added additional logging. In the future, when we could handle activation failures.

PR #192, opened 2025-02-18
Conform to new shell env API shape 

anthonykim1:
Adding isTrusted field as part of shell env api, meaning the shape would change a bit.
Accessing shell env would be `.env.value` instead of `.env`.

Trying to prevent breaking where this api is already being used, so that I can safely merge: https://github.com/microsoft/vscode/pull/240971

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
     "publisher": "ms-python",
     "preview": true,
     "engines": {
-        "vscode": "^1.97.0   ----> TODO: Pin to insiders after merging new shell env api shape" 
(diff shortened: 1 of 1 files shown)


rzhao271 (dismissed):

anthonykim1:
Ideally I think we would want 
1. the changes in the proposed to be merged in core first
2. pin engine version to tomorrow's insider 
3. merge this in

to prevent any breaking changes.

karthiknadig (approved):

Issue #136, opened 2025-01-23
User Approval Flow for Package Installation


eleanorjboyd:
Since this extension handles package installs, it should also correct communicate to the user about package installs and get user consent. The flow around package installs, user approval, when to allow background installs etc should be discussed and a plan outlined.

## Security:
This is an important part of the user trust boundary as we should only have extensions install packages that have user consent. There could be different categories, trusted extensions that are allowed to install packages and untrusted extensions that need user consent to install. 

## Scenarios:
- extension wants to install a package, it is *not* a pre-approved extension to install packages
- extension wants to install a package, it is a pre-approved extension to install packages
- user wants to reduce notifications and select to have packages installed in the background for a given extension
- user wants to add, revoke, or edit which extensions can install / remove packages in a given workspace
- user wants to add, revoke, or edit which extensions can install / remove packages at a user or more global level
- user wants to accept installing a package without consenting to approve this extension going forward

## Questions:
- How do we make sure users are in control of the package installs while not making the experience too noisy?
- Are there any extensions (like ones published by microsoft) that do not need to be prompted for the user to accept maybe only notified?
- Do we still notify users on package install even if the installer is a pre-approved extension? Do users want to be notified every time?

luabud:
I'm wondering if it'd make sense to have a setting which maps extensions with "approval" status? and by default, no extension is approved until a notification is displayed to the user asking for consent. 
Then in hits notification they would have a choice to approve package(s) installation just this time, always approve package(s) installation for that extension, decline package(s) installation. 

I don't think we should pre-approve any extension, even if it's from Microsoft. I don't think this would be too noisy, since I don't expect package installation to be something _that_ common on a session, especially coming from different providers/extensions 🤔

DonJayamanne:
> I don't think we should pre-approve any extension, even if it's from Microsoft

Agreed, in Jupyter extension we do the same
Data wrangler and synapse extensions execute code against a Jupyter kernel via the Jupyter extension API, here we have no special treatment for msft vs external extensions, same with Jupyter powertoys extension, which is managed by us as well,

eleanorjboyd:
hard to see sorry but began thinking of the flow, you can use[ mermaid.live ](https://mermaid.live/) to play around with it yourself

```
flowchart TD
    A["Install a Package"]

    C1["End: No changes needed"]
    A --> B{"Does the user have a config file?"}
    B -- "No" --> D1{"Create a new config file?"}
[...]

PR #193, opened 2025-02-18
Trust prompting

eleanorjboyd:
fixes https://github.com/microsoft/vscode-python-environments/issues/136

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,7 +87,7 @@
                         "preview"
                     ]
                 },
-                "python-envs.extensionPackageTrust": {
+                "python-envs.allowAutoPackageManagement": {
                     "type": "object",
                     "scope": "user",
                     "properties": {
--- a/package.json
+++ b/package.json
@@ -86,6 +86,21 @@
                         "onExP",
                         "preview"
                     ]
+                },
+                "python-envs.extensionPackageTrust": {
+                    "type": "object",
+                    "scope": "user",
+                    "properties": {
+                        "pub.name": {
+                            "type": "string",
+                            "enum": [
+                                "alwaysAllow",
+                                "alwaysAsk"
+                            ],
+                            "default": "alwaysAsk",
+                            "description": "Sets how installs by a given extension will be handled from the environment extension."
+                        }
+                    }
                 }
             }
         },
--- a/src/features/pythonApi.ts
+++ b/src/features/pythonApi.ts
@@ -46,8 +46,7 @@ import { runAsTask } from './execution/runAsTask';
 import { runInTerminal } from './terminal/runInTerminal';
 import { runInBackground } from './execution/runInBackground';
 import { EnvVarManager } from './execution/envVariableManager';
-import { packageManagementFlow } from './utils';
-
+import { packageManagementFlow } from './packageManagement';
 class PythonEnvironmentApiImpl implements PythonEnvironmentApi {
     private readonly _onDidChangeEnvironments = new EventEmitter<DidChangeEnvironmentsEventArgs>();
     private readonly _onDidChangeEnvironment = new EventEmitter<DidChangeEnvironmentEventArgs>();
--- a/src/features/pythonApi.ts
+++ b/src/features/pythonApi.ts
@@ -1,4 +1,4 @@
-import { Uri, Disposable, Event, EventEmitter, Terminal, TaskExecution, workspace } from 'vscode';
+import { Uri, Disposable, Event, EventEmitter, Terminal, TaskExecution, window } from 'vscode';
 import {
     PythonEnvironmentApi,
     PythonEnvironment,
@@ -48,7 +48,8 @@ import { runInBackground } from './execution/runInBackground';
 import { EnvVarManager } from './execution/envVariableManager';
 import { getCallingExtension } from '../common/utils/frameUtils';
 import { getConfiguration } from '../common/workspace.apis';
-type SettingsPackageTrust = {
+import { promptForInstallPermissions, promptForAlwaysAsk } from './utils';
+export type SettingsPackageTrust = {
     [key: string]: 'alwaysAllow' | 'alwaysAsk';
 };
 class PythonEnvironmentApiImpl implements PythonEnvironmentApi {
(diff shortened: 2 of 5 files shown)


src/features/pythonApi.ts
```diff
@@ -216,11 +221,75 @@ class PythonEnvironmentApiImpl implements PythonEnvironmentApi {
...
+    async installPackages(
+        context: PythonEnvironment,
+        packages: string[],
+        options: PackageInstallOptions,
+    ): Promise<void> {
+        // get the package manager to exit if not found
         const manager = this.envManagers.getPackageManager(context);
         if (!manager) {
             return Promise.reject(new Error('No package manager found'));
         }
```
eleanorjboyd, line 228: @karthiknadig thinking its best to move all of this out to another function / file, still determining the separation point

src/features/pythonApi.ts
```diff
@@ -216,11 +221,75 @@ class PythonEnvironmentApiImpl implements PythonEnvironmentApi {
...
+        traceInfo(`package trust settings for ${callingExtension} is ${callingExtensionTrustLevel}`);
+
+        if (!isConfigured) {
+            // calling extension has no config, user has no wildcard setup
+            // prompt user to "alwaysAsk" or "alwaysAllow"
+            const selectedOption = await promptForInstallPermissions(callingExtension, packages.join(', '));
+            if (selectedOption === 'cancel') {
+                // user cancelled the prompt, exit
+                window.showErrorMessage(`Package installation of ${packages.join(', ')} was canceled by the user.`);
+                return Promise.resolve();
```
eleanorjboyd: It seems like the only two options for what this function returns are resolving it or having it error. When a user cancels the install flow, the error seems to noisy but it is probably necessary to communicate to the calling extension the install didn't happen. Did this come to mind already / thoughts?

src/features/utils.ts
```diff
@@ -0,0 +1,68 @@
+import { window } from 'vscode';
+
+export const ALWAYS_ALLOW = 'Always Allow installs';
+export const ALWAYS_ASK = 'Always Ask before installs';
+export const INSTALL_NO_CONFIGURE = 'Install without configuring';
+
+export const YES_INSTALL = 'Yes, Install';
+export const NO_INSTALL = 'Do Not Install';
+
+export function promptForInstallPermissions(extensionName: string, packages: string): Thenable<string | undefined> {
```
eleanorjboyd: @karthiknadig what is the best way to handle all the string literals between the prompts and the decision cases? Is constants the right way?

package.json
```diff
@@ -86,6 +86,21 @@
                         "onExP",
                         "preview"
                     ]
+                },
+                "python-envs.extensionPackageTrust": {
```
karthiknadig: This is not really trust, it is more of automatic package management. So this should likely say, `allowAutoPackageManagement`

/cc @cwebster-99 @luabud

src/features/pythonApi.ts
karthiknadig: I agree, the install flow checking can move to another file.

src/features/pythonApi.ts
karthiknadig: We should surface the error when a package manager does is not allowed to install. I think it is ok for this error to go to the extension once. This will be useful to surface the UI to allow installs.

src/features/utils.ts
karthiknadig: These should be put in `localize.ts`

src/features/pythonApi.ts
eleanorjboyd: like one of these as the UI?
Screenshot 2025-02-18 at 2 46 06 PM
Screenshot 2025-02-18 at 2 45 40 PM

package.json
eleanorjboyd: switched it to this for now but very flexible on this name - don't have a strong preference at the moment

src/features/utils.ts
eleanorjboyd: sounds good, I will move them over once we determine what they should say 

cc @cwebster-99 @luabud

src/features/utils.ts
```diff
@@ -0,0 +1,68 @@
...
+    return new Promise((resolve) => {
+        window
+            .showInformationMessage(
+                'Select future permissions for package installs from the ' + extensionName + ' extension.',
```
cwebster-99:
```suggestion
                'Would you like to set permissions for future package installs from the ' + extensionName + ' extension?',
```

src/features/utils.ts
```diff
@@ -0,0 +1,68 @@
...
+                }
+            });
+    });
+}
+
+export function promptForAlwaysAsk(extensionName: string, packages: string): Thenable<string | undefined> {
+    return new Promise((resolve) => {
+        window
+            .showInformationMessage(
+                'Do you want to install the following package/s from the ' + extensionName + ' extension?',
```
cwebster-99: Will we be referencing the package names in this message as well?

src/features/utils.ts
```diff
@@ -0,0 +1,68 @@
+import { window } from 'vscode';
+
+export const ALWAYS_ALLOW = 'Always Allow installs';
```
cwebster-99:
```suggestion
export const ALWAYS_ALLOW = 'Always allow installs';
```

src/features/utils.ts
```diff
@@ -0,0 +1,68 @@
...
+export const ALWAYS_ASK = 'Always Ask before installs';
```
cwebster-99:
```suggestion
export const ALWAYS_ASK = 'Always ask before installing';
```

src/features/utils.ts
```diff
@@ -0,0 +1,68 @@
...
+export const INSTALL_NO_CONFIGURE = 'Install without configuring';
```
cwebster-99:
```suggestion
export const INSTALL_NO_CONFIGURE = 'Install without configuring permissions';
```

cwebster-99 (commented):
Left some comments but the options may be too verbose for the UI they are shown in. If that is the case, we can iterate.

src/features/pythonApi.ts
karthiknadig: Yes when they explicitly hit the cancel button. But if it is from a setting then it should just throw exception, and let the caller deal with it.

src/features/utils.ts
karthiknadig: Yes, but if there are multiple then we may want to say something like `pandas, …` or “`pandas` and other packages”.

src/features/utils.ts
karthiknadig: “Confirm before installing “ ?

src/features/utils.ts
eleanorjboyd: I added it in the detail: `package/s: "${packages}"`, section. Is this what you mean? What is the best way to word it?

src/features/pythonApi.ts
```diff
@@ -216,11 +221,75 @@ class PythonEnvironmentApiImpl implements PythonEnvironmentApi {
...
+            }
+            if (selectedOption !== 'installNoConfigure') {
+                // meaning the user selected "alwaysAsk" or "alwaysAllow", update the config
+                const newExtTrustConfig = { ...extPkgTrustConfig, [callingExtension]: selectedOption };
+                config.update('extensionPackageTrust', newExtTrustConfig, true);
+            }
+        } else {
+            // user has already configured package trust settings for this extension
+            if (callingExtensionTrustLevel === 'alwaysAsk') {
+                traceInfo('Package installation is pending user confirmation due to trust settings.');
```
cwebster-99: Should we swap out trust here for consistency? 
```suggestion
                traceInfo('Package installation is pending user confirmation due to permission settings.');
```

src/features/pythonApi.ts
```diff
@@ -216,11 +221,75 @@ class PythonEnvironmentApiImpl implements PythonEnvironmentApi {
...
+                // prompt user to allow or deny package installation
+                const simpleResponse = await promptForAlwaysAsk(callingExtension, packages.join(', '));
+                if (simpleResponse === 'noInstall' || simpleResponse === 'cancel') {
+                    // user cancelled the prompt, exit
+                    window.showErrorMessage(`Package installation of ${packages.join(', ')} was canceled by the user.`);
```
cwebster-99:
```suggestion
                    window.showErrorMessage(`Installation of ${packages.join(', ')} was canceled by the user.`);
```

src/features/utils.ts
cwebster-99: Oh I see! I missed that part in the initial review

src/features/pythonApi.ts
eleanorjboyd: so the full screen modal pop up is the better one between the two?

eleanorjboyd: this is fixed- it will pass the exception to the extension that called the install and it is up to that extension to handle the error as they see fit

package.json
```diff
@@ -86,6 +86,21 @@
...
+                    "scope": "user",
+                    "properties": {
+                        "pub.name": {
+                            "type": "string",
+                            "enum": [
+                                "alwaysAllow",
+                                "alwaysAsk"
+                            ],
+                            "default": "alwaysAsk",
+                            "description": "Sets how installs by a given extension will be handled from the environment extension."
```
cwebster-99, line 101:
```suggestion
                            "description": "Configures package installation permissions for a given extension."
```

PR #196, opened 2025-02-21
Add startup activation

karthiknadig:


src/features/terminal/shells/pwsh/pwshStartup.ts
```diff
@@ -0,0 +1,319 @@
...
+        assert(supportedShells.length > 0, 'At least one PowerShell shell must be supported');
+        this._supportedShells = supportedShells;
+    }
+
+    private async checkInstallations(): Promise<Map<PowerShellType, boolean>> {
+        const results = new Map<PowerShellType, boolean>();
+
+        await Promise.all(
+            this._supportedShells.map(async (shell) => {
+                if (shell === 'pwsh' && this._isPwshInstalled !== undefined) {
```
eleanorjboyd, line 209: why not use the shell constant here?

src/features/terminal/terminalManager.ts
```diff
@@ -93,37 +105,158 @@ export class TerminalManagerImpl implements TerminalManager {
...
-                async () => {
-                    await waitForShellIntegration(terminal);
-                    await this.activate(terminal, environment);
-                },
+        if (actType === 'command') {
+            if (isActivatableEnvironment(environment)) {
+                await withProgress(
+                    {
+                        location: ProgressLocation.Window,
+                        title: `Activating environment: ${environment.environmentPath.fsPath}`,
```
eleanorjboyd, line 218: does this need to be l10?

src/features/terminal/terminalManager.ts
```diff
@@ -93,37 +105,158 @@ export class TerminalManagerImpl implements TerminalManager {
...
+                if (varMap) {
+                    varMap.forEach((value, key) => {
+                        envVars = { ...envVars, [key]: value };
+                    });
+                }
+            });
+        }
+
+        // https://github.com/microsoft/vscode-python-environments/issues/172
         // const name = options.name ?? `Python: ${environment.displayName}`;
```
eleanorjboyd, line 253: seems like this was left behind?

src/features/terminal/terminalManager.ts
```diff
@@ -93,37 +105,158 @@ export class TerminalManagerImpl implements TerminalManager {
...
+            vars.forEach((varMap) => {
+                if (varMap) {
+                    varMap.forEach((value, key) => {
+                        envVars = { ...envVars, [key]: value };
+                    });
+                }
+            });
+        }
+
+        // https://github.com/microsoft/vscode-python-environments/issues/172
```
eleanorjboyd, line 252: might be worth clarifying why this issue is there

src/features/terminal/terminalManager.ts
```diff
@@ -93,37 +105,158 @@ export class TerminalManagerImpl implements TerminalManager {
...
-            await withProgress(
-                {
-                    location: ProgressLocation.Window,
-                    title: `Activating environment: ${environment.environmentPath.fsPath}`,
-                },
-                async () => {
-                    await waitForShellIntegration(terminal);
-                    await this.activate(terminal, environment);
-                },
+        if (actType === 'command') {
```
eleanorjboyd, line 213: might be worth having constants for all the activation types instead of the strings

src/managers/conda/condaUtils.ts
```diff
@@ -842,3 +1012,50 @@ export async function getCommonCondaPackagesToInstall(
     const selected = await selectCommonPackagesOrSkip(common, installed ?? [], !!options.showSkipOption);
     return selected;
 }
+
+async function installPython(
```
eleanorjboyd, line 1016: is it possible to give a version to install? Or is this like quick install?

eleanorjboyd (approved):

src/features/terminal/shells/pwsh/pwshStartup.ts
karthiknadig: This is actually executable names, `pwsh` and `powershell`. And not actually the shell name that VS Code uses. The reason to not use that constant here is to avoid coupling two different values. This can cause issues if we decide to change the constant value.

src/features/terminal/terminalManager.ts
karthiknadig: Great suggestion, will add in a flow up PR.

src/features/terminal/terminalManager.ts
karthiknadig: Yes

src/features/terminal/terminalManager.ts
karthiknadig: Will do

src/features/terminal/terminalManager.ts
karthiknadig: No, this is what we need to uncomment after the issue above is fixed.

src/managers/conda/condaUtils.ts
karthiknadig: This is like quick install

PR #218, opened 2025-03-17
feat: package manager permissions when using API

karthiknadig:


src/common/localize.ts
```diff
@@ -144,3 +144,10 @@ export namespace EnvViewStrings {
     export const selectedGlobalTooltip = l10n.t('This environment is selected for non-workspace files');
     export const selectedWorkspaceTooltip = l10n.t('This environment is selected for workspace files');
 }
+
+export namespace PermissionsCommon {
+    export const allow = l10n.t('Allow');
+    export const deny = l10n.t('Deny');
+    export const ask = l10n.t('Ask');
+    export const setPermissions = l10n.t('Set Permissions');
```
cwebster-99, line 152:
```suggestion
    export const setPermissions = l10n.t('Update Permissions');
```

src/common/localize.ts
```diff
@@ -144,3 +144,10 @@ export namespace EnvViewStrings {
     export const selectedGlobalTooltip = l10n.t('This environment is selected for non-workspace files');
     export const selectedWorkspaceTooltip = l10n.t('This environment is selected for workspace files');
 }
+
+export namespace PermissionsCommon {
+    export const allow = l10n.t('Allow');
+    export const deny = l10n.t('Deny');
+    export const ask = l10n.t('Ask');
```
cwebster-99, line 151: I'm not sure "Ask" is easily understandable in this context. I wonder if we should do something like "Always Ask" or "Confirm each time" at the tradeoff of having a slightly longer button title

src/features/permissions/packageManagerPermissions.ts
```diff
@@ -0,0 +1,193 @@
...
+            extensionId,
+        ),
+        {
+            modal: true,
+            detail: currentPermission ? l10n.t('Current permission: {0}', currentPermission) : undefined,
+        },
+        PermissionsCommon.ask,
+        PermissionsCommon.allow,
+        PermissionsCommon.deny,
+    );
```
cwebster-99, line 179: This scenario currently shows a "Cancel" option as well. Can we remove this in favor of the X on the dialog? 
image

src/features/permissions/packageManagerPermissions.ts
```diff
@@ -0,0 +1,193 @@
...
+
+    const currentPermission = await pm.getPermissions(extensionId);
+    if (currentPermission === 'Allow') {
+        return true;
+    } else if (currentPermission === 'Deny') {
+        traceLog(`Package management permissions denied for extension: ${extensionId}`);
+        setImmediate(async () => {
+            const response = await showWarningMessage(
+                l10n.t(
+                    'The extension `{0}` is not allowed to {1} packages into your Python environment.',
```
cwebster-99, line 119: Suggest we update language to match "permissions" 
```suggestion
                    'The extension `{0}` is not permitted to {1} packages into your Python environment.',
```

cwebster-99 (commented):
Not sure where this is changed but I also saw a new option for "Skip package installation" after setting permissions to "Ask". I think we could remove this - was there a specific reason this was added? 
image

src/common/localize.ts
eleanorjboyd: agreed- I also wanted this button to be explicit it was a long-term choice

eleanorjboyd:
- Do we think it adds something to specify to the user experience to call our the actions the extension can make instead of just a generally statement like "can make changes to packages in your environment" or something? Also should we add "downgrade" in addition to upgrade since technically that is an option an extension could take right?
Screenshot 2025-03-18 at 6 50 06 PM

---


- Love this page here where it shows the current permissions!!
Screenshot 2025-03-18 at 6 52 28 PM

---

- Also related to the conversation @cwebster-99 started on wording for the set permissions button text but I feel like both the setting long term permissions and in this menu the short term permissions should not have the same button since I find it a bit confusing what action Im actually taking. Since this is the prompt that shows up when someone has this pkg manager at "always ask" so that dialog should be different than the setting long term permissions.
Screenshot 2025-03-18 at 6 52 49 PM

---

- when selecting "reset package permissions" I feel like a confirmation dialog would be good in case someone hits it by accident or so the user knows the weight of what they are doing. Would be nice to just have a dialog box that is like- remove all pkg management permissions set so far, this action is irreversible. We could even go a step further if we like a list all the existing pkg permissions set to show what would be deleted if we wanted. 

---

- Got a bit overwhelmed at this list- would be cool almost to surface the ones extensions that have values set at the top of the list in a different section, then all other extensions below but this is a stretch idea.
Screenshot 2025-03-18 at 6 48 47 PM

---


- one final random item, but any reason why you can't search in the extension menu for "publisher.extension" and only "extension". Might be worthwhile allowing users to search with that whole term or maybe even be able to search by a publisher (ie search for extensions published by microsft). Just an idea, curious @cwebster-99 and your thoughts here.

cwebster-99:
> Do we think it adds something to specify to the user experience to call our the actions the extension can make instead of just a generally statement like "can make changes to packages in your environment" or something?

I had the same thought initially. I think for these general views we could say "requests permissions to make changes to packages in your environment." 

> Since this is the prompt that shows up when someone has this pkg manager at "always ask" so that dialog should be different than the setting long term permissions.

Maybe we can say "Would you like to set permissions for {extension} to make changes to packages in your environment" and then the follow up prompts if a user selects "ask" are "the extension wants.....". So using set permissions vs the actual action that is being confirmed each time. 

> would be cool almost to surface the ones extensions that have values set at the top of the list in a different section, then all other extensions below but this is a stretch idea.

Ooh if would be cool if we could organize them similar to how we present environments i.e. a section for Allow, Deny, Ask and then no set permissions. A nice to have not necessarily a need to have IMO 

image

> Might be worthwhile allowing users to search with that whole term or maybe even be able to search by a publisher (ie search for extensions published by microsft).

I think this would be a good exploration but also a nice to have :)

karthiknadig:
We need to revisit this in a different way.

Issue #258, opened 2025-04-03
Fast create for Python Environments


karthiknadig:
This is where all selections are defaults. Contribute this as a LM Tool to make it really easy for Copilot to create environments

karthiknadig:
This was done last iteration. This is already in use by Jupyter and Python extension.

PR #265, opened 2025-04-07
feat: add "Quick Create" option for venvs

karthiknadig:
closes https://github.com/microsoft/vscode-python-environments/issues/258

image

/cc @cwebster-99 Need your input on the text

### DIFF
--- a/src/common/localize.ts
+++ b/src/common/localize.ts
@@ -89,6 +89,12 @@ export namespace VenvManagerStrings {
 
     export const installEditable = l10n.t('Install project as editable');
     export const searchingDependencies = l10n.t('Searching for dependencies');
+
+    export const selectQuickOrCustomize = l10n.t('Select Quick Create or Customize');
+    export const quickCreate = l10n.t('Quick Create');
+    export const quickCreateDescription = l10n.t('Create a .venv environment in the workspace root');
+    export const customize = l10n.t('Customize');
+    export const customizeDescription = l10n.t('Select python version, location, packages, etc.');
 }
 
 export namespace SysManagerStrings {
--- a/src/common/localize.ts
+++ b/src/common/localize.ts
@@ -90,11 +90,11 @@ export namespace VenvManagerStrings {
     export const installEditable = l10n.t('Install project as editable');
     export const searchingDependencies = l10n.t('Searching for dependencies');
 
-    export const selectQuickOrCustomize = l10n.t('Select Quick Create or Customize');
+    export const selectQuickOrCustomize = l10n.t('Select environment creation mode');
     export const quickCreate = l10n.t('Quick Create');
     export const quickCreateDescription = l10n.t('Create a virtual environment in the workspace root');
-    export const customize = l10n.t('Customize');
-    export const customizeDescription = l10n.t('Select python version, location, packages, etc.');
+    export const customize = l10n.t('Custom');
+    export const customizeDescription = l10n.t('Choose python version, location, packages, name, etc.');
 }
 
 export namespace SysManagerStrings {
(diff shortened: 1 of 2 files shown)


src/common/localize.ts
```diff
@@ -89,6 +89,12 @@ export namespace VenvManagerStrings {
 
     export const installEditable = l10n.t('Install project as editable');
     export const searchingDependencies = l10n.t('Searching for dependencies');
+
+    export const selectQuickOrCustomize = l10n.t('Select Quick Create or Customize');
+    export const quickCreate = l10n.t('Quick Create');
+    export const quickCreateDescription = l10n.t('Create a virtual environment in the workspace root');
+    export const customize = l10n.t('Customize');
```
cwebster-99:
```suggestion
    export const customize = l10n.t('Custom');
```

src/managers/builtin/venvUtils.ts
```diff
@@ -347,6 +347,90 @@ export async function getGlobalVenvLocation(): Promise<Uri | undefined> {
     return undefined;
 }
 
+async function createWithCustomization(version: string): Promise<boolean | undefined> {
+    const selection: QuickPickItem | undefined = await showQuickPick(
+        [
+            {
+                label: VenvManagerStrings.quickCreate,
+                description: VenvManagerStrings.quickCreateDescription,
+                detail: l10n.t('Will use Python version {0} and install workspace dependencies .', version),
```
cwebster-99:
```suggestion
                detail: l10n.t('Uses Python version {0} and installs workspace dependencies .', version),
```

src/managers/builtin/venvUtils.ts
```diff
@@ -347,6 +347,90 @@ export async function getGlobalVenvLocation(): Promise<Uri | undefined> {
...
+                    } else {
+                        await runPython(
+                            basePython.execInfo.run.executable,
+                            ['-m', 'venv', envPath],
+                            venvRoot.fsPath,
+                            manager.log,
+                        );
+                    }
+                    if (!(await fsapi.pathExists(pythonPath))) {
+                        log.error('no python executable found in virtual environment');
```
cwebster-99, line 414: Should we hid the quick create option if we already know there is no python? Eventually we could auto install python for the user but in the meantime it might be better to hide

src/common/localize.ts
```diff
@@ -89,6 +89,12 @@ export namespace VenvManagerStrings {
...
+    export const customizeDescription = l10n.t('Select python version, location, packages, etc.');
```
cwebster-99:
```suggestion
    export const customizeDescription = l10n.t('Choose python version, location, packages, name, etc.');
```

src/common/localize.ts
```diff
@@ -89,6 +89,12 @@ export namespace VenvManagerStrings {
 
     export const installEditable = l10n.t('Install project as editable');
     export const searchingDependencies = l10n.t('Searching for dependencies');
+
+    export const selectQuickOrCustomize = l10n.t('Select Quick Create or Customize');
```
cwebster-99: Not sure if "mode" is the best in this context
```suggestion
    export const selectQuickOrCustomize = l10n.t('Select environment creation mode');
```

src/managers/builtin/venvUtils.ts
karthiknadig: This is actually a different thing. This is after creating the venv, we could not find a python in the virtual environment. Typically this should not happen at all. This is just me being overly cautious.

karthiknadig: We actually throw error the moment user clicks create with no python installed.

eleanorjboyd (approved):

Issue #278, opened 2025-04-09
Add uri handling for notebooks and cells


karthiknadig:

eleanorjboyd:
looking at additions here to check for notebook uris

`PythonProjectManager: getProjects(uris?: Uri[]): ReadonlyArray<PythonProject>;`
`EnvironmentManagers:  getEnvironmentManager(scope: EnvironmentManagerScope): InternalEnvironmentManager | undefined;`
`PythonEnvironmentManagers: getPackageManager`

my only concern is that all actions where a URI is passed in (get, set, create, etc) seem to require this check which seems to require a huge number of additions. Wondering if there is a more streamlined way to do this or fewer points where we can implement it.

eleanorjboyd:
this is a bit complicated to verify since the entry-points here are hard to trigger. The best way I have found is:

1. create a notebook
2. run the notebook to get some simple output
3. attach the output to copilot chat
4. once it is attached, right click on it to get the link to the cell (it should be something like 'vscode-notebook-cell-output:/Users/eleanorboyd/testingFiles/notebook_workspace/adding_notebook.ipynb?openIn%3Dnotebook%26outputIndex%3D0#W1sZmlsZQ%3D%3D')
5.  now ask the agent specifically to get the python environment for the URI: xxxx (sometimes I even have to say like "use this as the resource path")
6. check the tool run to make sure it has the whole notebook URI
7. run the tool, confirm it works and returns you some information instead of erroring

osortega:
@eleanorjboyd it does seem to be listing the information related to the input path but when clicking on the attached output I get this error:

<img width="283" alt="Image" src="https://github.com/user-attachments/assets/a512746b-b621-471d-9edc-7fa21e4b6f34" />

Not sure if it's related to the same issue

DonJayamanne:
@osortega that’s a known issues
 @Yoyokrazy is looking into this

Yoyokrazy:
"Unable to Open" issue will be fixed with https://github.com/microsoft/vscode/pull/247795, so that'll unblock verification. 

It'll land in the next insiders.

amunger:
@eleanorjboyd - I just see "open to the side" (which doesn't do anything) when right clicking the reference.

eleanorjboyd:
yeah I think this is a bug I need to investigate which is separate

Yoyokrazy:
Talked with eleanor, verification of this issue can be done just confirming the URI

[...]

PR #280, opened 2025-04-09
Copilot tools

eleanorjboyd:
adds support for two copilot tools:
- get environment information
- install packages
closes https://github.com/microsoft/vscode-python-environments/issues/284

### DIFF
--- a/package.json
+++ b/package.json
@@ -491,22 +491,52 @@
         ],
         "languageModelTools": [
             {
-                "name": "python_get_packages",
-                "displayName": "Get Python Packages",
-                "modelDescription": "Returns the packages installed in the given Python file's environment. You should call this when you want to generate Python code to determine the users preferred packages. Also call this to determine if you need to provide installation instructions in a response.",
-                "toolReferenceName": "pythonGetPackages",
+                "name": "python_environment_tool",
+                "displayName": "Get Python Environment Information",
+                "modelDescription": "Returns the information about the Python environment for the given file or workspace. Information includes environment type, python version, run command and installed packages with versions.",
+                "toolReferenceName": "pythonGetEnvironmentInfo",
                 "tags": [],
                 "icon": "$(files)",
+                "canBeReferencedInPrompt": true,
                 "inputSchema": {
                     "type": "object",
                     "properties": {
-                        "filePath": {
+                        "resourcePath": {
                             "type": "string"
                         }
                     },
-                    "description": "The path to the Python file or workspace to get the installed packages for.",
+                    "description": "The path to the Python file or workspace to get the environment information for.",
                     "required": [
-                        "filePath"
+                        "resourcePath"
+                    ]
+                }
+            },
+            {
+                "name": "python_install_package_tool",
+                "displayName": "Install Python Package",
+                "modelDescription": "Installs Python packages in a workspace. You should call this when you want to install packages in the user's environment.",
+                "toolReferenceName": "pythonInstallPackage",
+                "tags": [],
+                "icon": "$(package)",
+                "canBeReferencedInPrompt": true,
+                "inputSchema": {
+                    "type": "object",
+                    "properties": {
+                        "packageList": {
+                            "type": "array",
+                            "items": {
+                                "type": "string"
+                            },
+                            "description": "The list of packages to install."
+                        },
+                        "workspacePath": {
+                            "type": "string",
+                            "description": "The path to the Python workspace to identify which environment to install packages in."
+                        }
+                    },
+                    "required": [
+                        "packageList",
+                        "workspacePath"
                     ]
                 }
             }
--- a/package.json
+++ b/package.json
@@ -493,7 +493,7 @@
             {
                 "name": "python_environment_tool",
                 "displayName": "Get Python Environment Information",
-                "modelDescription": "Returns the information about the Python environment for the given file or workspace. Information includes environment type, python version, run command and installed packages with versions.",
+                "modelDescription": "Provides details about the Python environment for a specified file or workspace, including environment type, Python version, run command, and installed packages with their versions.",
                 "toolReferenceName": "pythonGetEnvironmentInfo",
                 "tags": [],
                 "icon": "$(files)",
@@ -514,7 +514,7 @@
             {
                 "name": "python_install_package_tool",
                 "displayName": "Install Python Package",
-                "modelDescription": "Installs Python packages in a workspace. You should call this when you want to install packages in the user's environment.",
+                "modelDescription": "Installs Python packages in the given workspace. Use this tool to install packages in the user's chosen environment.",
                 "toolReferenceName": "pythonInstallPackage",
                 "tags": [],
                 "icon": "$(package)",
@@ -531,7 +531,7 @@
                         },
                         "workspacePath": {
                             "type": "string",
-                            "description": "The path to the Python workspace to identify which environment to install packages in."
+                            "description": "Path to Python workspace that determines the environment for package installation."
                         }
                     },
                     "required": [
--- a/src/features/copilotTools.ts
+++ b/src/features/copilotTools.ts
@@ -8,26 +8,40 @@ import {
     PreparedToolInvocation,
     Uri,
 } from 'vscode';
-import { PythonPackageGetterApi, PythonProjectEnvironmentApi } from '../api';
+import {
+    PythonCommandRunConfiguration,
+    PythonEnvironment,
+    PythonEnvironmentExecutionInfo,
+    PythonPackageGetterApi,
+    PythonPackageManagementApi,
+    PythonProjectEnvironmentApi,
+} from '../api';
 import { createDeferred } from '../common/utils/deferred';
 
-export interface IGetActiveFile {
-    filePath?: string;
+export interface IResourceReference {
+    resourcePath?: string;
+}
+
+interface EnvironmentInfo {
+    type: string; // e.g. conda, venv, virtualenv, sys
+    version: string;
+    runCommand: string;
+    packages: string[] | string; //include versions too
 }
 
 /**
- * A tool to get the list of installed Python packages in the active environment.
+ * A tool to get the information about the Python environment.
  */
-export class GetPackagesTool implements LanguageModelTool<IGetActiveFile> {
+export class GetEnvironmentInfoTool implements LanguageModelTool<IResourceReference> {
     constructor(private readonly api: PythonProjectEnvironmentApi & PythonPackageGetterApi) {}
     /**
-     * Invokes the tool to get the list of installed packages.
+     * Invokes the tool to get the information about the Python environment.
      * @param options - The invocation options containing the file path.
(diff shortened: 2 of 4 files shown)


src/features/copilotTools.ts
```diff
@@ -36,55 +50,181 @@ export class GetPackagesTool implements LanguageModelTool<IGetActiveFile> {
...
+                packages: 'no packages found',
+                runCommand: 'no run command found',
+            };
+
+            // environment
+            const environment: PythonEnvironment | undefined = await this.api.getEnvironment(resourcePath);
             if (!environment) {
                 // Check if the file is a notebook or a notebook cell to throw specific error messages.
-                if (fileUri.fsPath.endsWith('.ipynb') || fileUri.fsPath.includes('.ipynb#')) {
+                if (resourcePath.fsPath.endsWith('.ipynb') || resourcePath.fsPath.includes('.ipynb#')) {
```
eleanorjboyd: how do we want to handle getting a notebook at this point?

src/features/copilotTools.ts
```diff
@@ -36,55 +50,181 @@ export class GetPackagesTool implements LanguageModelTool<IGetActiveFile> {
...
                 }
-                throw new Error('No environment found');
+                throw new Error('No environment found for the provided resource path: ' + resourcePath.fsPath);
             }
+
+            const execInfo: PythonEnvironmentExecutionInfo = environment.execInfo;
+            const run: PythonCommandRunConfiguration = execInfo.run;
+            envInfo.runCommand = run.executable + (run.args && run.args.length > 0 ? ` ${run.args.join(' ')}` : '');
+            // TODO: check if this is the right way to get type
+            envInfo.type = environment.envId.managerId.split(':')[1];
```
eleanorjboyd: @karthiknadig should I error if the managerId isn't able to be split on `:`? or not include it? Also do I include a test for it?

src/features/copilotTools.ts
```diff
@@ -36,55 +50,181 @@ export class GetPackagesTool implements LanguageModelTool<IGetActiveFile> {
...
-        const message = 'Preparing to fetch the list of installed Python packages...';
+        const message = 'Preparing to fetch Python environment information...';
+        return {
+            invocationMessage: message,
+        };
+    }
+}
+
+function BuildEnvironmentInfoContent(envInfo: EnvironmentInfo): LanguageModelTextPart {
+    // Create a formatted string that looks like JSON but preserves comments
```
eleanorjboyd, line 134: specifically looking for feedback on how this is returned

src/features/copilotTools.ts
karthiknadig: I added this item as task for you https://github.com/microsoft/vscode-python-environments/issues/278

src/features/copilotTools.ts
eleanorjboyd: added error handling here

karthiknadig:
@cwebster-99 Need your input on the names and descriptions.

package.json
```diff
@@ -494,22 +494,52 @@
...
                     "required": [
-                        "filePath"
+                        "resourcePath"
+                    ]
+                }
+            },
+            {
+                "name": "python_install_package_tool",
+                "displayName": "Install Python Package",
+                "modelDescription": "Installs Python packages in a workspace. You should call this when you want to install packages in the user's environment.",
```
karthiknadig: This should probably be:
Installs Python packages for the given workspace. You should call this when you want to install packages in the user's chosen environment.

src/features/copilotTools.ts
```diff
@@ -36,55 +50,181 @@ export class GetPackagesTool implements LanguageModelTool<IGetActiveFile> {
...
+                if (resourcePath.fsPath.endsWith('.ipynb') || resourcePath.fsPath.includes('.ipynb#')) {
                     throw new Error('Unable to access Jupyter kernels for notebook cells');
                 }
-                throw new Error('No environment found');
+                throw new Error('No environment found for the provided resource path: ' + resourcePath.fsPath);
             }
+
+            const execInfo: PythonEnvironmentExecutionInfo = environment.execInfo;
+            const run: PythonCommandRunConfiguration = execInfo.run;
+            envInfo.runCommand = run.executable + (run.args && run.args.length > 0 ? ` ${run.args.join(' ')}` : '');
```
karthiknadig: I recommend looking at `runInTerminal.ts` to calculate the string for this.

src/features/copilotTools.ts
karthiknadig: Actually, an alternative here it to get the Env Manager instance, using managerId. That should have `name` property so you don't have to extract this.

src/features/copilotTools.ts
```diff
@@ -36,55 +50,181 @@ export class GetPackagesTool implements LanguageModelTool<IGetActiveFile> {
...
+            // TODO: check if this is the right way to get type
+            envInfo.type = environment.envId.managerId.split(':')[1];
+            envInfo.version = environment.version;
+
+            // does this need to be refreshed prior to returning to get any new packages?
```
karthiknadig: Eventually, we should not need to refresh packages. This provides a good fall back if terminal is used to install packages for some reason.

src/features/copilotTools.ts
```diff
@@ -36,55 +50,181 @@ export class GetPackagesTool implements LanguageModelTool<IGetActiveFile> {
...
+        return {
+            invocationMessage: message,
+        };
+    }
+}
+
+function BuildEnvironmentInfoContent(envInfo: EnvironmentInfo): LanguageModelTextPart {
+    // Create a formatted string that looks like JSON but preserves comments
+    const content = `{
+  // type of python environment; sys means it is the system python
```
karthiknadig: The classification for system pythons is `system`. we don't use `sys`. I think the description here should change based on `system` vs other Pythons. We need to say something like this:

if type is system then use this description: "System pythons are pythons that ship with the OS or are installed globally. These python installs may be used by the OS for running services and core functionality. Confirm with the user before installing packages into this environment, as it can lead to issues with any services on the OS"

if type is not system: "This environment is managed by {type} environment manager. Use the install tool to install packages into this environment."

src/features/copilotTools.ts
```diff
@@ -36,55 +50,181 @@ export class GetPackagesTool implements LanguageModelTool<IGetActiveFile> {
...
+  "environmentType": ${JSON.stringify(envInfo.type)},
+  // python version of the environment
+  "pythonVersion": ${JSON.stringify(envInfo.version)},
+  // command to run python in this environment, will include command with active environment if applicable
```
karthiknadig: The description could be assertive:
"Use this command to run python script or code in the terminal"

I don't think there is need to specify activation.

src/features/copilotTools.ts
```diff
@@ -36,55 +50,181 @@ export class GetPackagesTool implements LanguageModelTool<IGetActiveFile> {
...
+                throw new Error('No environment found');
+            }
+
+            // Install the packages
+            await this.api.installPackages(environment, parameters.packageList);
+            const resultMessage = `Successfully installed ${packagePlurality}: ${parameters.packageList.join(', ')}`;
+
+            // Refresh packages after installation to update the package view
+            //TODO: do I want the await?
+            await this.api.refreshPackages(environment);
```
karthiknadig: Since you are installing using `installPackages` there is no need to run `refreshPackages` again.  The `getPackages` command here should give you the right set of packages.

karthiknadig (changes requested):

package.json
```diff
@@ -494,22 +494,52 @@
         ],
         "languageModelTools": [
             {
-                "name": "python_get_packages",
-                "displayName": "Get Python Packages",
-                "modelDescription": "Returns the packages installed in the given Python file's environment. You should call this when you want to generate Python code to determine the users preferred packages. Also call this to determine if you need to provide installation instructions in a response.",
-                "toolReferenceName": "pythonGetPackages",
+                "name": "python_environment_tool",
+                "displayName": "Get Python Environment Information",
+                "modelDescription": "Returns the information about the Python environment for the given file or workspace. Information includes environment type, python version, run command and installed packages with versions.",
```
cwebster-99:
```suggestion
                "modelDescription": "Provides details about the Python environment for a specified file or workspace, including environment type, Python version, run command, and installed packages with their versions.",
```

package.json
```diff
@@ -494,22 +494,52 @@
         ],
         "languageModelTools": [
             {
-                "name": "python_get_packages",
-                "displayName": "Get Python Packages",
-                "modelDescription": "Returns the packages installed in the given Python file's environment. You should call this when you want to generate Python code to determine the users preferred packages. Also call this to determine if you need to provide installation instructions in a response.",
-                "toolReferenceName": "pythonGetPackages",
+                "name": "python_environment_tool",
+                "displayName": "Get Python Environment Information",
```
cwebster-99, line 498: Is the norm for these tools to be named "Get...."? If not, I might suggest we rename to "Python Environment Details" or "Python Environment Information"

package.json
cwebster-99:
```suggestion
                "modelDescription": "Installs Python packages in the given workspace. Use this tool to add packages to the user's chosen environment.",
```

package.json
```diff
@@ -494,22 +494,52 @@
...
+                        "packageList": {
+                            "type": "array",
+                            "items": {
+                                "type": "string"
+                            },
+                            "description": "The list of packages to install."
+                        },
+                        "workspacePath": {
+                            "type": "string",
+                            "description": "The path to the Python workspace to identify which environment to install packages in."
```
cwebster-99:
```suggestion
                            "description": "Path to Python workspace that determines the environment for package installation."
```

src/features/copilotTools.ts
eleanorjboyd: great thank you!

src/features/copilotTools.ts
eleanorjboyd: Do I care about shellType ? Like should I incorporate this: `if (shellType === 'pwsh' && !text.startsWith('&')) {`

src/features/copilotTools.ts
eleanorjboyd: will leave for now but add todo

src/features/copilotTools.ts
eleanorjboyd: I don't think im using `getPackages` in this tool- should I do that first before installing to avoid attempting to install something that already exists?

eleanorjboyd: (pulling in your package api changes now, question still stands tho regarding running get before install)

src/features/copilotTools.ts
karthiknadig: No need to use shell type. At least for now.

src/features/copilotTools.ts
karthiknadig: The thing is, the installer will handle the case where something is already installed. So no need to use `getPackages` before calling install.

PR #293, opened 2025-04-11
add handling for notebook-cells as inputs to API

eleanorjboyd:
fixes https://github.com/microsoft/vscode-python-environments/issues/278

### DIFF
--- a/src/common/utils/pathUtils.ts
+++ b/src/common/utils/pathUtils.ts
@@ -1,4 +1,16 @@
+import { Uri } from 'vscode';
 import { isWindows } from '../../managers/common/utils';
+
+export function convertNotebookCellUriToNotebookUri(uri: Uri): Uri {
+    if (uri.scheme === 'vscode-notebook-cell') {
+        return Uri.from({
+            scheme: 'vscode-notebook',
+            path: uri.path,
+            authority: uri.authority,
+        });
+    }
+    return uri;
+}
 
 export function normalizePath(path: string): string {
     const path1 = path.replace(/\\/g, '/');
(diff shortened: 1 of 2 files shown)


src/features/pythonApi.ts
```diff
@@ -171,32 +171,84 @@ class PythonEnvironmentApiImpl implements PythonEnvironmentApi {
...
 
-        const items = await manager.getEnvironments(scope);
+        const items = await manager.getEnvironments(currentScope);
         return items;
     }
     onDidChangeEnvironments: Event<DidChangeEnvironmentsEventArgs> = this._onDidChangeEnvironments.event;
     setEnvironment(scope: SetEnvironmentScope, environment?: PythonEnvironment): Promise<void> {
-        return this.envManagers.setEnvironment(scope, environment);
+        let currentScope: SetEnvironmentScope = scope;
+        if (scope instanceof Uri && scope.scheme === 'vscode-notebook-cell') {
```
karthiknadig: This seems to be duplicated a lot. We could consolidate this.

karthiknadig (changes requested):

src/features/pythonApi.ts
eleanorjboyd: addressed

karthiknadig (approved):

src/common/utils/pathUtils.ts
```diff
@@ -1,5 +1,24 @@
+import { Uri } from 'vscode';
 import { isWindows } from '../../managers/common/utils';
 
+export function checkUri(scope?: Uri | Uri[] | string): Uri | Uri[] | string | undefined {
+    if (scope instanceof Uri) {
+        if (scope.scheme === 'vscode-notebook-cell') {
+            return Uri.from({
```
DonJayamanne, line 7: @eleanorjboyd  this isn't a valid notebook uri
 haven't looked at the code, perhaps this has some meaning in this extension
However of you want to find the notebook Uri, you must iterate through all notebooks and find the notebook that has a cell with the matching Uri .

Issue #305, opened 2025-04-14
auto-find projects bugs / edits


eleanorjboyd:
## in `create` function:
- `src/features/creators/autoFindProjects.ts`
- `if (filtered.length === 0) {` line 80ish (same with line 88ish)
    - This should have some way to notify the user so they understand what happened and that no projects were found.

### in `addPythonProject` function:
- `src/features/envCommands.ts`
already made the edit to add 
```
    if (
        resource instanceof ProjectPackageRootTreeItem ||
        resource instanceof ProjectPackage ||
        resource instanceof ProjectEnvironment
    ) {
        await addPythonProject(undefined, wm, em, pc);
    }
```
but I think it needs even more refinement. Might be worth trying if its not a recognized resource then we re-call with undefined so we don't have to handle so many exceptions?

eleanorjboyd:
1. open workspace with no pyproject.toml
2. click folder icon in projects view to add new project
3. select "auto find projects"
4. make sure you see a notification that no projects were found

PR #306, opened 2025-04-14
update for existing project creation

eleanorjboyd:
fixes https://github.com/microsoft/vscode-python-environments/issues/305

### DIFF
--- a/src/features/creators/existingProjects.ts
+++ b/src/features/creators/existingProjects.ts
@@ -4,6 +4,8 @@ import { ProjectCreatorString } from '../../common/localize';
 import { showOpenDialog, showWarningMessage } from '../../common/window.apis';
 import { PythonProjectManager } from '../../internal.api';
 import { traceInfo } from '../../common/logging';
+import { Uri, window, workspace } from 'vscode';
+import { traceLog } from '../../common/logging';
 
 export class ExistingProjects implements PythonProjectCreator {
     public readonly name = 'existingProjects';
@@ -23,6 +25,7 @@ export class ExistingProjects implements PythonProjectCreator {
         });
 
         if (!results || results.length === 0) {
+            // User cancelled the dialog & doesn't want to add any projects
             return;
         }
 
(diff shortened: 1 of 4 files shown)


src/features/creators/existingProjects.ts
```diff
@@ -22,7 +26,30 @@ export class ExistingProjects implements PythonProjectCreator {
...
         }
 
-        return results.map((r) => ({
+        // do we have any limitations that need to be applied here?
+        // like selected folder not child of workspace folder?
+
+        const filtered = results.filter((uri) => {
+            const p = this.pm.get(uri);
+            if (p) {
+                // Skip this project if there's already a project registered with exactly the same path
```
eleanorjboyd, line 38: wondering if also we should add in the child project check here too

src/features/creators/existingProjects.ts
```diff
@@ -22,7 +26,30 @@ export class ExistingProjects implements PythonProjectCreator {
...
+                return np !== nf;
+            }
+            return true;
+        });
+
+        if (filtered.length === 0) {
+            // No new projects found that are not already in the project manager
+            traceInfo('All discovered projects are already registered in the project manager');
+            setImmediate(() => {
+                showWarningMessage('No new projects found');
```
eleanorjboyd, line 50: I found it confusing as a user to see nothing when no projects were found but not sure exactly how to surface that error to the user

src/features/envCommands.ts
```diff
@@ -348,7 +348,12 @@ export async function addPythonProject(
         return;
     }
 
-    if (resource instanceof ProjectPackageRootTreeItem || resource instanceof ProjectPackage) {
+    if (
+        resource instanceof ProjectPackageRootTreeItem ||
+        resource instanceof ProjectPackage ||
+        resource instanceof ProjectEnvironment
+    ) {
```
eleanorjboyd: Here I found again another object type for the resource that caused no behavior to occur. Wondering if we switch to just retrying if its not a recognized resource so we don't have to handle so many exceptions? @karthiknadig

eleanorjboyd (commented):
@karthiknadig let me know your thoughts- just a few items I saw as I was reviewing the existing project creators

src/features/envCommands.ts
karthiknadig: We could just add an `else` case here, that captures all these variants.

src/features/creators/existingProjects.ts
karthiknadig: I think this warning makes sense.

src/features/creators/existingProjects.ts
karthiknadig: I think this is fine for now. I need to handle some multi-root scenarios better here.

karthiknadig (approved):

PR #312, opened 2025-04-16
Get Notebook Uri from Cell Logic Fix

eleanorjboyd:


amunger (dismissed):

src/common/utils/pathUtils.ts
```diff
@@ -1,24 +1,50 @@
...
+import { NotebookCell, NotebookDocument, Uri, workspace } from 'vscode';
 import { isWindows } from '../../managers/common/utils';
 
 export function checkUri(scope?: Uri | Uri[] | string): Uri | Uri[] | string | undefined {
+    if (!scope) {
+        return undefined;
+    }
+
+    if (Array.isArray(scope)) {
+        return scope.map((item) => checkUri(item) as Uri);
```
DonJayamanne: Curious, why are we casting to `Uri` here?

src/common/utils/pathUtils.ts
```diff
@@ -1,24 +1,50 @@
...
+    const index = notebook
+        .getCells()
+        .findIndex(
+            (cell) =>
+                isEqual(cell.document.uri, cellUri) ||
+                (cell.document.uri.fragment === cellUri.fragment && cell.document.uri.path === cellUri.path),
+        );
+    if (index !== -1) {
+        return notebook.getCells()[index];
+    }
```
DonJayamanne:
```suggestion
    return notebook
        .getCells()
        .find((cell) => isEqual(cell.document.uri, cellUri));
```

There's no need to check fragment etc, thats done in copilot extension/core due to cell output scheme (the new scheme you added).

src/common/utils/pathUtils.ts
```diff
@@ -1,24 +1,50 @@
...
+}
+
+function isEqual(a: Uri, b: Uri): boolean {
+    return a.toString() === b.toString();
```
DonJayamanne: To avoid running into issues with Casing of paths such as drive letters and the like, i would suggest pulling in the code for `Uri` comparison from Core.
This works, but best to pull stuff from core.

src/common/utils/pathUtils.ts
```diff
@@ -1,24 +1,50 @@
...
     if (scope instanceof Uri) {
         if (scope.scheme === 'vscode-notebook-cell') {
-            return Uri.from({
-                scheme: 'vscode-notebook',
-                path: scope.path,
-                authority: scope.authority,
+            // If the scope is a cell Uri, we need to find the notebook document it belongs to.
+            const matchingDoc = workspace.notebookDocuments.find((doc) => {
+                const cell = findCell(scope, doc);
+                return cell !== undefined;
```
DonJayamanne:
```suggestion
            const matchingDoc = workspace.notebookDocuments.find((doc) => findCell(scope, doc));
```

src/common/utils/pathUtils.ts
DonJayamanne: This change is required, else I fear this style of checking fragments will continue to be used elsewhere for matching cell uris.

DonJayamanne (changes requested):
Please avoid using `fragment`, this is done in core only because of the cell uri output scheme

DonJayamanne (changes requested):
Need a few changes

amunger (approved):

PR #324, opened 2025-04-22
New package Flow

eleanorjboyd:
Implement the flow to allow for "New Package" support


src/features/creators/creationHelpers.ts
```diff
@@ -0,0 +1,129 @@
...
+            (m) => m.id === 'ms-python.python:venv',
+        );
+        const destUri = Uri.parse(destFolder);
+        if (envManager && envManager.create) {
+            // with quickCreate enabled, user will not be prompted when creating the environment
+            const options: CreateEnvironmentOptions = { quickCreate: false };
+            if (envManager.quickCreateConfig) {
+                options.quickCreate = true;
+            }
+            const pyEnv = await envManager.create(destUri, options);
```
eleanorjboyd: getting an error when creating this but see the venv folder correctly setup inside my project. I think it tries to run ` Running: uv pip install --python /Users/eleanorboyd/testingFiles/notebook-for-copilot/yay_project/.venv/bin/python` right after creating the venv and that is causing issues? 

```
2025-04-22 15:02:57.432 [info] error: / does not appear to be a Python project, as neither `pyproject.toml` nor `setup.py` are present in the directory

2025-04-22 15:02:57.433 [error] Error managing packages Failed to run uv pip install --python /Users/eleanorboyd/testingFiles/notebook-for-copilot/yay_project/.venv/bin/python -r /Users/
```

src/features/creators/creationHelpers.ts
```diff
@@ -0,0 +1,129 @@
...
+}
+
+export async function removeCopilotInstructions(destFolder: string) {
+    const copilotFolder = path.join(destFolder, '.github');
+    if (await fs.pathExists(copilotFolder)) {
+        await fs.remove(copilotFolder);
+    }
+}
+
+export async function quickCreateNewVenv(envManagers: EnvironmentManagers, destFolder: string) {
```
eleanorjboyd, line 73: @karthiknadig specifically looking for feedback here. Not sure if I am doing these steps in the recommended way

src/features/creators/creationHelpers.ts
```diff
@@ -0,0 +1,129 @@
...
+            const options: CreateEnvironmentOptions = { quickCreate: false };
+            if (envManager.quickCreateConfig) {
+                options.quickCreate = true;
+            }
+            const pyEnv = await envManager.create(destUri, options);
+            // comes back as undefined if this doesn't work
+            traceVerbose(`Created venv at: ${pyEnv?.environmentPath} using ${envManager.name}`);
+        } else {
+            // find an environment manager that supports create
+            const envManager = envManagers.managers.find((m) => m.create);
```
eleanorjboyd: do we expect this case to ever happen? Where the venv env manager doesn't exist? I don't think so but unsure if I should make that assumption

src/features/creators/creationHelpers.ts
```diff
@@ -0,0 +1,129 @@
...
+    const regex = new RegExp(escapedSearchValue, 'g');
+    let content = await fs.readFile(filePath, 'utf8');
+    if (content.includes(searchValue)) {
+        content = content.replace(regex, replaceValue);
+        await fs.writeFile(filePath, content, 'utf8');
+    }
+}
+
+// Helper to recursively replace all occurrences of a string in file/folder names and file contents
+export async function replaceInFilesAndNames(dir: string, searchValue: string, replaceValue: string) {
```
eleanorjboyd, line 105: @karthiknadig is it the correct strategy here to copy then rename or should I not do the rename in place like this?

src/features/creators/newPackageProject.ts
```diff
@@ -1,18 +1,114 @@
...
+        }
+        const execInfo = pythonEnvironment.execInfo;
+        if (execInfo.run) {
+            let execRunStr = execInfo.run.executable;
+            if (execInfo.run.args) {
+                execRunStr += ` ${execInfo.run.args.join(' ')}`;
+            }
+            await replaceInFile(readmeFilePath, '<run_exec>', execRunStr);
+        }
+        // TODO: replace <activation_command> in README.md ?
```
eleanorjboyd: @karthiknadig do you think its useful to include activation command?

src/features/creators/templates/newPackageTemplate/README.md
```diff
@@ -0,0 +1,44 @@
+# <project_name>
+
+> TODO: Write a short description of your project and what it does.
+
+## Table of Contents
+
+- [\<project\_name\>](#project_name)
```
eleanorjboyd: need to fix this as it won't get replaced correctly on the find and replace call

src/features/creators/templates/newPackageTemplate/README.md
```diff
@@ -0,0 +1,44 @@
...
+## How to Package
+
+> TODO: Explain how to build and distribute the package (e.g., using setuptools, poetry, etc.).
+
+1. `<run_exec> -m build`
+
+
+## How to Test
+Tests are stored in the `tests` folder. Tests can be run using pytest with the following command:
+`<run_exec> pytest`
```
eleanorjboyd: I think its ok to be opinionated and just push for pytest right?

src/features/creators/templates/newPackageTemplate/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Setup for New Contributors
+
+> TODO: List steps for a new developer to get started.
+
+1. Activate and select the necessary environment, `<activation_command>`
+2. Install dependencies:
```
eleanorjboyd: is it worth including both here? Considering just the first but not sure

src/features/creators/templates/newPackageTemplate/README.md
```diff
@@ -0,0 +1,44 @@
...
+  - [How to Run](#how-to-run)
+  - [How to Package](#how-to-package)
+  - [How to Test](#how-to-test)
+  - [Setup for New Contributors](#setup-for-new-contributors)
+  - [License](#license)
+
+## How to Run
+
+> TODO: Describe how to use or run your package, with example commands or code.
+1. `<run_exec> -m <package_name>`
```
eleanorjboyd: the run exec is returned as absolute here which we don't want in the readme- need to figure out how to make it relative and if it should be relative to root folder? I would assume thats the goal
Screenshot 2025-04-22 at 3 33 05 PM

src/features/creators/creationHelpers.ts
```diff
@@ -0,0 +1,149 @@
...
+import { CreateEnvironmentOptions, EnvironmentManager } from '../../api';
+import { traceVerbose } from '../../common/logging';
+
+/**
+ * Prompts the user to choose whether to create a new virtual environment (venv) for a package.
+ * @returns {Promise<boolean | undefined>} Resolves to true if 'Yes' is selected, false if 'No', or undefined if cancelled.
+ */
+export async function promptForVenv(): Promise<boolean | undefined> {
+    const venvChoice = await window.showQuickPick(['Yes', 'No'], {
+        placeHolder: 'Would you like to create a new venv for this package?',
```
cwebster-99: We should probably say "virtual environment" in case the user has a different default that `venv`
```suggestion
        placeHolder: 'Would you like to create a new virtual environment for this package?',
```

src/features/creators/creationHelpers.ts
```diff
@@ -0,0 +1,149 @@
...
+/**
+ * Prompts the user to choose whether to create a Copilot instructions file, only if Copilot is installed.
+ * @returns {Promise<boolean | undefined>} Resolves to true if 'Yes' is selected, false if 'No', or undefined if cancelled or Copilot is not installed.
+ */
+export async function promptForCopilotInstructions(): Promise<boolean | undefined> {
+    if (!isCopilotInstalled()) {
+        return undefined;
+    }
+    const copilotChoice = await window.showQuickPick(['Yes', 'No'], {
+        placeHolder: 'Would you like to create a Copilot instructions file?',
```
cwebster-99, line 52: I am not sure we should prompt for this vs just including it if the user has Copilot installed. Users are still discovering what this is but we know it helps improve their experience with Copilot so I would opt to just include it for packages

src/features/creators/creationHelpers.ts
cwebster-99: As far as I understand this will always be supported via our default with `venv`

src/features/creators/newPackageProject.ts
```diff
@@ -1,18 +1,115 @@
...
+            'newPackageTemplate',
+        );
+        if (!(await fs.pathExists(templateFolder))) {
+            window.showErrorMessage('Template folder does not exist.');
+            return undefined;
+            // might need another check or error handling here
+        }
+        const workspaceFolders = workspace.workspaceFolders;
+        if (!workspaceFolders || workspaceFolders.length === 0) {
+            window.showErrorMessage('No workspace folder is open.');
```
cwebster-99: Are we able to open a workspace folder for them if it is empty using the logic Bhavya does with Create New?

src/features/creators/templates/newPackageTemplate/.github/copilot-instructions.md
```diff
@@ -0,0 +1,8 @@
+# Copilot Instructions for <package_name>
```
cwebster-99: See https://github.com/microsoft/vscode-python-templates/pull/4/files#diff-8263202d70318618891dd1a86fb62ccab9567201cc456e0085703a8c400b95f6 for additional items we can add to the instructions file

src/features/creators/templates/newPackageTemplate/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+## How to Package
```
cwebster-99:
```suggestion
## How to package
```

src/features/creators/templates/newPackageTemplate/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+> TODO: Explain how to build and distribute the package (e.g., using setuptools, poetry, etc.).
+
+1. `<run_exec> -m build`
+
+
+## How to Test
```
cwebster-99:
```suggestion
## How to test
```

src/features/creators/templates/newPackageTemplate/README.md
```diff
@@ -0,0 +1,44 @@
...
+Tests are stored in the `tests` folder. Tests can be run using pytest with the following command:
+`<run_exec> pytest`
+
+## Setup for New Contributors
```
cwebster-99:
```suggestion
## Setup for new contributors
```

src/features/creators/templates/newPackageTemplate/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+> TODO: List steps for a new developer to get started.
+
+1. Activate and select the necessary environment, `<activation_command>`
+2. Install dependencies:
+    a. with `<run_exec> pip install -e .[dev]`
+    b. with `<run_exec> pip install -r dev_requirements.txt` and `<run_exec> pip install -e .`
+3. then run the package with: `<run_exec> -m <package_name>`
```
cwebster-99:
```suggestion
3. Run the package with: `<run_exec> -m <package_name>`
```

src/features/creators/creationHelpers.ts
karthiknadig: There is no package being passed in for this to install, and that might be causing this.

src/features/creators/templates/newPackageTemplate/.gitignore
```diff
@@ -0,0 +1,35 @@
+# Environments
+.env
```
karthiknadig: `.env` is a type of file. we need to go over these entries.

src/features/creators/templates/newPackageTemplate/.vscode/launch.json
```diff
@@ -0,0 +1,6 @@
+{
+    "configurations": [
+        { "name": "Python Debugger: Module", "type": "debugpy", "request": "launch", "module": "<package_name>" }
+    ]
+}
+// TODO: Add more configurations as needed
```
karthiknadig: remove this

src/features/creators/templates/newPackageTemplate/README.md
karthiknadig: You would probably want to make this an editable install, that way you can run with just `python -m package_name`

src/features/creators/templates/newPackageTemplate/package_name/__main__.py
```diff
@@ -0,0 +1,8 @@
+# TODO: Update the main function to your needs or remove it.
+
+
+def main() -> None:
+    print("Start coding in Python today!")
+
+
+main()
```
karthiknadig:
```suggestion
if __name__ == "__main__":
    main()
```

src/features/creators/creationHelpers.ts
```diff
@@ -0,0 +1,149 @@
...
+import { EnvironmentManagers } from '../../internal.api';
+import { CreateEnvironmentOptions, EnvironmentManager } from '../../api';
+import { traceVerbose } from '../../common/logging';
+
+/**
+ * Prompts the user to choose whether to create a new virtual environment (venv) for a package.
+ * @returns {Promise<boolean | undefined>} Resolves to true if 'Yes' is selected, false if 'No', or undefined if cancelled.
+ */
+export async function promptForVenv(): Promise<boolean | undefined> {
+    const venvChoice = await window.showQuickPick(['Yes', 'No'], {
```
DonJayamanne: You'd want to localize these strings, easily done using `vscode.l10n.t('Yes')`

src/features/creators/templates/newPackageTemplate/.github/copilot-instructions.md
```diff
@@ -0,0 +1,10 @@
+# Copilot Instructions for package_name
```
eleanorjboyd: @karthiknadig:

- I have it right now that the template as a folder called `<my_package>-folder` which then has a folder `<my_package>.` Did this so I had somewhere to store the `pyproject.toml` and `dev-requirements.txt`. This does make it more complicated to do an editable install of the package at the root, not sure if this is worth the concern? I also worry people will delete / more that folder with higher frequency which then makes these instructions out of date

src/features/creators/creationHelpers.ts
eleanorjboyd: ok currently investigating. What seems to happen is when calling  `quickCreateVenv`, it calls `const project = api.getPythonProject(venvRoot); `. On the new project folder. This then ends up with the getter for python project which gets nothing for `let pythonProject = this._projects.get(uri.toString());` and falls back to `pythonProject = this.findProjectByUri(uri);`. Somewhere I must not be setting the project for the given URI and so it attemps to find it during quick create and fails. This causes quick create to get all of the workspace's package installs (which are like all -r dev_requirements.txt) which it tries to install all at once and errors.

eleanorjboyd: so the problem then boils down to, how do I make it so the venv gets created after the project gets registered or remove the requirement for registration as part of the quickEnv process? Since the function for create is called in `envCommands.ts` with `export async function addPythonProject`. This calls create and only after the create function finishes the project is pushed to the known projects. But as above, when I create the venv it tries to find the project which is not yet registered. Not sure the best design idea- curious your thoughts @karthiknadig

src/features/creators/newPackageProject.ts
eleanorjboyd: Bhayva should supply a path where the new project should be created- is the concern if that path doesn't exist?

src/features/creators/creationHelpers.ts
eleanorjboyd: switched to just always including if they have copilot

src/features/creators/templates/newPackageTemplate/.github/copilot-instructions.md
eleanorjboyd: my question here is if I should add all these general rules or if that wont feel applicable to the action the user is taking of adding the project? cc @karthiknadig

src/features/creators/creationHelpers.ts
eleanorjboyd: hm I am no longer seeing this error anymore and the PR to fix this is not merged with this- need to do more investigation as it seems this is no longer an issue (edit: nvm no error message is being thrown but too many requirements are being installed)

src/features/creators/creationHelpers.ts
```diff
@@ -0,0 +1,199 @@
...
+import { showQuickPickWithButtons } from '../../common/window.apis';
+import { EnvironmentManagers, InternalEnvironmentManager } from '../../internal.api';
+
+/**
+ * Prompts the user to choose whether to create a new virtual environment (venv) for a project, with a clearer return and early exit.
+ * @returns {Promise<boolean | undefined>} Resolves to true if 'Yes' is selected, false if 'No', or undefined if cancelled.
+ */
+export async function promptForVenv(callback: () => void): Promise<boolean | undefined> {
+    try {
+        const venvChoice = await showQuickPickWithButtons([{ label: l10n.t('Yes') }, { label: l10n.t('No') }], {
```
karthiknadig, line 15: We have localized versions of this in `localize.ts`. You can address this later.

PR #334, opened 2025-04-23
update return type of project create

eleanorjboyd:
update the API surface to change return type of the project creator

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -703,6 +703,7 @@ export interface PythonProjectCreator {
 
     /**
      * Creates a new Python project(s) or, if files are not a project, returns Uri(s) to the created files.
+     * Anything that needs its own python environment constitutes a project.
      * @param options Optional parameters for creating the Python project.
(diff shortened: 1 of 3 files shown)


src/api.ts
```diff
@@ -702,11 +702,14 @@ export interface PythonProjectCreator {
     readonly iconPath?: IconPath;
 
     /**
-     * Creates a new Python project or projects.
-     * @param options - Optional parameters for creating the Python project.
-     * @returns A promise that resolves to a Python project, an array of Python projects, or undefined.
-     */
-    create(options?: PythonProjectCreatorOptions): Promise<PythonProject | PythonProject[] | undefined>;
+     * Creates a new Python project(s) or, if files are not a project, returns Uri(s) to the created files.
```
karthiknadig, line 710: Can we add a minor addition to the `PythonProject`? something like:
Anything that needs its own python environment constitutes a project.

karthiknadig (approved):

PR #336, opened 2025-04-24
switch new project add to create method

eleanorjboyd:
- renamed `addPythonProject` in  src/features/envCommands.ts  to `addPythonProjectCommand` (as it is only a command entry point)
- moved responsibility so ProjectCreators are responsible for adding their created projects to the ProjectManager list
- updated `addPythonProjectCommand` so it calls the selected ProjectCreator create method OR (if the resource param already a ProjectItem) just add it to the ProjectManager list


src/features/envCommands.ts
```diff
@@ -338,84 +329,59 @@ export async function setPackageManagerCommand(em: EnvironmentManagers, wm: Pyth
...
 }
 
-export async function addPythonProject(
+/**
+ * Creates a new Python project using a selected PythonProjectCreator.
+ *
+ * This function calls create on the selected creator and handles the creation process. Will return
+ * without doing anything if the resource is a ProjectItem, as the project is already created.
+ *
+ * @param resource - The resource to use for project creation (can be a Uri, ProjectItem, or undefined).
```
eleanorjboyd: is it right that these are the options? Can it be an array of items? @karthiknadig

eleanorjboyd: nvm yes it can be an array, fixed it for this

eleanorjboyd:
@karthiknadig question here is how we might know if findExisting project should be a project or not. Like if it has an env or not. Not sure how to tell this and build it into the logic

karthiknadig:
We should remove the autofind projects feature and think about a better approach to it. What we have now was a demonstration of what you could do.

karthiknadig (approved):

PR #365, opened 2025-04-29
Expanding setting description for auto activate

cwebster-99:


cwebster-99:
@karthiknadig is this the only place I need to make changes to update the setting description?

rzhao271 (approved):

PR #374, opened 2025-05-01
Reformatting links to media

cwebster-99:


README.md
```diff
@@ -19,13 +19,13 @@ The "Python Projects" fold shows you all of the projects that are currently in y
 
 The "Environment Managers" fold shows you all of the environment managers that are available on your machine with all related environments nested below. From this view, you can create new environments, delete old environments, and manage packages.
 
-<img src=<https://raw.githubusercontent.com/microsoft/vscode-python-environments/main/images/python-envs-overview.gif> width=734 height=413>
+![Gif showing the Python Environments UI and features.](/images/python-envs-overview.gif)
```
karthiknadig: I think this will break the readme on market place. I think the `<` and `>` in the href was the problem. Just removing that should be enough.

cwebster-99: Reverted the changes and add " " around the ref links based on how we format in the release notes as well

karthiknadig (approved):

Issue #398, opened 2025-05-09
Bug Fixes: New Project: Auto-Find


eleanorjboyd:
bug fixes left for the new project creator `Auto-Find`

- [ ] ignore `pyproject.toml` files located in `.venv`
- [ ] add support for `quick create`

eleanorjboyd:
no need to verify- verify in sub-issues

PR #405, opened 2025-05-13
bug fixes autoFindProjects

eleanorjboyd:
fixes https://github.com/microsoft/vscode-python-environments/issues/398

### DIFF
--- a/src/features/creators/autoFindProjects.ts
+++ b/src/features/creators/autoFindProjects.ts
@@ -54,6 +54,8 @@ export class AutoFindProjects implements PythonProjectCreator {
     public readonly displayName = ProjectCreatorString.autoFindProjects;
     public readonly description = ProjectCreatorString.autoFindProjectsDescription;
 
+    supportsQuickCreate = true;
+
     constructor(private readonly pm: PythonProjectManager) {}
 
     async create(_options?: PythonProjectCreatorOptions): Promise<PythonProject | PythonProject[] | undefined> {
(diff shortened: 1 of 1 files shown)


src/features/creators/autoFindProjects.ts
```diff
@@ -54,10 +54,12 @@ export class AutoFindProjects implements PythonProjectCreator {
     public readonly displayName = ProjectCreatorString.autoFindProjects;
     public readonly description = ProjectCreatorString.autoFindProjectsDescription;
 
+    supportsQuickCreate = true;
+
     constructor(private readonly pm: PythonProjectManager) {}
 
     async create(_options?: PythonProjectCreatorOptions): Promise<PythonProject | PythonProject[] | undefined> {
-        const files = await findFiles('**/{pyproject.toml,setup.py}');
+        const files = await findFiles('**/{pyproject.toml,setup.py}', '**/.venv/**');
```
karthiknadig, line 62: Actually, we need to use the Prefix path for the environment. This is not always good enough. I think for now `.venv` is ok, but really we should use the sysPrefix path for environment for the workspace and ignore all `pyproject.toml` or `setup.py` from there.

karthiknadig (approved):

src/features/creators/autoFindProjects.ts
eleanorjboyd: created https://github.com/microsoft/vscode-python-environments/issues/410

PR #407, opened 2025-05-14
Updates to display rich and localized tool messages

DonJayamanne:


package.json
```diff
@@ -514,7 +514,9 @@
                 "displayName": "Get Python Environment Information",
                 "modelDescription": "Provides details about the Python environment for a specified file or workspace, including environment type, Python version, run command, and installed packages with their versions. Use this tool to determine the correct command for executing Python code in this workspace.",
                 "toolReferenceName": "pythonGetEnvironmentInfo",
-                "tags": [],
+                "tags": [
+                    "ms-python.python"
```
DonJayamanne, line 518: To be used elsewhere.

src/features/copilotTools.ts
```diff
@@ -242,13 +243,46 @@ export class InstallPackageTool implements LanguageModelTool<IInstallPackageInpu
         options: LanguageModelToolInvocationPrepareOptions<IInstallPackageInput>,
         _token: CancellationToken,
     ): Promise<PreparedToolInvocation> {
-        const packageList = options.input.packageList || [];
-        const packageCount = packageList.length;
-        const packageText = packageCount === 1 ? 'package' : 'packages';
-        const message = `Preparing to install Python ${packageText}: ${packageList.join(', ')}...`;
```
DonJayamanne, line 248: @eleanorjboyd we should avoid the use of `...`

package.json
```diff
@@ -533,9 +535,12 @@
             {
                 "name": "python_install_package",
                 "displayName": "Install Python Package",
+                "userDescription": "Installs Python packages in the given workspace",
```
eleanorjboyd, line 538: wonder if its worth clarifying that it isn't the workspace but instead installed in the environment connected to the resource provided as an arg? Or do we think that will be too confusing? (also noticing the modelDescription is similar so this is likely a mistake I made earlier

DonJayamanne: We don't do that for the model? hence not sure how useful it would be.
I would have thought more useful messages/info should be for the model vs user.
Are you suggesting we change this for the model as well.

As for user, I'm not sure it will make much sense to the user. I'm assuming the user would expect that behaviour out of the box.

karthiknadig (approved):

PR #504, opened 2025-06-05
bug: support user aborting env deletion via modal venv

eleanorjboyd:
fixes https://github.com/microsoft/vscode-python-environments/issues/505


Tyriar (dismissed):
@eleanorjboyd if this just fixes that one sub-problem, we should have a separate issue so we can verify the problem for May.

joshspicer (approved):

PR #531, opened 2025-06-11
bug: update quickCreateVenv to avoid name collisions for virtual environments

eleanorjboyd:


src/managers/builtin/venvUtils.ts
```diff
@@ -483,26 +488,11 @@ export async function createPythonVenv(
...
-            nativeFinder,
-            api,
-            log,
-            manager,
-            sortedEnvs[0],
-            venvRoot,
-            path.join(venvRoot.fsPath, '.venv'),
-            { install: allPackages, uninstall: [] },
-        );
+        return quickCreateVenv(nativeFinder, api, log, manager, sortedEnvs[0], venvRoot, options.additionalPackages);
```
eleanorjboyd, line 491: @karthiknadig just want to confirm this is good here to call the other command- seems intuitive to me but wasn't sure if there was a reason these were left separate

karthiknadig: As long as it is able to get the packages that it needs to install correctly.

eleanorjboyd: yep confirmed that works

PR #549, opened 2025-06-16
fix: debounce UI updates to python panel

eleanorjboyd:
huge amount of calls to reload the tree view for the environments and projects view for the python panel during activation. Want to add in a debounce so all the UI rendering calls get de-duped


src/extension.ts
```diff
@@ -117,8 +118,9 @@ async function collectEnvironmentInfo(
                 info.push(`  ... and ${allEnvironments.length - 10} more environments`);
             }
         }
-        
+
         // Python projects
+        console.log('getProjects called from extension.ts activate');
```
rzhao271, line 123: Extra console log?

TylerLeonhardt (dismissed):

src/extension.ts
eleanorjboyd: oops yes sorry

Issue #625, opened 2025-07-22
Loading a .env file into a terminal never happens.


mhcwainwri:
I have had a settings.json with these three lines for the better part of the last year:

```
  "python.envFile": "${workspaceFolder}/deploy/.env",
  "python.terminal.activateEnvironment": true,    
  "python.terminal.activateEnvInCurrentTerminal": true,
```

This has always worked but now is not loading as a result of this new update to the python environment manager. This kills my tasks.json which rely on the .env file I load into each terminal.

Was working yesterday and now doesn't work after the update to this extension last night.

eleanorjboyd:
Hi! Sorry about this- I am working on a fix now. See discussion on: https://github.com/microsoft/vscode-python/issues/25294#issuecomment-3103557405

eleanorjboyd:
Hi! You should be able to use the setting "python.terminal.useEnvFile" to ensure the .env file loads. The default value of this setting is false- so you need to enable using the env file at this point. Please follow up if there is a bug and this setting doesn't work

PR #626, opened 2025-07-22
Inject Environment Variables on Terminal Create

eleanorjboyd:
fixes: https://github.com/microsoft/vscode-python-environments/issues/625

adds setting `python-envs.injectEnvVarsInTerminals` with default value `false` to set if envVars will be injected

added call to inject env vars into ONLY create terminal function done through the environments extension

### DIFF
--- a/package.nls.json
+++ b/package.nls.json
@@ -1,6 +1,6 @@
 {
     "python-envs.defaultEnvManager.description": "The default environment manager for creating and managing environments.",
-    "python-envs.injectEnvVarsInTerminals.description": "This setting defaults to false and toggles if environment variables found in the workspace are set for new terminals.",
+    "python-envs.injectEnvVarsInTerminals.description": "Whether to inject environment variables into terminals on creating new terminals through the environments extension, defaults to false.",
     "python-envs.defaultPackageManager.description": "The default package manager for installing packages in environments.",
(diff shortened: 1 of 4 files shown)


package.nls.json
```diff
@@ -1,5 +1,6 @@
 {
     "python-envs.defaultEnvManager.description": "The default environment manager for creating and managing environments.",
+    "python-envs.injectEnvVarsInTerminals.description": "This setting defaults to false and toggles if environment variables found in the workspace are set for new terminals.",
```
eleanorjboyd: @cwebster-99 interested your thoughts here on the name of the setting and the description. Would appreciate your rewrite since this is not very good

src/features/terminal/terminalManager.ts
```diff
@@ -369,6 +395,12 @@ export class TerminalManagerImpl implements TerminalManager {
     }
 
     public async initialize(api: PythonEnvironmentApi): Promise<void> {
+        for (const t of terminals()) {
```
eleanorjboyd: @karthiknadig confirming it makes sense to go through all the open terminals to add these. my two questions:
- confirming this is desired behavior
- checking if there is a better way to do it (/ if I should rework the function) to be more efficient

package.nls.json
eleanorjboyd: hold on this- rewriting implementation

PR #652, opened 2025-07-25
Attach & for shell integration enabled pwsh case

anthonykim1:
Resolves https://github.com/microsoft/vscode-python-environments/issues/649

### DIFF
--- a/src/features/terminal/runInTerminal.ts
+++ b/src/features/terminal/runInTerminal.ts
@@ -28,9 +28,17 @@ export async function runInTerminal(
                 deferred.resolve();
             }
         });
+
+        const shouldSurroundWithQuotes =
+            executable.includes(' ') && !executable.startsWith('"') && !executable.endsWith('"');
+        // Handle case where executable contains white-spaces.
+        if (shouldSurroundWithQuotes) {
+            executable = `"${executable}"`;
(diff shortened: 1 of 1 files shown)


src/features/terminal/runInTerminal.ts
```diff
@@ -29,10 +28,14 @@ export async function runInTerminal(
                 deferred.resolve();
             }
         });
+        if (shellType === ShellConstants.PWSH && !executable.startsWith('&')) {
+            executable = quoteArgs([executable, ...allArgs]).join(' ');
```
eleanorjboyd: we would only want to quote the executable here and not concat it with the args since the args are passed in separately to the terminal.shellIntegration.executeCommand

src/features/terminal/runInTerminal.ts
```diff
@@ -29,10 +28,21 @@ export async function runInTerminal(
                 deferred.resolve();
             }
         });
+
+        const shouldSurroundWithQuotes =
```
anthonykim1, line 32: wrapping executable in double quotes after checking if there is white space.
Also checking they are not already wrapped in double quotes.

eleanorjboyd (approved):

Issue #635, opened 2025-07-24
Terminal hanging with `autoActivationType` set to `shellStartup`


renan-r-santos:
Hi, I'm working on an extension adding support for Pixi. I'm having trouble with the `shellStartup` option, and I was hoping to have some help debugging it. When the setting is either `command` or `off`, the terminal works fine, though.

Here's what I see when I try to open a zsh terminal on macOS:

```
WARNING: Did not detect successful shell initialization within 3 second(s).
         Please check on https://pixi.sh/latest/advanced/pixi_shell/#issues-with-pixi-shell for more tips.
rrodrigues/.ssh/id_rsa (rrodrigues@RRODRIGUES-MAC)
WARNING: Did not detect successful shell initialization within 3 second(s).
         Please check on https://pixi.sh/latest/advanced/pixi_shell/#issues-with-pixi-shell for more tips.
rrodrigues/.ssh/id_rsa (rrodrigues@RRODRIGUES-MAC)
WARNING: Did not detect successful shell initialization within 3 second(s).
         Please check on https://pixi.sh/latest/advanced/pixi_shell/#issues-with-pixi-shell for more tips.
rrodrigues/.ssh/id_rsa (rrodrigues@RRODRIGUES-MAC)
WARNING: Did not detect successful shell initialization within 3 second(s).
         Please check on https://pixi.sh/latest/advanced/pixi_shell/#issues-with-pixi-shell for more tips.
rrodrigues/.ssh/id_rsa (rrodrigues@RRODRIGUES-MAC)
```

This keeps going in a loop until I Ctrl+C it multiple times. My `execInfo` is set to this:
```typescript
                execInfo: {
                    run: { executable },
                    activatedRun: {
                        executable: pixi,
                        args: ['run', '--manifest-path', manifestPath, '-e', pixiEnv.name, 'python'],
                    },
                    activation: [
                        {
                            executable: pixi,
                            args: ['shell', '--manifest-path', manifestPath, '-e', pixiEnv.name],
                        },
                    ],
                    deactivation: [
                        {
                            executable: 'exit',
                            args: [],
                        },
                    ],
                },
```

And with log level set to `Debug` for all extensions, the only log I see for `vscode-python-environments` is not very helpful:
```
2025-07-23 22:51:35.257 [debug] Shell startup is supported for zsh, using shell startup activation
2025-07-23 22:51:35.257 [debug] Shell profile for zsh is already setup.
[...]

PR #663, opened 2025-07-28
fix: add guard for shell script re-run

karthiknadig:
Fix for #635


eleanorjboyd:
were you thinking of doing this for all shells or just these 3?

karthiknadig:
This does it for all the shells we support I think. `cmd` already had it. So I added the same for `sh/bash/zsh` (this is the bash change), fish, and powershell.

eleanorjboyd:
sweet- can I merge this in then or did you want to keep it as a draft?

Issue #662, opened 2025-07-28
does not respect "python.terminal.activateEnvironment": false config


alita-moore:
it's activating environments in my terminals despite my setting "python.terminal.activateEnvironment": false

karthiknadig:
@alita-moore as a work around can you set, `python-envs.terminal.autoActivationType` to `off`.

alita-moore:
I disabled the extension. I'll check back up on it in a few months. The python envy breaking is a deal breaker

eleanorjboyd:
Hello! The fix for respecting python.terminal.activateEnvironment is now on the newest pre-release of the environments extension. We have not gotten to python envy analysis but this first issue is fixed. Thanks

PR #665, opened 2025-07-29
require python envs ext respecting terminal.activateEnvironment

eleanorjboyd:
fixes https://github.com/microsoft/vscode-python-environments/issues/662

only follow disablement if both settings are set


eleanorjboyd:
This is how it should work with this update- are we happy with this design?

| "python.terminal.activateEnvironment"    | python-envs.terminal.autoActivationType | Will auto-activate? |
| -------- | ------- | ------- |
| false 🚫   |  off  🚫 | 🚫  |
| false 🚫 |  shellStartup or Command ✅   | shellStartup or Command ✅ |
| true ✅    | off  🚫  | 🚫 |
| true ✅ |  shellStartup or Command  ✅  | shellStartup or Command ✅ |
| unset 🟡  | off 🚫 | 🚫  |
| unset 🟡 | shellStartup or Command  ✅  | shellStartup or Command ✅ |
| false 🚫 |  unset  🟡 (update this to false)  | 🚫  |
| true ✅    | unset 🟡  | ✅ |

eleanorjboyd:
update table to reflect new system: Priority goes to the new setting  if explicitly set but otherwise we fall back to old setting. If old setting is "off" we update the new setting to reflect that user preference going forward

karthiknadig (approved):

PR #671, opened 2025-07-30
Check profiles for conda init before adding conda sourcing

eleanorjboyd:
add support so all shell profiles are checked for conda init flag before adding the sourcing script the activation commands for the environment extension. If user already has conda init in their profile then the sourcing script is redundant and introduces more opportunity for errors


eleanorjboyd:
curious initial thoughts on something like this and if this would be the right place / way to do it. Will refine more if we want to move forward

Issue #697, opened 2025-08-07
bug: Warning Message is Parsed as a Package


sjsikora:
When the extension tries to read all packages installed in an environment, it runs the pip list command and parses the output. However, I encountered an issue where a warning message about my pip version was parsed incorrectly as a package in my environment.

<img width="1293" height="332" alt="Image" src="https://github.com/user-attachments/assets/f8817ffa-ac9a-47a9-a1de-6ea18a6e973e" />

Python Environments Version: 1.2.0
VS Code Version: 1.102.1
OS: Mac, M1 Chip

eleanorjboyd:
Thank you for contributing to our repository! Your issue has been merged and will be ready to test tomorrow morning on the newest release of vscode pre-release. If you are able to confirm it works in your use case that would be great. Thanks again!

sjsikora:
Yep! Confirmed in my use case for extension pre-release version 1.3.12251014.

<img width="1189" height="255" alt="Image" src="https://github.com/user-attachments/assets/74f47089-c65d-4b26-9a71-a1e40f6ad768" />

PR #698, opened 2025-08-07
bug fix: Stricter pip list Package Parsing

sjsikora:
Addressing bug #697.

When parsing the output of pip list for packages, two checks were added. First, does the line contain exactly two space-separated keywords? The assumption being we will exactly see "[package name] [package version]" in a valid package list. Secondly, it uses the [regex in the PEP 440 docs](https://peps.python.org/pep-0440/#appendix-b-parsing-version-strings-with-regular-expressions) to verify a valid pip version. Adding these checks will help ensure we only parse valid packages and don't pick up any other information, such as warning messages. Lastly, added a unit test for the situation described in the bug.

### DIFF
--- a/src/managers/builtin/pipListUtils.ts
+++ b/src/managers/builtin/pipListUtils.ts
@@ -4,6 +4,11 @@ export interface PipPackage {
     displayName: string;
     description: string;
 }
+export function isValidVersion(version: string): boolean {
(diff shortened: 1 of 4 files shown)


sjsikora:

karthiknadig (approved):

PR #717, opened 2025-08-13
Adapt to evaluating activate in core shell integration

anthonykim1:
Issue related: https://github.com/microsoft/vscode/issues/259637
Adapting to: https://github.com/microsoft/vscode/pull/261094

FYI if activation fails when eval from shell integration scripts, we're making it very obvious for them:  https://github.com/microsoft/vscode/pull/261094#issuecomment-3182440074 

TODO: Remove activation debris in shell init files for existing users. 

/cc @cwebster-99


package.json
```diff
@@ -6,7 +6,7 @@
     "publisher": "ms-python",
     "preview": true,
     "engines": {
-        "vscode": "^1.100.0-20250407"
+        "vscode": "^1.103.0"
```
anthonykim1: Update version after https://github.com/microsoft/vscode/pull/261548 is merged.
* https://github.com/microsoft/vscode/pull/261094 got merged.
* Waiting on  https://github.com/microsoft/vscode/pull/261548

src/features/terminal/shellStartupSetupHandlers.ts
```diff
@@ -11,9 +11,12 @@ export async function handleSettingUpShellProfile(
     callback: (provider: ShellStartupScriptProvider, result: boolean) => void,
 ): Promise<void> {
     const shells = providers.map((p) => p.shellType).join(', ');
+    // TODO: Get opinions on potentially modifying the prompt
+    //       - If shell integration is active, we won't need to modify user's shell profile, init scripts.
+    //       - Current prompt we have below may not be the most accurate description.
     const response = await showInformationMessage(
         l10n.t(
-            'To enable "{0}" activation, your shell profile(s) need to be updated to include the necessary startup scripts. Would you like to proceed with these changes?',
+            'To enable "{0}" activation, your shell profile(s) may need to be updated to include the necessary startup scripts. Would you like to proceed with these changes?',
```
anthonykim1, line 19: Need wording feedbacks on this. 
Because if activate happens from shell integration script, we will NOT need to and will NOT modify user's shell scripts.

src/features/terminal/shells/bash/bashConstants.ts
```diff
@@ -1,3 +1,3 @@
-export const BASH_ENV_KEY = 'VSCODE_BASH_ACTIVATE';
-export const ZSH_ENV_KEY = 'VSCODE_ZSH_ACTIVATE';
+export const BASH_ENV_KEY = 'VSCODE_PYTHON_BASH_ACTIVATE';
```
anthonykim1, line 1: Aligning these to match with those used in core.
Note that we specifically block other extensions from contributing value variable as its leading to code execution.

src/features/terminal/shells/common/shellUtils.ts
```diff
@@ -95,3 +97,16 @@ export function extractProfilePath(content: string): string | undefined {
...
     return undefined;
 }
+
+export function shellIntegrationForActiveTerminal(name: string, profile: string): boolean {
+    const hasShellIntegration = activeTerminalShellIntegration();
+
+    if (hasShellIntegration) {
+        traceInfo(
+            `SHELL: Shell integration is available on your active terminal.  Python activate scripts will be evaluated at shell integration level. 
+                Skipping modification of ${name} profile at: ${profile}`,
```
eleanorjboyd, line 107: this is a bit confusing- `Skipping modification of ${name} profile at: ${profile}` maybe clarifying its a shell profile idk

eleanorjboyd (dismissed):

src/features/terminal/shells/common/shellUtils.ts
anthonykim1: I think we should be fine since ${name} will be shell name itself such as bash, pwsh, etc.
We could add the word `shell` after `${name}` but for shells like powershell, it might get little redundant with `powershell shell profile.. etc`

Issue #726, opened 2025-08-15
Remove shellStartup debris from user's shell init file if shell integration is enabled


anthonykim1:
Coming from: https://github.com/microsoft/vscode-python-environments/pull/717 

For existing users, we'd want to remove the shell startup code from their shell init files since we eval from shell integration instead.

PR #727, opened 2025-08-15
Remove debris if shell integration is available

anthonykim1:
Resolves: https://github.com/microsoft/vscode-python-environments/issues/726 
Coming from https://github.com/microsoft/vscode-python-environments/pull/717 
/cc @karthiknadig

### DIFF
--- a/src/features/terminal/shells/bash/bashStartup.ts
+++ b/src/features/terminal/shells/bash/bashStartup.ts
@@ -71,10 +71,10 @@ async function isStartupSetup(profile: string, key: string): Promise<ShellSetupS
 
 async function setupStartup(profile: string, key: string, name: string): Promise<boolean> {
     if (shellIntegrationForActiveTerminal(name, profile)) {
+        removeStartup(profile, key);
         return true;
     }
     const activationContent = getActivationContent(key);
-
     try {
         if (await fs.pathExists(profile)) {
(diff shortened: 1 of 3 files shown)


src/features/terminal/shells/bash/bashStartup.ts
```diff
@@ -71,10 +71,10 @@ async function isStartupSetup(profile: string, key: string): Promise<ShellSetupS
 
 async function setupStartup(profile: string, key: string, name: string): Promise<boolean> {
     if (shellIntegrationForActiveTerminal(name, profile)) {
+        removeStartup(profile, key);
```
Tyriar, line 74: No zsh?

Tyriar (approved):

src/features/terminal/shells/bash/bashStartup.ts
anthonykim1: Yeah, it seems like right now both bash and zsh share the same file..https://github.com/microsoft/vscode-python-environments/blob/66ff5dd97bffa1e3ff7d4a5e81740eb7f36296bf/src/features/terminal/shells/bash/bashStartup.ts#L187 

I can go refactor during debt week.

anthonykim1: Just created item here: https://github.com/microsoft/vscode-python-environments/issues/728

PR #731, opened 2025-08-18
add quoting, logging, and resolution for run as task and background

eleanorjboyd:


eleanorjboyd:
@karthiknadig just wanted your review specifically to confirm the resolution calls looked right

PR #740, opened 2025-08-20
Improve shell command formatting

almarouk:
Consolidates shell command formatting logic into a shared utility function that is used for running in terminal and PET command execution, among other places.
This improves consistency and command path handling.

This should also fix PET command execution on Windows using PowerShell (requires `&`).

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -497,7 +497,7 @@ export async function activate(context: ExtensionContext): Promise<PythonEnviron
                     const execString = getShellCommandAsString(identifyTerminalShell(terminal), [
                         { executable: petPath, args: [command.subcommand, ...(command.args || [])] },
                     ]);
-                    terminal.sendText(execString, true);
+                    terminal.sendText(execString);
                     traceInfo(`Running PET ${command.subcommand} command: ${execString}`);
                 }
             } catch (error) {


src/extension.ts
```diff
@@ -435,6 +436,7 @@ export async function activate(context: ExtensionContext): Promise<PythonEnviron
         commands.registerCommand('python-envs.runPetInTerminal', async () => {
             try {
                 const petPath = await getNativePythonToolsPath();
+                let command: { subcommand: string; args?: string[] } | undefined;
```
anthonykim1, line 439: Thanks for the PR.
Looks like command will never be undefined here, we could probably remove the `| undefined`

src/extension.ts
```diff
@@ -489,8 +490,15 @@ export async function activate(context: ExtensionContext): Promise<PythonEnviron
                     }
 
                     // Run pet resolve with the provided path
-                    terminal.sendText(`"${petPath}" resolve "${inputPath.trim()}"`, true);
-                    traceInfo(`Running PET resolve command: ${petPath} resolve "${inputPath.trim()}"`);
+                    command = { subcommand: 'resolve', args: [inputPath.trim()] };
+                }
+
+                if (command) {
```
anthonykim1, line 496: `if (command) {` could also go away if we remove `undefined` from type above.

src/extension.ts
```diff
@@ -489,8 +490,15 @@ export async function activate(context: ExtensionContext): Promise<PythonEnviron
...
+                    const execString = getShellCommandAsString(identifyTerminalShell(terminal), [
+                        { executable: petPath, args: [command.subcommand, ...(command.args || [])] },
+                    ]);
+                    terminal.sendText(execString, true);
```
anthonykim1: The default for `shouldExecute` is true, I believe. We could probably remove the `true`.

src/extension.ts
almarouk: `undefined` is required here since the `if ... else` tests on `selectedOption.label` below don't have a default `else`, therefore the ts compiler will complain with `ts(2454)`.

src/extension.ts
almarouk: See reply to previous comment.

src/features/execution/execUtils.ts
```diff
@@ -1,4 +1,5 @@
 export function quoteStringIfNecessary(arg: string): string {
+    arg = arg.trim();
```
anthonykim1, line 2: I'm not sure if we really need this. Seems like the only places we call this are:

https://github.com/microsoft/vscode-python-environments/blob/e0eedd37e041d438a69aceaa823dbd2267c6c540/src/features/execution/runAsTask.ts#L29-L35 
and 
https://github.com/microsoft/vscode-python-environments/blob/e0eedd37e041d438a69aceaa823dbd2267c6c540/src/features/execution/runInBackground.ts#L10-L16

And the executable string that is passed in seems like its path to Python binary.

src/extension.ts
almarouk: That's correct but I just kept the syntax from before my introduced changes. Will add a commit to remove the `true`.

src/extension.ts
anthonykim1: Now I'm thinking can we just simplify the second `else if (...Resolve Environment)` into `else`? 
There is only two options that `selectionOption.label` can take.

src/features/execution/execUtils.ts
almarouk: Here:

https://github.com/microsoft/vscode-python-environments/blob/e0eedd37e041d438a69aceaa823dbd2267c6c540/src/extension.ts#L492

`inputPath` is user-provided and could have leading white spaces. I did keep the `.trim()` in this case before eventually processing the string with `getShellCommandAsString` but I thought also of adding it in the quoting logic for consistency and  robustness in future use cases. What do you think?

src/extension.ts
almarouk: That's true but I didn't want to introduce any unrelated changes with possibly undesirable effects, keeping the PR context relevant and minimal. Should I still change it?

anthonykim1: That's fine, we can leave it as else if, in case there is additional label in future. We should probably return early in case command is ever undefined though and log. 

If you do this, we won't need `if (command) { ..`

src/features/execution/execUtils.ts
anthonykim1: looks like `inputPath.trim()` is called right before `getShellCommandAsString` from your PR: 
https://github.com/microsoft/vscode-python-environments/blob/0390029e87943e9af66adaaf2e97b1295c696e43/src/extension.ts#L493

anthonykim1:
@almarouk Sorry things changed quite a bit. Could you resolve conflicts or do you want to start it off in recent branch?

eleanorjboyd:
closing as it has been a while without any activity but please comment and I can re-open

PR #751, opened 2025-08-22
fix conda env refresh not waiting for promises

almarouk:
This was causing the refresh function to return with empty or missing environments before waiting for all async calls that update the env collection.

This bug was somehow causing duplicate environments to appear in the list of environments.


eleanorjboyd:
great catch! Thank you for adding this and sorry for the delay

PR #756, opened 2025-08-23
Log spawning of processes

DonJayamanne:
For https://github.com/microsoft/vscode/issues/266411

### DIFF
--- a/src/features/terminal/shells/cmd/cmdStartup.ts
+++ b/src/features/terminal/shells/cmd/cmdStartup.ts
@@ -10,8 +10,14 @@ import { ShellConstants } from '../../../common/shellConstants';
 import { hasStartupCode, insertStartupCode, removeStartupCode } from '../common/editUtils';
 import { ShellScriptEditState, ShellSetupState, ShellStartupScriptProvider } from '../startupProvider';
 import { CMD_ENV_KEY, CMD_SCRIPT_VERSION } from './cmdConstants';
+import { StopWatch } from '../../../../common/stopWatch';
 
-const exec = promisify(cp.exec);
+function execCommand(command: string) {
+    const timer = new StopWatch();
+    return promisify(cp.exec)(command, { windowsHide: true }).finally(() =>
+        traceInfo(`Executed command: ${command} in ${timer.elapsedTime}`),
+    );
+}
 
 async function isCmdInstalled(): Promise<boolean> {
     if (!isWindows()) {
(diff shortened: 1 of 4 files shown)


src/features/terminal/shells/utils.ts
```diff
@@ -1,13 +1,16 @@
...
+import { traceError, traceInfo } from '../../../common/logging';
+import { StopWatch } from '../../../common/stopWatch';
 
 export async function runCommand(command: string): Promise<string | undefined> {
     return new Promise((resolve) => {
+        const timer = new StopWatch();
         cp.exec(command, (err, stdout) => {
             if (err) {
-                traceVerbose(`Error running command: ${command}`, err);
+                traceError(`Error running command: ${command} (${timer.elapsedTime})`, err);
```
DonJayamanne, line 10: Surely if success is `info`, then failure must be warning/error or info, but not something less

src/managers/builtin/helpers.ts
```diff
@@ -9,7 +9,7 @@ export async function isUvInstalled(log?: LogOutputChannel): Promise<boolean> {
     if (available.completed) {
         return available.promise;
     }
-
+    log?.info(`Running: uv --version`);
```
DonJayamanne, line 12: I don't think the code is right here, as `log` is optional 
And looking at the calling code, it isn't passed

Not sure why, I'd try to resolve this in debt, as logger shouldn't be optional. we either turn off logging or the like, but wouldn't make the logger component optional.

src/managers/conda/condaUtils.ts
```diff
@@ -202,6 +203,8 @@ async function _runConda(
 ): Promise<string> {
     const deferred = createDeferred<string>();
     args = quoteArgs(args);
+    const timer = new StopWatch();
+    deferred.promise.finally(() => traceInfo(`Ran conda in ${timer.elapsedTime}: ${conda} ${args.join(' ')}`));
```
DonJayamanne, line 207: We haven't logged this at all, now we will.
We know conda is slow

src/features/terminal/shells/cmd/cmdStartup.ts
```diff
@@ -10,8 +10,14 @@ import { ShellConstants } from '../../../common/shellConstants';
 import { hasStartupCode, insertStartupCode, removeStartupCode } from '../common/editUtils';
 import { ShellScriptEditState, ShellSetupState, ShellStartupScriptProvider } from '../startupProvider';
 import { CMD_ENV_KEY, CMD_SCRIPT_VERSION } from './cmdConstants';
+import { StopWatch } from '../../../../common/stopWatch';
 
-const exec = promisify(cp.exec);
+function execCommand(command: string) {
+    const timer = new StopWatch();
+    return promisify(cp.exec)(command, { windowsHide: true }).finally(() =>
```
DonJayamanne, line 17: We haven't logged this at all, now we will.
I believe we've seen delays in reading registry in the past as well.

PR #787, opened 2025-09-04
bug: surface warning for broken defaultEnvManager and resolve

eleanorjboyd:
fixes https://github.com/microsoft/vscode-python/issues/25430

image

### DIFF
--- a/src/features/settings/settingHelpers.ts
+++ b/src/features/settings/settingHelpers.ts
@@ -9,7 +9,7 @@ import {
 } from 'vscode';
 import { PythonProject } from '../../api';
 import { DEFAULT_ENV_MANAGER_ID, DEFAULT_PACKAGE_MANAGER_ID } from '../../common/constants';
-import { traceError, traceWarn } from '../../common/logging';
+import { traceError, traceInfo, traceWarn } from '../../common/logging';
 import { getWorkspaceFile, getWorkspaceFolders } from '../../common/workspace.apis';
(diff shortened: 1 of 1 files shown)


src/managers/common/utils.ts
```diff
@@ -194,3 +197,41 @@ export async function getShellActivationCommands(binDir: string): Promise<{
...
+    projectManager: PythonProjectManager,
+    api: PythonEnvironmentApi,
+) {
+    const defaultEnvManager = getDefaultEnvManagerSetting(projectManager);
+    if (defaultEnvManager === managerId) {
+        setDefaultEnvManagerBroken(true);
+        await api.refreshEnvironments(undefined);
+        window
+            .showErrorMessage(
+                `The default environment manager is set to '${defaultEnvManager}', but the ${
```
DonJayamanne: We need to localize the messages.

src/managers/common/utils.ts
```diff
@@ -194,3 +197,41 @@ export async function getShellActivationCommands(binDir: string): Promise<{
         shellDeactivation,
     };
 }
+
+export async function notifyMissingManagerIfDefault(
+    managerId: string,
+    projectManager: PythonProjectManager,
+    api: PythonEnvironmentApi,
+) {
+    const defaultEnvManager = getDefaultEnvManagerSetting(projectManager);
```
DonJayamanne, line 220: Does this notification get displayed just once? We should display this error only once.

I dont see any variable or the like that tracks whether this was already displayed once before.

PR #818, opened 2025-09-09
sort by version (desc) in conda create picker

eleanorjboyd:


src/managers/conda/condaUtils.ts
```diff
@@ -755,11 +754,16 @@ export async function pickPythonVersion(
...
-        .reverse();
-    if (!versions) {
-        versions = ['3.11', '3.10', '3.9', '3.12', '3.13'];
+    );
+
+    // Sort versions by major version (descending), ignoring minor/patch for simplicity
+    versions = versions.sort((a, b) => {
+        const getMajor = (v: string) => parseInt(v.split('.')[0], 10);
+        return getMajor(b) - getMajor(a);
+    });
```
Tyriar: Nicer and avoids maybe rebuilding the function:

```suggestion
    const getMajor = (v: string) => parseInt(v.split('.')[0]);
    versions = versions.sort((a, b) => getMajor(b) - getMajor(a));
```

10 on `parseInt` used to be required but don't think that's true anymore

src/managers/conda/condaUtils.ts
```diff
@@ -755,11 +754,16 @@ export async function pickPythonVersion(
...
-    )
-        .sort()
-        .reverse();
-    if (!versions) {
-        versions = ['3.11', '3.10', '3.9', '3.12', '3.13'];
+    );
+
+    // Sort versions by major version (descending), ignoring minor/patch for simplicity
+    versions = versions.sort((a, b) => {
+        const getMajor = (v: string) => parseInt(v.split('.')[0], 10);
```
Tyriar: Not sure how safe this needs to be but a regex would be safer: `match(/(?<major>\d+)\./).groups?.major`

Issue #765, opened 2025-08-26
Wrong activate button state when using `shellStartup`


renan-r-santos:
When opening a new terminal while having the `shellStartup` activation mode selected, it appears that the state of the activate button isn't updated to reflect that the Python environment was activated. With activation mode set to `command`, you can see in the video below that the button is correctly displaying the `deactivate` option.

This is important because not all activation mechanisms are idempotent, e.g. https://pixi.sh/latest/advanced/pixi_shell/.

https://github.com/user-attachments/assets/97320473-a32c-4fad-b530-779259c5f6e6

eleanorjboyd:
Thank you for collecting this bug report- will review when I am back from vacation, thanks!

eleanorjboyd:
<img width="1143" height="309" alt="Image" src="https://github.com/user-attachments/assets/5b59b113-1316-4af9-83a9-f3c372e4bc06" />

given shellStartup, if I click the "deactivate" button it switches the button "activate" afterwords but then reverts to "deactivate" I think this is because the terminal is considered activated because it is shellStartup even though that specific terminal was deactivated. 

I can take a look in a bit although I don't know at this point what the fix for that part would look like

renan-r-santos:
Hi @eleanorjboyd, does this happen with the code changes in my https://github.com/microsoft/vscode-python-environments/pull/837 PR? I tried it over here, and it seems to work well. If you have a way to reproduce this, I'd be happy to take a look.

PR #837, opened 2025-09-13
Fix activation icon state when using `shellStartup`

renan-r-santos:
Fixes https://github.com/microsoft/vscode-python-environments/issues/765


anthonykim1:
https://github.com/user-attachments/assets/2aac33c2-62f2-4bc4-93dc-d5a54e475f40

Seems to work pretty good :) Thank you for the PR.
One thing I noticed was the slight flickering when the activation type was command, but thats probably because it takes time for ENV var to be set, and extension to confirm terminal is activated (since command type activation is not fast as shellStartup).

PR #863, opened 2025-09-24
implement MVP of workspace and global searchPaths

eleanorjboyd:


src/managers/common/nativePythonFinder.ts
```diff
@@ -326,10 +326,12 @@ class NativePythonFinderImpl implements NativePythonFinder {
      * Must be invoked when ever there are changes to any data related to the configuration details.
      */
     private async configure() {
+        // Get all extra search paths including legacy settings and new searchPaths
+        const extraSearchPaths = await getAllExtraSearchPaths();
+
         const options: ConfigurationOptions = {
             workspaceDirectories: this.api.getPythonProjects().map((item) => item.uri.fsPath),
-            // We do not want to mix this with `search_paths`
```
eleanorjboyd, line 331: not sure about this comment @karthiknadig - seems contrary to what I am using it for

src/managers/common/nativePythonFinder.ts
```diff
@@ -380,6 +382,193 @@ function getPythonSettingAndUntildify<T>(name: string, scope?: Uri): T | undefin
...
+/**
+ * Gets the most specific workspace-level setting available for workspaceSearchPaths.
+ */
+function getWorkspaceSearchPaths(): string[] {
+    try {
+        const envConfig = getConfiguration('python-env');
+        const inspection = envConfig.inspect<string[]>('workspaceSearchPaths');
+
+        if (inspection?.globalValue) {
+            traceLog(
```
eleanorjboyd: @karthiknadig any smarter way to handle this do you think? since we don't want it set globally

src/managers/common/nativePythonFinder.ts
karthiknadig: This was because it was easier to classify using known environment directories.

karthiknadig: But I don’t think we need to bind by that.

src/managers/common/nativePythonFinder.ts
```diff
@@ -380,6 +382,193 @@ function getPythonSettingAndUntildify<T>(name: string, scope?: Uri): T | undefin
...
+/**
+ * Gets all extra environment search paths from various configuration sources.
+ * Combines legacy python settings (with migration), globalSearchPaths, and workspaceSearchPaths.
+ * @returns Array of search directory paths
+ */
+export async function getAllExtraSearchPaths(): Promise<string[]> {
+    const searchDirectories: string[] = [];
+
+    // Handle migration from legacy python settings to new search paths settings
+    const legacyPathsCovered = await handleLegacyPythonSettingsMigration();
```
karthiknadig: I don’t think we need to migrate yet. For MVP, all we should do is consolidate to a single list.

src/managers/common/nativePythonFinder.ts
```diff
@@ -380,6 +382,193 @@ function getPythonSettingAndUntildify<T>(name: string, scope?: Uri): T | undefin
...
+        return [];
+    }
+}
+
+/**
+ * Applies untildify to an array of paths
+ * @param paths Array of potentially tilde-containing paths
+ * @returns Array of expanded paths
+ */
+function untildifyArray(paths: string[]): string[] {
```
karthiknadig: This should go into string utils

karthiknadig (changes requested):

package.json
```diff
@@ -109,6 +109,24 @@
                     "description": "%python-envs.terminal.useEnvFile.description%",
                     "default": false,
                     "scope": "resource"
+                },
+                "python-env.globalSearchPaths": {
+                    "type": "array",
+                    "description": "%python-env.globalSearchPaths.description%",
+                    "default": [],
+                    "scope": "application",
```
karthiknadig: This should be "machine", to allow setting this for remote VS Code.

karthiknadig (changes requested):

karthiknadig (approved):

Issue #883, opened 2025-10-02
Environment configuration change warning shows up even if there is no change in environment


karthiknadig:
<img width="683" height="351" alt="Image" src="https://github.com/user-attachments/assets/2dfcd9d7-9c98-47c0-bd35-ec031a6e1cfb" />

PR #886, opened 2025-10-02
fix env var clearing

eleanorjboyd:
fixes https://github.com/microsoft/vscode-python-environments/issues/883


src/features/terminal/terminalEnvVarInjector.ts
```diff
@@ -110,7 +110,6 @@ export class TerminalEnvVarInjector implements Disposable {
                 await this.injectEnvironmentVariablesForWorkspace(workspaceFolder);
             } else {
                 // No provided workspace - update all workspaces
-                this.envVarCollection.clear();
```
anthonykim1, line 113: I dont know about this one though, I think we should still clear if in no workspace scenario.

PR #911, opened 2025-10-08
support pipenv path setting

eleanorjboyd:


src/managers/pipenv/pipenvUtils.ts
```diff
@@ -46,6 +47,12 @@ export async function clearPipenvCache(): Promise<void> {
     pipenvPath = undefined;
 }
 
+function getPipenvPathFromSettings(): Uri[] {
+    // python.pipenvPath is a workspace-scoped setting
```
Tyriar: Seems redundant?

eleanorjboyd: thank you- great catch

Tyriar (approved):

PR #928, opened 2025-10-10
Bug in create venv flow with selecting skipping package installation

eleanorjboyd:
fixes https://github.com/microsoft/vscode-python-environments/issues/930


src/managers/builtin/pipUtils.ts
```diff
@@ -129,7 +129,7 @@ async function selectWorkspaceOrCommon(
                 return await selectFromCommonPackagesToInstall(common, installed, undefined, { showBackButton });
             } else if (selected.label === PackageManagement.skipPackageInstallation) {
                 traceInfo('Package Installer: user selected skip package installation');
-                return undefined;
+                return { install: [], uninstall: [] } as PipPackages;
```
amunger:
```suggestion
                return { install: [], uninstall: [] } satisfies PipPackages;
```

amunger: provides type safety, (`as` does not)

PR #952, opened 2025-10-24
Fix unexpected blocking for pwsh command activation 

anthonykim1:
Resolves: https://github.com/microsoft/vscode/issues/273017 

* Coming from: https://github.com/microsoft/vscode/pull/272369 
* Further ensure so that pwsh command activation does not hang.
* Took the regex from: https://github.com/microsoft/vscode/blob/e1f78f13a11e6a1f20dfacbe03f8c7b64b031ac2/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/executeStrategy/executeStrategy.ts#L53-L97 
* Already updated distro to enable onTerminalDataWrite proposed api.

### DIFF
--- a/src/common/utils/asyncUtils.ts
+++ b/src/common/utils/asyncUtils.ts
@@ -1,5 +1,3 @@
 export async function timeout(milliseconds: number): Promise<void> {
     return new Promise<void>((resolve) => setTimeout(resolve, milliseconds));
 }
-
-// TODO: Advanced timeout from core async: https://github.com/microsoft/vscode-python-environments/issues/953


package.json
```diff
@@ -12,7 +12,8 @@
         "Other"
     ],
     "enabledApiProposals": [
-        "terminalShellEnv"
+        "terminalShellEnv",
+        "terminalDataWriteEvent"
```
karthiknadig, line 16: This will need an engine version update if the proposed API is not in stable, and a distro change.

anthonykim1: I believe `terminalDataWriteEvent`is pretty old proposed api. 

Also updated the distro for env extension to enable this, and updated vscode repo with the new hash: https://github.com/microsoft/vscode/pull/273148.

I'd want to  pin vscode version to one from this morning to make sure we have https://github.com/microsoft/vscode/pull/272369 in though  --> done see commit below

Tyriar: Yep it's old

src/features/terminal/utils.ts
```diff
@@ -1,19 +1,107 @@
...
+
+    // Custom prompts ending with the starship character (\u276f)
+    if (/\u276f\s*/.test(cursorLine)) {
+        return true;
+    }
+
+    // Generic prompts ending with common prompt characters
+    if (/[>%]\s*/.test(cursorLine)) {
+        return true;
+    }
```
Tyriar, line 118: Without the `$` these are basically just checking if a single character is in the string

```ts
/#\s*/.test(cursorLine)
```

Is equivalent to:

```ts
/#/.test(cursorLine)
```

Or

```ts
cursorLine.includes('#')
```

That's probably fine for the most part as often as soon as one of these characters show up the terminal will be ready, there may be cases where that's no true though.

src/features/terminal/utils.ts
```diff
@@ -1,19 +1,107 @@
...
+        ]);
+
+        return result;
+    } finally {
+        disposables.forEach((d) => d.dispose());
+    }
+}
+
+// Detects if the given text content appears to end with a common prompt pattern.
+function detectsCommonPromptPattern(cursorLine: string): boolean {
```
Tyriar: `cursorLine` isn't correct here

src/features/terminal/utils.ts
Tyriar: You should be able to bring the $ and those checks back by running the data through `removeAnsiEscapeCodes`:

https://github.com/microsoft/vscode/blob/3446d06b3361d7e41b959285d130c9bdde630059/src/vs/base/common/strings.ts#L808-L822

src/features/terminal/utils.ts
```diff
@@ -1,19 +1,107 @@
...
+    const timeoutMs = timeoutValue === undefined || -1 ? 5000 : timeoutValue;
+
+    const disposables: Disposable[] = [];
+
+    try {
+        const result = await Promise.race([
+            // Condition 1: Shell integration timeout setting
+            new Promise<boolean>((resolve) => {
+                setTimeout(() => resolve(false), timeoutMs);
+            }),
```
Tyriar: In core you would do this simpler like this:

```ts
timeout(timeoutMs).then(() => false),
```

src/features/terminal/utils.ts
Tyriar: Can you debounce this function to 50ms or so?

package.json
```diff
@@ -6,13 +6,14 @@
     "publisher": "ms-python",
     "preview": true,
     "engines": {
-        "vscode": "^1.104.0-20250815"
+        "vscode": "^1.104.0-20251024"
```
Tyriar: Shouldn't need to do this, I think the only consequence of it is some Insiders won't get it until after they update.

src/features/terminal/utils.ts
anthonykim1: I dont think we have nice decorator like in core yet  but I think [b4579f3](https://github.com/microsoft/vscode-python-environments/pull/952/commits/b4579f3455431f9dec5cdbdd0936adb135542ffe) should still do it

src/features/terminal/utils.ts
anthonykim1: This is pure gold 🚀 
https://github.com/microsoft/vscode-python-environments/pull/952/commits/a3feed7c10fb300abedcdfd7aab731d14b25103b

src/common/utils/asyncUtils.ts
```diff
@@ -1,3 +1,9 @@
 export async function sleep(milliseconds: number) {
     return new Promise<void>((resolve) => setTimeout(resolve, milliseconds));
 }
+
+export function timeout(ms: number): Promise<void> {
+    return new Promise((resolve) => setTimeout(resolve, ms));
+}
+
+// TODO: Advanced timeout from core async: https://github.com/microsoft/vscode-python-environments/issues/953
```
Tyriar: Looks like it's identical to `sleep`. I recommend renaming this to `timeout` so it's consistent with the core codebase since the core team works on this repo.

src/features/terminal/utils.ts
```diff
@@ -1,19 +1,115 @@
...
-    while (!terminal.shellIntegration && timeout < SHELL_INTEGRATION_TIMEOUT) {
-        await sleep(SHELL_INTEGRATION_POLL_INTERVAL);
-        timeout += SHELL_INTEGRATION_POLL_INTERVAL;
+    if (terminal.shellIntegration) {
+        return true;
+    }
+
+    const config = getConfiguration('terminal.integrated');
+    const timeoutValue = config.get<number | undefined>('shellIntegration.timeout');
+    const timeoutMs = timeoutValue === undefined || -1 ? 5000 : timeoutValue;
```
Tyriar: You can make it more consistent with `getShellIntegrationTimeout` by checking `terminal.integrated.shellIntegration.enabled` (mostly will match `siInjectionEnabled`) and the window's `remoteAuthority` for isRemote.

src/features/terminal/utils.ts
```diff
@@ -1,19 +1,115 @@
...
+                            resolve(true);
+                        }
+                    }),
+                );
+            }),
+
+            // Condition 3: Detect prompt patterns in terminal output
+            new Promise<boolean>((resolve) => {
+                let dataSoFar = '';
+                const debounced = createSimpleDebounce(50, () => {
```
Tyriar, line 55: This should be disposable, created https://github.com/microsoft/vscode-python-environments/issues/954

src/features/terminal/utils.ts
```diff
@@ -1,19 +1,115 @@
...
+                    }
+                });
+                disposables.push(
+                    onDidWriteTerminalData((e) => {
+                        if (e.terminal === terminal) {
+                            dataSoFar += e.data;
+                            debounced.trigger();
+                        }
+                    }),
+                );
```
Tyriar: nit: This would faster/more memory friendly if it was a string builder:


```suggestion
                const dataEvents: string = [];
                const debounced = createSimpleDebounce(50, () => {
                    if (dataEvents && detectsCommonPromptPattern(dataEvents.join(''))) {
                        resolve(false);
                    }
                });
                disposables.push(
                    onDidWriteTerminalData((e) => {
                        if (e.terminal === terminal) {
                            dataEvents.push(e.data);
                            debounced.trigger();
                        }
                    }),
                );
```

src/common/utils/asyncUtils.ts
anthonykim1: I got rid of the sleep and just made them async returning Promise<void> like this: 
[f64b62c](https://github.com/microsoft/vscode-python-environments/pull/952/commits/f64b62cb44db0b983d3720eecc366573a91d394e)

src/features/terminal/utils.ts
anthonykim1: Sounds good,, I think the remoteAuthority is from proposed api though: https://github.com/microsoft/vscode/blob/c23478af52b09c278e21e8396a0e120eb134ae3a/src/vscode-dts/vscode.proposed.resolvers.d.ts#L472  

Would `env.remoteName` be good enough? [28f5261](https://github.com/microsoft/vscode-python-environments/pull/952/commits/28f5261d294977bc53cd252d01b478db291914cc)

src/features/terminal/utils.ts
anthonykim1: I tried to do them: [73f3f3c](https://github.com/microsoft/vscode-python-environments/pull/952/commits/73f3f3c9ee554339d262edc41fa83a1c2e028dac)

src/common/utils/asyncUtils.ts
```diff
@@ -1,3 +1,5 @@
-export async function sleep(milliseconds: number) {
+export async function timeout(milliseconds: number): Promise<void> {
     return new Promise<void>((resolve) => setTimeout(resolve, milliseconds));
 }
+
+// TODO: Advanced timeout from core async: https://github.com/microsoft/vscode-python-environments/issues/953
```
Tyriar:
```suggestion

```

Tyriar (changes requested):

Tyriar (approved):

bpasero (approved):

PR #964, opened 2025-11-06
Do not activate for task terminal 

anthonykim1:
Resolves: https://github.com/microsoft/vscode/issues/275667 

We'd probably want to switch over to more reliable long term approach like https://github.com/microsoft/vscode/issues/234440 in the future though. /cc @meganrogge  I think we can either add the api to tell if terminal is task terminal, or just straight up block extension from contributing commands to task terminal. 

I don't know how aggressive the later would be, maybe some extension actually wants to contribute to task terminal via input?

### DIFF
--- a/src/common/tasks.apis.ts
+++ b/src/common/tasks.apis.ts
@@ -1,5 +1,13 @@
-import { Task, TaskExecution, tasks } from 'vscode';
-
+import { Disposable, Task, TaskExecution, TaskProcessStartEvent, tasks } from 'vscode';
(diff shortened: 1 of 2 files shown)


Tyriar:
Let's do the fix in core: https://github.com/microsoft/vscode/pull/275782

Tyriar:
Also ideally the python extension(s) wouldn't attempt to activate these, there's a long standing problem with the processId API though so I worry about this fix introducing hard to diagnose bugs: https://github.com/microsoft/vscode/issues/91905

PR #1054, opened 2025-12-11
Fix a bug with Quick Create not installing workspace dependencies

StellaHuang95:
There are two ways user can do `Quick Create`.

1. When user goes from `Create Virtual Environment -> Quick Create`, the workspace dependencies are installed in the newly created virtual environment.

2. When user goes from `Create Virtual Environment -> venv -> Quick Create`, the workspace dependencies are NOT installed in the newly created virtual environment. This is because the packages are not collected in the code flow. This PR adds that step back.

image

fixes https://github.com/microsoft/vscode-python-environments/issues/1080

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5035,7 +5035,8 @@
         "node_modules/tslib": {
             "version": "1.14.1",
             "resolved": "https://registry.npmjs.org/tslib/-/tslib-1.14.1.tgz",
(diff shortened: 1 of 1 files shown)


eleanorjboyd (approved):

eleanorjboyd:
great catch! Thanks!

Issue #411, opened 2025-05-14
More gracefully fail on `pyproject.toml` error


eleanorjboyd:
tried going through the "create environment..." flow with a `pyproject.toml` file but I had an error since my `pyproject.toml` file was incorrectly formatted. Would be good if the error message was better or it failed more gracefully

```
 Resolved Python Environment /Users/eleanorboyd/python-envs-scenarios/existing-env-folder/.venv/bin/python
2025-05-14 09:19:39.308 [info] Running: uv pip install --python /Users/eleanorboyd/python-envs-scenarios/existing-env-folder/.venv/bin/python -e /Users/eleanorboyd/python-envs-scenarios/existing-env-folder
2025-05-14 09:19:39.378 [info] error: Failed to extract static metadata from `pyproject.toml`
  Caused by: TOML parse error at line 6, column 8
  |
6 | name = "existing env folder"
  |        ^^^^^^^^^^^^^^^^^^^^^
Not a valid package or extra name: "existing env folder". Names must start and end with a letter or digit and may only contain -, _, ., and alphanumeric characters.


```

StellaHuang95:
I looked into the issue and found several problems with the current user experience:

1. Although the popup window shows that an error occurred, users only see a generic “Error managing packages” message with no context. It doesn’t mention `pyproject.toml` at all. The actual error is buried in the logs, and users have to click “View Output” and scroll through the output to find the real message.
<img width="505" height="120" alt="Image" src="https://github.com/user-attachments/assets/212f3e34-f01b-48d1-a8ad-3305cb546325" />


2. The environment is created successfully first, but the package installation then fails, leaving users with a half-working environment in a confusing state.

I’m thinking of improving the user experience by adding early validation. We could validate `pyproject.toml` when the user clicks `Create Environment`. If it’s invalid, we would show a warning with the error message and give the user three options: `Open pyproject.toml`, `Create anyway`, or `Cancel`, which corresponds to open the file, proceed despite the issue, or stop the process. 

One challenge is that the warning dialog has limited space, so the error message will need to be concise and easy to understand. Benefit of this approach is that once we start parsing `pyproject.toml` up front, we can also validate other formatting and configuration issues early based on relevant PEPs, making the fix more general.

@eleanorjboyd What do you think about this approach?

eleanorjboyd:
Thanks for this detailed investigation! When you say the "create environment" step which one are you talking about specifically? Like before this one?

[...]

PR #1066, opened 2025-12-15
Validate pyproject.toml before installing dependencies

StellaHuang95:
This PR implements validation for `pyproject.toml` when installing dependencies, addressing issue https://github.com/microsoft/vscode-python-environments/issues/411. It's an initial implementation that covers some most common formatting errors, we can definitely expand coverage as we receive user feedback.

This PR

1. Ensured `pyproject.toml` adheres to relevant PEP standards (PEP 508, PEP 440, PEP 621, PEP 518) before attempting to install packages from it and provides feedback when validation fails, allowing the user to fix the file or proceed if they choose.

It added `validatePyprojectToml` in `pipUtils.ts` which checks:

- Package Name (PEP 508): Validates name format (alphanumeric, dots, hyphens, underscores).
image

- Version Format (PEP 440): Validates version strings against the standard (including pre/post/dev/local releases).
image

- Required Fields (PEP 621): Checks for the presence of the required name field in the [project] section.
image

- Build System (PEP 518): Checks for the presence of the required requires field in the [build-system] section.
image

2. Implemented `shouldProceedAfterPyprojectValidation` to display an error message with options to "Open pyproject.toml", "Continue Anyway", or "Cancel", only when user selects dependencies from incorrectly formatted `pypyroject.toml` file to install when creating environment.

Example user flow:

https://github.com/user-attachments/assets/4e919f2a-1934-4b8e-80d3-579880830b5a

3. Added a new test file `helpers.validatePyproject.unit.test.ts` with comprehensive unit tests.

### DIFF
--- a/src/managers/builtin/pipUtils.ts
+++ b/src/managers/builtin/pipUtils.ts
@@ -13,21 +13,46 @@ import { Installable } from '../common/types';
 import { mergePackages } from '../common/utils';
 import { refreshPipPackages } from './utils';
 
-export function validatePyprojectToml(toml: tomljs.JsonMap): string | undefined {
-    // 1. Validate package name (PEP 508)
-    if (toml.project && (toml.project as tomljs.JsonMap).name) {
-        const name = (toml.project as tomljs.JsonMap).name as string;
-        // PEP 508 regex: must start and end with a letter or digit, can contain -_., and alphanumeric characters. No spaces allowed.
-        // See https://peps.python.org/pep-0508/
-        const nameRegex = /^([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9._-]*[a-zA-Z0-9])$/;
-        if (!nameRegex.test(name)) {
-            return l10n.t('Invalid package name "{0}" in pyproject.toml.', name);
-        }
+export interface PyprojectToml {
+    project?: {
+        name?: string;
+        version?: string;
+    };
+    'build-system'?: {
+        requires?: unknown;
+    };
+}
+export function validatePyprojectToml(toml: PyprojectToml): string | undefined {
+    // 1. Validate required "requires" field in [build-system] section (PEP 518)
+    const buildSystem = toml['build-system'];
+    if (buildSystem && !buildSystem.requires) {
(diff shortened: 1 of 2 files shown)


src/managers/builtin/pipUtils.ts
```diff
@@ -13,6 +13,56 @@ import { Installable } from '../common/types';
 import { mergePackages } from '../common/utils';
 import { refreshPipPackages } from './utils';
 
+export function validatePyprojectToml(toml: tomljs.JsonMap): string | undefined {
+    // 1. Validate package name (PEP 508)
+    if (toml.project && (toml.project as tomljs.JsonMap).name) {
```
karthiknadig: This can be simplified if you check `toml.project` and exit out early.
You can additionally have `const project = toml.project as tomljs.JsonMap`, you could even create a `interface` for the version and name fields that you need and avoid `tomljs.JsonMap` entirely.

StellaHuang95: Sounds good, made some changes to address your feedback.

src/managers/builtin/pipUtils.ts
```diff
@@ -13,21 +13,46 @@ import { Installable } from '../common/types';
...
+    const project = toml.project;
+    if (!project) {
+        return undefined;
+    }
+
+    const name = project.name;
+    // 2. Validate required "name" field in [project] section (PEP 621)
+    // See PEP 621: https://peps.python.org/pep-0621/
+    if (!name) {
+        return l10n.t('Missing required field "name" in [project] section of pyproject.toml.');
```
eleanorjboyd, line 42: Love how well formatted this is! Easy to read and understand- thank you! With any of these msgs it aborts pkg install but the env still get created right? might be worth updating the msg to explain that to the user- unless I am missing a msg where that is included

StellaHuang95: image

So it depends on which option user picks, for the example above, the first and third one will abort both env creation and pkg install, but the second one will still create env.

eleanorjboyd: fantastic! looks great

Issue #640, opened 2025-07-24
CTRL-C is immediately issued upon run, execution gets cancelled by activation command


raydrak64:
Happens on multiple Ubuntu 24.04 workstations but not on any Windows 11 machines.
All machines are configured with the suggested fixes/settings per your emails for the Python Env extension, that seems to be doing well.

However...

Either from the hotkey or from the menus, when you run a Python file a CTRL-C is immediately issued. It is consistently intermittent, meaning, it's not 100% of the time but from observation it is 75-80% of the time. So, 8 out of 10 times you have to run the script again...and again... until it runs as expected. Sometimes you get lucky, and the CTRL-C is thrown to soon to interfere, like right before the venv is activated.

Thanks folks and cheers!

eleanorjboyd:
Hi! Thanks for the issue report and sorry this is happening! How are you running your python file? ie what command, entry point, and environment manager? This might give us a clue. Also is it CTRL-C specifically or does the terminal just reload? We were seeing that occur with some other users and wondering if they are connected

raydrak64:
Hello. Thank you for responding. I am awkward in giving information on issues so thank you.
I will answer in bullet points:

- How the file is being ran: Pull down menu: Run -> Run Without Debugging -or- hotkey (CTRL+F5)
- Environment Manager: Microsoft Python Environments or if you mean Python distro, just good ol' plain CPython and pip
- The terminal does not reload
- It is CTRL-C specifically... you see the ^C in the terminal and immediately shuts the party down like it's the cops

This has been happening for a while. Even before the venv fiasco. I just thought I'd say something now since in a communique it was said y'all be focusing on the Python Environments extension. This did not happen with the old extension, so I kept that installed until it stopped working altogether. Then I just never kept MS Python Environments installed because it would do this as soon as it was installed. But I have it installed now of course, and this is the only hiccup.

This is happening on all the Ubuntu 24.04 installs, so I am surprised it wasn't brought up by someone else.

Cheers!

raydrak64:
It still exists. Every 7 or 8 out of 10 attempts to run a python script through VSC results in CTRL-C being thrown. It doesn't matter the project or where it's being run. You have to run it multiple times before CTRL-C is not being thrown automatically. 

I don't bother anymore. Just run a terminal at the location and run the script manually. Debug or run the script manually in a separate terminal window.

^C..^C...^C....^C......^C....^C.....ok....is my life...

karthiknadig:
[...]

PR #1093, opened 2026-01-11
Add small timeout to avoid ctrl+c on Windows after activation 

anthonykim1:
Resolves: https://github.com/microsoft/vscode-python-environments/issues/640 
I couldnt seem to repro ^C to show up on Python file run on Mac, but only on Windows.

This is happening because we have some extra polling happening before `onCommandStarted` event gets fired: https://github.com/microsoft/vscode/blob/86db4eb05dbfd82dc9b4ea022883494257ac3d63/src/vs/platform/terminal/common/capabilities/commandDetectionCapability.ts#L727 Which is what leads to [resetting the value of promptInputModel](https://github.com/microsoft/vscode/blob/86db4eb05dbfd82dc9b4ea022883494257ac3d63/src/vs/platform/terminal/common/capabilities/commandDetection/promptInputModel.ts#L213): 

We need the promptInputModel's value to be empty and state to be not `execute` in order to not cause ^C upon `executeCommand` run: https://github.com/microsoft/vscode/blob/86db4eb05dbfd82dc9b4ea022883494257ac3d63/src/vs/workbench/contrib/terminal/browser/terminalInstance.ts#L1010 

So in that meantime before we clear the value of promptInput to empty string, the state will be `execute` and value of it will point to activation script. Which will lead to ^C.  

I think we'd have better chance when we remove some Windows specific heuristics in command detection, after rolling out newer conpty as default.

### DIFF
--- a/src/features/pythonApi.ts
+++ b/src/features/pythonApi.ts
@@ -49,6 +49,8 @@ import { runInBackground } from './execution/runInBackground';
 import { EnvVarManager } from './execution/envVariableManager';
 import { checkUri } from '../common/utils/pathUtils';
 import { waitForAllEnvManagers, waitForEnvManager, waitForEnvManagerId } from './common/managerReady';
+import { timeout } from '../common/utils/asyncUtils';
+import { isWindows } from '../common/utils/platformUtils';
 
 class PythonEnvironmentApiImpl implements PythonEnvironmentApi {
     private readonly _onDidChangeEnvironments = new EventEmitter<DidChangeEnvironmentsEventArgs>();
@@ -304,6 +306,12 @@ class PythonEnvironmentApiImpl implements PythonEnvironmentApi {
             options.cwd instanceof Uri ? options.cwd : Uri.file(options.cwd),
             environment,
         );
+        if (isWindows()) {
+            // Since we have extra polliing on Windows that could lead to race stale promptInputModel value,
+            // temporarily add small timeout to avoid ^C.
+            // TODO: Consider removing when we clean up polling with newer conpty: https://github.com/microsoft/vscode/issues/224488
+            await timeout(50);
+        }
         await runInTerminal(environment, terminal, options);
         return terminal;
     }


src/features/pythonApi.ts
```diff
@@ -304,6 +306,12 @@ class PythonEnvironmentApiImpl implements PythonEnvironmentApi {
             options.cwd instanceof Uri ? options.cwd : Uri.file(options.cwd),
             environment,
         );
+        if (isWindows()) {
+            // Since we have extra polliing on Windows that could lead to race stale promptInputModel value,
+            // temporarily add small timeout to avoid ^C.
+            // TODO: Consider removing when we clean up polling with newer conpty: https://github.com/microsoft/vscode/issues/224488
+            await timeout(50);
```
Tyriar, line 313: A timeout like this will always be flaky unfortunately.

Have you considered awaiting the activate call if triggered via `executeCommand`, and queuing any further commands until that resolves?

Tyriar (changes requested):

src/features/pythonApi.ts
anthonykim1: @Tyriar We actually do await for the activate call when activating with `executeCommand` https://github.com/microsoft/vscode-python-environments/blob/2693262312a8a6bbc66c923703e199c2d898e743/src/features/pythonApi.ts#L303-L307 
which goes to:
- https://github.com/microsoft/vscode-python-environments/blob/2693262312a8a6bbc66c923703e199c2d898e743/src/features/terminal/terminalActivationState.ts#L118 
- then waiting https://github.com/microsoft/vscode-python-environments/blob/2693262312a8a6bbc66c923703e199c2d898e743/src/features/terminal/terminalActivationState.ts#L183 for activated terminal 
- Eventually https://github.com/microsoft/vscode-python-environments/blob/2693262312a8a6bbc66c923703e199c2d898e743/src/features/terminal/terminalActivationState.ts#L265-L267 where we wait until `onDidEndShellExecution` is called


Maybe the better fix is to move where https://github.com/microsoft/vscode/blob/86db4eb05dbfd82dc9b4ea022883494257ac3d63/src/vs/platform/terminal/common/capabilities/commandDetection/promptInputModel.ts#L213 happens, to after execution in core?

Tyriar: Clearing the command when `onCommandFinished` fires makes sense (in addition to `onCommandStart` just in case?)

Issue #1096, opened 2026-01-13
Conda does not work with fish


eleanorjboyd:
When I activate a conda env and start a fish terminal I see

```
~/miniconda3/bin/activate (line 2): Unsupported use of '='. In fish, please use 'set _CONDA_ROOT "/root/miniconda3"'.
_CONDA_ROOT="/root/miniconda3"
^~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
from sourcing file ~/miniconda3/bin/activate
source: Error while reading file '/root/miniconda3/bin/activate'
```

I have run `conda init --all`, searched on web for every answer but just no one help. Please help, thanks a lot.

copied over from: https://github.com/microsoft/vscode-python/discussions/25727

PR #1097, opened 2026-01-13
Remove redundant shell activation and deactivation for FISH in getShellActivationCommands

eleanorjboyd:
fixes https://github.com/microsoft/vscode-python-environments/issues/1096


karthiknadig:
Is this redundant, because the user has run `conda init`? If they have not then what happens?

joaomoreno (approved):

PR #1119, opened 2026-01-23
Remove repo specific issue-label actions

anthonykim1:
This has been un-assigning me when manually triaging issues myself on this repo.

I think this workflow goes against the workflow @Yoyokrazy built, we should just use workflow from upstream instead of having one specific to env repo.

Do you see anything other actions that should be removed on behalf of your new workflow? @Yoyokrazy  https://github.com/microsoft/vscode-python-environments/tree/d84e05b4b58b208a0a4a33a9207aae53d8802f34/.github/workflows


Yoyokrazy:
For now this will take care of the assignment issues. There's a lot of overlap there though, info needed and tpi validator are low hanging removals that can take place, but for this PR I'll approve. With the triage processes being changed so much these days as Ben and I improve things, I'm going to opt to keep this as is for now. We can reassess in ~1mo when the dust settles.

Issue #1124, opened 2026-01-26
Regression: interpreter selection status bar not working when switching environments


StellaHuang95:
**Repro Steps**

1. Open a workspace with Python files and a `.venv` in the workspace and an empty workspace `settings.json`
2. Observe that the status bar shows the `.venv` as the selected environment
3. Click the status bar and select a global interpreter from the dropdown

**Expected Behavior**
The interpreter switches to the selected global one, and the status bar updates accordingly.

**Actual Behavior**
The status bar still shows the venv as the selected environment

https://github.com/user-attachments/assets/672fe5bc-baeb-4329-b464-f1d17e48ee84

This does not repro before the changes in https://github.com/microsoft/vscode-python-environments/pull/1120
Root cause still needs investigation, but we may want to revert the changes in the meantime.

StellaHuang95:
Note: This doesn't repro if there's already "python-envs.defaultEnvManager" and "python-envs.defaultPackageManager" entries in the `settings.json`

eleanorjboyd:
Hello! This fix is now out on python-environment-extension version 1.17.10341640 (the newest pre-release). To switch, go to the marketplace for this extension and either select to install the latest pre-release or click the button that says "switch to pre-release" for the extension version. We would appreciate greatly if you are able to give this a try on the pre-release version and let us know if your bug is fixed!

PR #1126, opened 2026-01-28
Fix Regression: interpreter selection status bar not working when switching environments 

StellaHuang95:
fixes https://github.com/microsoft/vscode-python-environments/issues/1124

The changes made in https://github.com/microsoft/vscode-python-environments/pull/1120 introduced logic to avoid writing unnecessary settings when the selected environment manager is the implicit default (system manager). This was intended to prevent polluting `settings.json`.

However, this optimization broke manual interpreter selection. When a user clicks the status bar and explicitly selects a global/system interpreter in a workspace with an empty `settings.json`, the selection is silently ignored because the code determines "this is the default, so don't write it."

This PR added an optional `forceWrite` flag to `EditAllManagerSettings`, `EditEnvManagerSettings`, and `EditPackageManagerSettings` interfaces and updated all user-initiated callers in `envManagers.ts` and `envCommands.ts` to pass `forceWrite: true`.

image

### DIFF
--- a/src/features/envCommands.ts
+++ b/src/features/envCommands.ts
@@ -341,7 +341,7 @@ export async function setEnvManagerCommand(em: EnvironmentManagers, wm: PythonPr
     if (projects && projects.length > 0) {
         const manager = await pickEnvironmentManager(em.managers);
         if (manager) {
-            await setEnvironmentManager(projects.map((p) => ({ project: p, envManager: manager })));
+            await setEnvironmentManager(projects.map((p) => ({ project: p, envManager: manager, forceWrite: true })));
         }
     }
 }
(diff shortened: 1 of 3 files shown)


eleanorjboyd:
thinking on this and the edge cases here- thanks for the PR

StellaHuang95:
> thinking on this and the edge cases here- thanks for the PR

Thanks. Looking back at the original bug, the user reported that `python-envs.defaultEnvManager: "ms-python.python:system" and python-envs.pythonProjects: []` were added to `settings.json` unexpectedly.

It seems reasonable to me that `python-envs.defaultEnvManager` gets added when an interpreter is selected for the workspace, based on the logic in `settingHelper.ts`. Would it make sense to check with Karthik whether this behavior is intentional? But I do agree that the logic around `python-envs.pythonProjects` could be improved so that the setting isn’t written when the value is empty.

eleanorjboyd:
ended up doing a larger overhaul since this needed area in general needed some work. Incorporated your PR improvements into here: https://github.com/microsoft/vscode-python-environments/pull/1155

Issue #936, opened 2025-10-12
.env File no longer updates commented variables


solman252:
On the previous release, commenting out a variable in the .env file would prompt the user to reload the terminal and reinject from .env,
but now, commenting out a variable doesn't prompt, nor will the variable update the next time the user is prompted for another change.

eleanorjboyd:
To confirm you have the setting "python.useEnvFile" set to true right?

solman252:
Correct,
in my .vscode/settings.json:

{
    "python.terminal.useEnvFile": true,
    "python.envFile": "${workspaceFolder}/.env"
}

eleanorjboyd:
what do you have for `terminal.integrated.shellIntegration.enabled`? Wondering if @anthonykim1 would have any ideas either

solman252:
Doesn't show under either my workspace or user settings.json, but in the UI settings it shows as enabled.

anthonykim1:
I dont remember us supporting 

> On the previous release, commenting out a variable in the .env file would prompt the user to reload the terminal and reinject from .env

solman252:
Supporting commented vars? I assure you, there is support for that in both the current and previous releases.

I've attatched a video showcasing the previous behavior in v1.8.0, and the problematic behavior in v1.10.0.
(I had to zip it because you can't attach videos directly on Github, plus I also attatched the workspace folder)

I forgot to showcase it in the video, but closing the terminal and reopening it manually also does not fix the issue, and the environment variable seems to be cached.

[Environment Test.zip](https://github.com/user-attachments/files/22901340/Environment.Test.zip)

PR #1131, opened 2026-01-30
Fix issue where terminal environment variables are not removed when they are commented out or deleted from .env files.

StellaHuang95:
Fixes https://github.com/microsoft/vscode-python-environments/issues/936

The problem this PR fixed is when users comment out or remove a variable from their .env file:
```
API_KEY=secret
# DEBUG=true   ← commented out
```

The DEBUG variable remained in terminal environments because:

1. `dotenv.parse()` ignores commented lines (seems the right behavior per dotenv spec)
2. The returned object only contained active variables: `{ API_KEY: 'secret' }`
3. The old code only iterated over keys in `envVars` and it never saw `DEBUG`
4. The `delete()` call was never reached for removed/commented variables

Solution is to call `clear()` before re-injecting variables, which makes the fix stays at the terminal injection layer rather than modifying the parsing layer, avoiding impacting on other consumers.

I also noticed some other issues with setting watcher, currently there's no file watcher watching the settings change, so a reload will be required when toggling the setting. We can consider adding a config listener.


src/features/terminal/terminalEnvVarInjector.ts
```diff
@@ -149,6 +149,8 @@ export class TerminalEnvVarInjector implements Disposable {
                 traceVerbose(
                     `TerminalEnvVarInjector: Env file injection disabled for workspace: ${workspaceUri.fsPath}`,
                 );
+                // Clear any previously set variables when injection is disabled
+                envVarScope.clear();
```
anthonykim1: Can you test this with `python-envs.terminal.autoActivationType` set to shell Startup? 
We don't want to clear variables from shell startup as that is what allows auto-activation for shell startup.

My concern is that calling clear on this will also clear activation related variable as well. These are variables with `VSCODE_PYTHON_(SHELL)_ACTIVATE`

eleanorjboyd: I think there is a potential issue here since when envVarScope.clear() is called on the scoped collection, it clears ALL variables in that workspace scope - not just the .env variables, but also the shell activation variables set by ShellStartupActivationVariablesManager. If you could review and confirm if this is the case or not that would be great!

StellaHuang95: Thanks for pointing that out. Since all variables are stored together in the same collection, there's no good way to distinguish which variables came from which feature. 

The new approach I came up with is that I made a map tracking only those .env-specific keys instead of using the merged environment containing all of `process.env`, only deletes those specific keys when the setting is off or any keys from `.env` are removed, preserving shell startup activation keys and all other variables in the collection. Let me what you think, thanks!

StellaHuang95: Good point, I came up with a new approach, please see comment above.

Issue #912, opened 2025-10-08
support pipenv path setting


eleanorjboyd:

eleanorjboyd:
Testing:


## Overview
This tests the new support for the `python.pipenvPath` VS Code setting, which allows users to specify a custom path to the pipenv executable.

## Prerequisites
```bash
# Install pipenv if not already available
pip install pipenv

# Create workspace with nested project structure
mkdir pipenv-test-workspace && cd pipenv-test-workspace
mkdir -p projects/nested-app
cd projects/nested-app
echo "requests==2.28.0" > requirements.txt
```

## Test Steps

### 1. Create Pipenv in Non-Standard Location  
```bash
# From the nested-app directory
pipenv install -r requirements.txt
# This creates a Pipfile and virtual environment in projects/nested-app/
```

### 2. Verify Baseline (Should NOT be found)
- Open the workspace root (`pipenv-test-workspace`) in VS Code
- Check Python environment picker 
- The nested pipenv environment should NOT appear automatically
- Extension only searches standard locations by default

### 3. Configure Custom Search Path
Add to workspace `.vscode/settings.json`:
```json
{
  "python.pipenvPath": "/path/to/projects/nested-app"
}
```

### 4. Test Discovery with Custom Path
- Reload VS Code window (Ctrl/Cmd + Shift + P → "Developer: Reload Window")
- Open Python environment picker
- Nested pipenv environment should now appear
- Extension searches the specified path for pipenv environments

amunger:
I'm not seeing it in windows

<img width="1374" height="648" alt="Image" src="https://github.com/user-attachments/assets/935cf28f-8069-4357-8411-2b5f162b69a0" />

<img width="600" height="380" alt="Image" src="https://github.com/user-attachments/assets/9085d077-3413-4b9c-81ab-bf7aae084404" />

eleanorjboyd:
need to make a design decision here before proceeding. I misunderstood the setting purpose originally. see https://github.com/microsoft/vscode-python-environments/issues/990

Issue #918, opened 2025-10-09
support poetry path setting


eleanorjboyd:
support users setting python.poetryPath in their settings

eleanorjboyd:
bug with: https://github.com/microsoft/python-environment-tools/issues/267

PR #1244, opened 2026-02-18
fix pipenv and poetry setting

eleanorjboyd:
ensure that pipenvPath and poetryPath are supported as settings

fixes https://github.com/microsoft/vscode-python-environments/issues/912 and https://github.com/microsoft/vscode-python-environments/issues/918

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "vscode-python-envs",
-    "version": "1.20.0",
+    "version": "1.21.0",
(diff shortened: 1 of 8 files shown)


karthiknadig:
@eleanorjboyd should we check other settings too that they follow the same priority?

eleanorjboyd:
@karthiknadig which ones? I confirmed it happens for `condaPath` then `pixiPath` would be supported through the community extension right?

Issue #1288, opened 2026-02-26
Remote Reconnect Forces Ctrl+C on Running Terminal Process for Environment Activation


MartinKliemank:
This was mainly copied from #947 , as it just happened to me and the other issue was closed but apparently not fixed:

"When Ctrl+C reconnects to a remote terminal, the extension appears to attempt sending Ctrl+c to terminate the currently running user process in the terminal before activating the python environment.

We believe this behavior is undesirable. Specifically, for processes that are long-running and have extremely high restart costs (such as deep learning training), the extension forcibly killing these processes without user confirmation causes significant disruption and potential data loss.

Furthermore, I suggest avoiding the use of Ctrl+C or similar methods to interfere with or kill user-running processes when the remote terminal is reconnected."

I happened to have one terminal with tmux open that had been running for > 2days during a short disconnect. Upon reconnect "Ctrl + C, conda activate ..." was issued to all open terminals, killing the process. If a user process is already running in the terminal, the environment is likely already properly selected, otherwise it could not be running, so this behaviour does not seem desirable or helpful.

This is with the current release version of VSCode and the extension, all stock settings in the extension.

anthonykim1:
Thanks for the issue @MartinKliemank 

Two ways to go about this,

1. Please set `python-envs.terminal.autoActivationType` to `shellStartup` in your settings.
Also add `  "python.useEnvironmentsExtension": true,` to your settings.json.

2. If you want activation by command specifically, try modifying value (in ms unit) 
See more descriptions here: https://code.visualstudio.com/updates/v1_106#_consolidated-shell-integration-timeout-setting

ShishckovA:
I completely agree with the author of issue. I've just lost 4 hours of my work due to a reconnect with an open tmux. Martin, I understand your pain

Issue #861, opened 2025-09-24
Settings Review: Python ext & Python Envs ext


eleanorjboyd:
Settings in Python Extension and their status on support in the Python Environments Extension

| Setting | Description | Supported in Python Envs Ext | Deatils |
|--------|--------|--------|--------|
| python.condaPath | Path to the conda executable to use for activation (version 4.4+) | yes | should carry over  |
| python.createEnvironment.contentButton | Show or hide Create Environment button in the editor for requirements.txt or other dependency files. | ? | need to decide if it is worth keeping |
| python.createEnvironment.trigger | Detect if environment creation is required for the current project | ? | should carry over |
| python.defaultInterpreterPath | Path to default Python to use when extension loads up for the first time, no longer used once an interpreter is selected for the workspace. See [here](https://aka.ms/AAfekmf) to understand when this is used | Yes | Yes |
| python.envFile | Absolute path to a file containing environment variable definitions. | Yes | Yes, must have python.terminal.useEnvFile enabled |
| python.globalModuleInstallation | Whether to install Python modules globally when not using an environment. | ? | useful and make sure it carries over |
| python.pipenvPath | Path to the pipenv executable to use for activation.| ? | should carry over |
| python.pixiToolPath | Path to the pixi executable. | ? | discuss with ext author |
| python.poetryPath | Path to the poetry executable | ? | should carry over |
| python.terminal.activateEnvInCurrentTerminal | Activate Python Environment in the current Terminal on load of the Extension.|  yes->no | deprecating in favor of  `python-envs.terminal.autoActivationType` |
| python.terminal.activateEnvironment | Activate Python Environment in all Terminals created.| yes->no | deprecating in favor of  `python-envs.terminal.autoActivationType`  |
| python.terminal.executeInFileDir | When executing a file in the terminal, whether to use execute in the file's directory, instead of the current open folder. | works | should but need to investigate if it currently does |
| python.terminal.launchArgs | Python launch arguments to use when executing a file in the terminal. | Yes | not impacted by move |
| python.venvFolders | Folders in your home directory to look into for virtual environments (supports pyenv, direnv and virtualenvwrapper by default). | yes->no | to be deprecated in favor of globalSearchPath |
| python.venvPath | Path to folder with a list of Virtual Environments (e.g. ~/.pyenv, ~/Envs, ~/.virtualenvs). | yes->no | to be deprecated in favor of globalSearchPath |

eleanorjboyd:
related issues: 
[...]

PR #1299, opened 2026-02-28
turn off terminal.autoActivationType by default

akgit18:
As a VSCode user who did not manually download this extension, I do not want it to automatically run `^C source <activate>` in new terminals, so that I can run commands when I open new terminals without getting interrupted.

Also, I might not want to be automatically put into a venv, either.


akgit18:

karthiknadig:
You can set this to `off` at user level and turn on settings sync in VS Code so any new instance of VS code on any machine will have that setting set to `off`. There is another alternative, set it to shellStartup, that mode will do activations without sending command if you still need activated envs.

Most users have command as selected default. So turning this off be default will break large number of users.

I appreciate you taking the time to create the PR, but unfortunately this is the value that nearly all users have set.

akgit18:
> You can set this to off at user level

You can set this to off at the user level _if_ you know that this extension was installed. This extension seems to frequently be installed without the users' knowledge--see [most comments written in the reviews](https://marketplace.visualstudio.com/items?itemName=ms-python.vscode-python-envs&ssr=false#review-details).

> Most users have command as selected default. So turning this off be default will break large number of users.

I would argue that fixing the experience for users who don't know about this extension would provide more value than leaving it on command for the users who know about the extension and intentionally leave the default as `command`.

Not installing the extension for users who don't ask for it would probably satisfy both groups. Is that in consideration?

karthiknadig:
I agree with you on fixing the experience. But I disagree with the diagnosis and solution.

There are 2 separate bugs here. The `^C` bug is from core VS Code feature called "Shell Integration", the second bug is this extension not honoring the Python extensions activation settings automatically.

**first issue**: The `^C` core bug will affect all cases where an extension (any extension) has to send command to the focused terminal. This should be addressed (tracking issue for this https://github.com/microsoft/vscode-python-environments/issues/1288). This means any other extension like git or R or anyone that relies on shell state APIs from core will be broken, so that is independent issue from this extension. Since you are running into this please share your logs form Output > Terminal

image



**second** issue: Core Python extension has these two settings `python.terminal.activateEnvironment` (where default is `true`) and `python.terminal.activateEnvInCurrentTerminal` (where the default is `false`). This extension is supposed to inherit that setting, in the following way. If both of those two are `true` then this extension should use `autoActivatoinType` as `command` (eventually when Shell support goes out fully it would be switched to `shellStartup`). If either is set to `false` then `autoActivationType` should be `off`. I created a tracking bug to investigate if this https://github.com/microsoft/vscode-python-environments/issues/1300. The problem I think is that currently it is attempting to activate terminals that are currently focused, because in the state where one of those settings is `true` and the other is `false`, the behavior is a bit broken. This is because we want to use `shellStartup` where activations always happen in the background, there is no distinction between focused and un-focused terminals. The `command` is a fallback mode because `shellStartup` is not set as default. One solution could be setting `shellStartup` as default then we can ignore the (focused and background cases) and make `command` activation aware of (current and background) and more closely follow the core settings. That would accurately match the behavior but turning it `off` breaks for a significantly large set of users.

Regarding not installing the extension automatically or turning activation off by default, I understand the concern about unexpected behavior. However, activation is currently the expected workflow for the vast majority of Python users, and most users rely on that default. Changing the default to `off` would silently break environment activation for a significantly larger group than the set of users encountering this issue. The better long-term direction is to reduce surprise without regressing the common case. As `shellStartup` matures, making it the default would eliminate command injection into focused terminals and match the behavior more closely with the core Python settings, while still preserving auto activation. That approach improves the experience broadly rather than shifting the cost from one user group to another.

anthonykim1:
@akgit18 Which version of VS Code and Python environment extension are you on? 

For the  `^C` keyboard interrupt, you should be seeing less of that latest VS Code version.
There has been PRs like: https://github.com/microsoft/vscode/pull/287139 which closed https://github.com/microsoft/vscode-python-environments/issues/640 

Also would be helpful to see a gif with your setup, given you would update to insiders + latest pre-release of env extension.
[Insiders build](https://code.visualstudio.com/insiders) and latest pre-release version of environment extension.

There is also a timeout setting where you can just the time in ms to prevent/mitigate as much as possible ^C from happening ( But yes it will take longer for activate command to show in terminal because you are waiting that much time, by default it just waits until your prompt is ready in the terminal): https://code.visualstudio.com/updates/v1_106#_consolidated-shell-integration-timeout-setting

In most of the cases `^C` is to ensure activate command runs successfully regardless of what is running before/what user is typing before. 

You will see this a bunch with copilot commands in terminal. Highly recommend shellStartup if you want activation, but you are always welcome to set `python-envs.terminal.autoActivationType` to false. We were supposed to deprecate the two old setting `python.terminal.activateEnvironment (where default is true) and python.terminal.activateEnvInCurrentTerminal (default false)` which is causing confusion. I'll think more about how we can best transition.

These were discussed earlier --> https://github.com/microsoft/vscode-python-environments/issues/861

dogganon:
> In most of the cases `^C` is to ensure activate command runs successfully regardless of what is running before/what user is typing before.

Sounds like a "feature" that nobody asked for in the first place 🤦 Why would any sane person want unsolicited `^C`s on their terminal, if they are running a critical process?

And such "feature" relies on the assumption that the running process will respond to `^C` and terminate correctly (obviously some will not if SIGINT is handled differently). So to get the facts straight, whatever "feature" you are adding to VS Code it's not a feature, it's a HACK.

@anthonykim1

anthonykim1:
If you think there is a better way to ensure sending commands in the terminal when there could be multiple of them and the user input, please feel free to contribute. 

Always open to constructive feedbacks or PRs https://github.com/microsoft/vscode :) 

@dogganon

dogganon:
@anthonykim1 The point is that sending commands to a terminal with a running process is simply not supported (at least in POSIX shells). If you insist that the feature needs to be implemented, write your own "VS Code" shell and don't use the one shipped with the OS.

anthonykim1:
> sending commands to a terminal with a running process is simply not supported (at least in POSIX shells)

@dogganon If the point is "you can't make an interactive shell execute a command while it's handed the terminal to a child via tcsetpgrp() and is parked on waitpid(..., WUNTRACED)" — sure, obviously. That's exactly why ^C goes first (interrupt → shell reclaims the terminal → then send), and why the timeout setting exists for people who'd rather wait for the prompt than interrupt.

But "terminal" is doing a lot of work in that sentence. The PTY, the line discipline, the terminal emulator — all of that keeps working while a foreground job runs. 

Writing bytes to the PTY master is how every keystroke reaches that foreground program; there's nothing to "support," it's the mechanism. The only thing that can't process new input is the shell, and the shell isn't the terminal.

And ^C isn't a shell feature. It's the VINTR byte (0x03). With ISIG set, the kernel's line discipline turns that into SIGINT delivered to the foreground process group — termios(3). The shell doesn't have to be reading for that to happen. stty -a shows the config; tmux send-keys / expect / screen -X stuff are all writing the same bytes.

"Write your own shell" doesn't land here — this is the OS behavior. We're not working around the TTY layer, we're using it.

dogganon:
> That's exactly why ^C goes first (interrupt → shell reclaims the terminal → then send), and why the timeout setting exists for people who'd rather wait for the prompt than interrupt.

@anthonykim1 This doesn't make sense at all; You should wait for the current process to finish rather than relying on timeout and "hoping" the process would finish. The python interactive shell, for example, will not exit upon `^C` and oftentimes your so-called feature wrongfully injects `conda activate`, resulting in syntax errors.

The best solution is to not rely on `^C` but some other mechanism that does not interfere with the user input. AFAIK this is NOT currently possible with shells such as bash, unless you write your own shell (or wrapper) that does the scheduling of commands after the current one finishes.

PR #1306, opened 2026-03-03
feat: add telemetry for PET process failures and exits

eleanorjboyd:
- Add telemetry and better diagnostics around the pet binary
-  PET_START_FAILED event is now emitted with the error code, reason, platform, and architecture
- PET_PROCESS_EXIT event is emitted flagging whether the exit was expected (shutdown/restart) or a crash
- For activation the single broad try-catch was split into two — one specifically for pet startup (which emits telemetry and bails early on failure) and one for manager registration — so errors are attributed correctly and don't bleed into each other.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -95,7 +95,11 @@ import { collectEnvironmentInfo, getEnvManagerAndPackageManagerConfigLevels, run
 import { EnvironmentManagers, ProjectCreators, PythonProjectManager } from './internal.api';
 import { registerSystemPythonFeatures } from './managers/builtin/main';
 import { SysPythonManager } from './managers/builtin/sysPythonManager';
-import { createNativePythonFinder, NativePythonFinder } from './managers/common/nativePythonFinder';
+import {
+    createNativePythonFinder,
+    NativePythonFinder,
+    PetBinaryNotFoundError,
+} from './managers/common/nativePythonFinder';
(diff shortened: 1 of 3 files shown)


src/extension.ts
```diff
@@ -522,12 +522,39 @@ export async function activate(context: ExtensionContext): Promise<PythonEnviron
...
+            // Errno error (has .code) = spawn failed (ENOENT, EACCES, EPERM, etc.).
+            const reason = errnoError.code ? 'spawn_failed' : 'binary_not_found';
+            sendTelemetryEvent(EventNames.PET_START_FAILED, undefined, {
+                errorCode: errnoError.code ?? 'UNKNOWN',
+                reason,
+                platform: process.platform,
+                arch: process.arch,
+            });
+
+            window.showErrorMessage(
```
eleanorjboyd: @karthiknadig if we cannot find the binary, is that a game-over scenario ? how should we present / prompt the user? I guess they could enter a interpreter path themselves so maybe its a notification instead?

eleanorjboyd:
@karthiknadig could you check my one question here

src/extension.ts
karthiknadig: If `pet` is not found, yes it is game over. Really something went wrong with that install. User will have to uninstall and re-install.

PR #1307, opened 2026-03-03
feat: pre-warm environment manager caches during activation and background refresh

eleanorjboyd:
Managers don't scan on registration- `initialize()` only runs when a consumer first calls `getEnvironments()` making the whole discovery process lengthly. Every step is sequential and lazy. On a large filesystem or slow disk this is 10–30 seconds of real time with the user watching an empty interpreter picker.

The Python extension avoids this by starting discovery in the `NativePythonFinderImpl` **constructor**, so results begin streaming before other exts makes its first call.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -10,7 +10,7 @@ import {
     window,
 } from 'vscode';
 import { PythonEnvironment, PythonEnvironmentApi, PythonProjectCreator } from './api';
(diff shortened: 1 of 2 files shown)


eleanorjboyd:
@StellaHuang95 what is your take on this one? do you think it would be useful?

StellaHuang95:
LGTM

StellaHuang95 (dismissed):

PR #1308, opened 2026-03-03
feat: add manager ready timeout setting (experimental)

eleanorjboyd:
add experimental setting allowing the user to control how long the manager timeout is (ie how long to wait for managers to register or attempt to register environments they discover)

### DIFF
--- a/package.json
+++ b/package.json
@@ -135,6 +135,15 @@
                     "description": "%python-envs.alwaysUseUv.description%",
                     "default": true,
                     "scope": "machine"
+                },
+                "python-envs.experimental.managerReadyTimeout": {
+                    "type": "integer",
+                    "default": 15,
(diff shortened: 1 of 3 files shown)


eleanorjboyd:
@StellaHuang95  what do you think about this addition? I am on the fence about if it is useful or not at this stage

StellaHuang95:
I am not sure if users will ever understand what a "manager ready timeout" means, the concept of registering is an internal implementation detail, and a user who hits timeout will probably just think the extension is broken. The only scenarios that I can think of that this helps is an advance user with slow remote file system that hits the timeout, reads the output warning, understands what it means. That's a very narrow audience in my opinion.  I'd rather focus on making the timeout behavior smarter internally rather than making it configurable externally.

PR #1328, opened 2026-03-06
Get rid of vscode-engineering template in the npm publish pipeline

StellaHuang95:
It's been a disaster trying to get the pipeline running from a different org. I keep running into service connection and permission issues, fixing one just reveals two more. At this point it doesn't seem feasible to use the `vscode-engineering` template from another org. 

I looked at some other pipelines we migrated for extensions like `pylint`, and they dropped the template and just run the raw jobs for npm pack and publishing. I'm going to try the same approach for this extension.


rchiodo (approved):

build/templates/test-steps.yml
```diff
@@ -1,19 +0,0 @@
-parameters:
```
bschnurr, line 1: r you sure we're supposed to delete these?

StellaHuang95: I added these before, it's part of using the template.

StellaHuang95: They can be removed as I am no longer using the template

PR #1329, opened 2026-03-06
Some more changes to npm publish pipeline

StellaHuang95:


edvilme:
I think we need to run `npm i` to update the lock file

StellaHuang95:
Test run works and npm is published as build artifact. https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=13485743&view=results

Issue #1361, opened 2026-03-16
[Python Environments] 


ARCHANA123G:
Type: <b>Bug</b>

<details>
<summary>Environment Information</summary>

```
Extension Version: 1.22.0
Python Extension Version: 2026.4.0

Registered Environment Managers (3):
  - ms-python.python:system (Global)
  - ms-python.python:venv (venv)
  - ms-python.python:conda (Conda)

Total Available Environments: 4
Environment Details:
  1. Python 3.11 (64-bit) (3.11.3) - C:\Users\Admin\AppData\Local\Programs\Python\Python311\python.exe
  2. env (3.13.12) (3.13.12) - c:\Users\Admin\mycode\env\Scripts\python.exe
  3. tf-env (3.10.19) (3.10.19) - C:\Users\Admin\Anaconda2\envs\tf-env
  4. base (2.7.15) (2.7.15) - C:\Users\Admin\Anaconda2

Python Projects (1):
  1. c:\Users\Admin\mycode
     Environment: env (3.13.12)

Extension Settings:
  Default Environment Manager: ms-python.python:venv
  Default Package Manager: ms-python.python:pip
Auto-activation is "command". Activation based on first 'py-env.terminal.autoActivationType' setting which is 'command' and 'python.terminal.activateEnvironment' if the first is undefined which is 'true'.

```

</details>

Extension version: 1.22.0
VS Code version: Code 1.111.0 (ce099c1ed25d9eb3076c11e4a280f3eb52b4fbeb, 2026-03-06T23:06:10Z)
OS version: Windows_NT x64 10.0.26200
Modes:

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) i5-8350U CPU @ 1.70GHz (8 x 1896)|
[...]

Issue #1360, opened 2026-03-15
[Python Environments] 


MIJOHETA:
Type: <b>Bug</b>

<details>
<summary>Environment Information</summary>

```
Extension Version: 1.22.0
Python Extension Version: 2026.4.0

Registered Environment Managers (3):
  - ms-python.python:system (Global)
  - ms-python.python:venv (venv)
  - ms-python.python:conda (Conda)

Total Available Environments: 5
Environment Details:
  1. Python 3.14.3 (3.14.3) - C:\Users\Hezek\AppData\Local\Python\pythoncore-3.14-64\python.exe
  2. Python 3.12.7 (3.12.7.final.0) - C:\msys64\ucrt64\bin\python3.exe
  3. Python 3.12 (64-bit) (3.12.4) - C:\Users\Hezek\AppData\Local\Programs\Python\Python312\python.exe
  4. spyder-runtime (3.11.13) (3.11.13) - C:\Users\Hezek\AppData\Local\spyder-6\envs\spyder-runtime
  5. base (3.11.9) (3.11.9) - C:\Users\Hezek\AppData\Local\spyder-6

Python Projects (1):
  1. e:\Tgs2WEB
     Environment: Python 3.14.3

Extension Settings:
  Default Environment Manager: ms-python.python:system
  Default Package Manager: ms-python.python:pip
Auto-activation is "command". Activation based on first 'py-env.terminal.autoActivationType' setting which is 'command' and 'python.terminal.activateEnvironment' if the first is undefined which is 'true'.

```

</details>

Extension version: 1.22.0
VS Code version: Code 1.111.0 (ce099c1ed25d9eb3076c11e4a280f3eb52b4fbeb, 2026-03-06T23:06:10Z)
OS version: Windows_NT x64 10.0.26200
Modes:

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|12th Gen Intel(R) Core(TM) i5-12500H (16 x 3110)|
[...]

Issue #1357, opened 2026-03-13
[BUG] Unable to detect or recognize python executable on certain machines


mogres:
Type: <b>Bug</b>

# Behaviour

Python environments extension is having trouble recognizing the python executable on certain machines accessed via remote explorer. Python environment tool fails after three attempts.

I am able to activate and execute python venv through the command line, but the extension fails with the error: `Selected file is not a valid python interpreter`

## Steps to reproduce:

1. Open a python file
2. Click `Select Python Interpreter` from the status bar (this should have been auto-recognized)
3. Provide path to executable (`.venv/bin/python`) in the local directory, verified working through a clean terminal outside vscode.

# Diagnostic data

<details>

<summary>Output for <code>Python</code> in the <code>Output</code> panel (<code>View</code>→<code>Output</code>, change the drop-down the upper-right of the <code>Output</code> panel to <code>Python</code>)
</summary>

<p>

```
2026-03-13 17:40:54.274 [error] [
  'Failed to activate a workspace, Class name = v, completed in 170ms, has a falsy return value, Arg 1: <Uri:[REDACTED]>, Return Value: undefined',
  [Error: Python Environment Tools (PET) failed after 3 restart attempts. Please reload the window or check the output channel for details. To debug, run "Python Environments: Run Python Environment Tool (PET) in Terminal" from the Command Palette.
```
Running the python environment tool from the terminal using:
```
.vscode-server/extensions/ms-python.vscode-python-envs-1.22.0-linux-x64/python-env-tools/bin/pet resolve .venv/bin/python
```
works as expected with output:
```
Environment found for ".venv/bin/python"
Environment (Uv)
   Name        : [REDACTED]
   Executable  : .venv/bin/python
   Version     : 3.11.12.final.0
   Prefix      : [REDACTED]
   Architecture: x64
   Symlinks    : ".venv/bin/python"
               : ".venv/bin/python3"
               : ".venv/bin/python3.11"
               : "[REDACTED]"

Resolve completed in 984ms
```
</p>
</details>

<details>

<summary>Output for <code>Python Environments</code> in the <code>Output</code> panel (<code>View</code>→<code>Output</code>, change the drop-down the upper-right of the <code>Output</code> panel to <code>Python Environments</code>)
</summary>

<p>

```
2026-03-13 18:03:11.767 [debug] Checking venv (ms-python.python:venv) for [REDACTED-LOCAL-PATH-TO-VENV-PYTHON-EXECUTABLE]
[...]

PR #1362, opened 2026-03-16
fix: revert workspaceSearchPaths default - Large Repo/Remote env recognition is now instant

Gorluxor:
After some debugging, I located the culprit of so many issues, @eleanorjboyd (tagging you, we discussed in https://github.com/microsoft/vscode-python-environments/issues/1335 and https://github.com/microsoft/vscode-python-environments/issues/1316), in addition to the PR that changed this (https://github.com/microsoft/vscode-python-environments/pull/1191), so you would know best why this was changed. 
Per my understanding, the workspaceSearchPaths should be extended by users for non-default paths, but it's being broadly applied to everyone. A temporary fix for affected versions could be to override the default, but in general, it makes more sense for the integrated default searches to find conda/pyenv, etc. (as they do so in milliseconds).  

People who are using affected versions just need to change .vscode/settings.json:
```
{
      "python-envs.workspaceSearchPaths": []
}
```
Additionally tagging @karthiknadig as you did https://github.com/microsoft/vscode-python-environments/pull/1154/

The default workspaceSearchPaths of `["./**/.venv"]` introduced in https://github.com/microsoft/vscode-python-environments/pull/1191/ (2026-02-06) causes the PET server to recursively walk the entire workspace during `configure`, which exceeds the 30s timeout (https://github.com/microsoft/vscode-python-environments/pull/1154/ on 2026-02-03) on large workspaces. PET's built-in Venv locator already discovers `.venv` directories at workspace roots without needing extra search paths. In addition to affecting Conda/Mamba users, who likely don't even need this search (pip files are usually installed under a predefined folder structure in conda/mamba). 

Affected versions on vscode extensions: v1.18.0, v1.18.1, v1.20.0, v1.20.1, v1.22.0, and preview/main (v1.16.0 and earlier used an empty default and are not affected).

Fixes an environment discovery failure in workspaces with deep directory trees (e.g., ML projects with datasets, checkpoints, and other large repos). As this is my first open-source commit, feel free to guide me if you need anything else. 

Issues likely fixed with this PR: https://github.com/microsoft/vscode-python-environments/issues/1330, https://github.com/microsoft/vscode-python-environments/issues/1319, https://github.com/microsoft/vscode-python-environments/issues/1313, https://github.com/microsoft/vscode-python-environments/issues/1310, https://github.com/microsoft/vscode-python-environments/issues/1286, https://github.com/microsoft/vscode-python-environments/issues/1407, https://github.com/microsoft/vscode-python-environments/issues/1316, https://github.com/microsoft/vscode-python-environments/issues/1309
New issues that are likely the same problem: https://github.com/microsoft/vscode-python-environments/issues/1357, https://github.com/microsoft/vscode-python-environments/issues/1354, https://github.com/microsoft/vscode-python-environments/issues/1335

Not sure how to check if this issue is reported by https://github.com/microsoft/vscode-python-environments/issues/1361, https://github.com/microsoft/vscode-python-environments/issues/1360 and https://github.com/microsoft/vscode-python-environments/issues/1358 I leave that to you. 

I think the first reported issue of this is actually your issue https://github.com/microsoft/vscode-python-environments/issues/1206 @eleanorjboyd, sounds to me like this, as in my tests, when I reload the window, it tries to force refresh conda envs because of PET configuration update, see below:
```
2026-03-16 15:07:47.814 [info] [pet] configure: Sending configuration update: {"workspaceDirectories":["<project_path>"],"environmentDirectories":["<project_path>/**/.venv"],"pipenvExecutable":"pipenv","poetryExecutable":"poetry","cacheDirectory":"/home/<username>/.vscode-server/data/User/globalStorage/ms-python.vscode-python-envs/pythonLocator"}
2026-03-16 15:07:48.340 [info] Refreshing conda environments (hardRefresh=false)
2026-03-16 15:07:48.379 [info] Refreshing Conda Environments
2026-03-16 15:07:48.379 [info] Refreshing conda environments (hardRefresh=true)
```

I used conda-forge's nodejs 24.14.0 (h3d65ac4_0) for the tests, and both for main and your commit (https://github.com/microsoft/vscode-python-environments/pull/1344), I get the following attached files for `npm run compile-tests` followed by `npm run unittest`, which should also be mentioned in the Contribution nodes see below, and I primarily deal with Python. 
https://github.com/microsoft/vscode-python-environments/blob/ecbcdacf414215477dd312d38914f80ca6611530/CONTRIBUTING.md?plain=1#L27-L35

[unittest_logs_my_fix.txt](https://github.com/user-attachments/files/26024818/unittest_logs_my_fix.txt)
[unittest_logs_with_1344PR.txt](https://github.com/user-attachments/files/26024820/unittest_logs_with_1344PR.txt)

Additional notes: not sure if doing a full subdirectory file search is the way to go; I leave those design choices up to you guys, but it shouldn't be the default for sure. As in my experimenting, this causes crashes in remote VS Code, unresponsive VS Code, PyLance not working at all, and other tools that rely on this fast discovery. Even if by miracle it resolves the issue in my setup (in some of the versions I tested), it still takes 5 minutes instead of 3 seconds.

### DIFF
--- a/package.json
+++ b/package.json
@@ -124,7 +124,7 @@
                 "python-envs.workspaceSearchPaths": {
                     "type": "array",
                     "description": "%python-envs.workspaceSearchPaths.description%",
-                    "default": ["./**/.venv"],
+                    "default": [],
                     "scope": "resource",
                     "items": {
                         "type": "string"


JTMarsh556:
Im sorry but what is the temporary fix for this?  I use anaconda and everything worked great but now this has become a mess.  

ETA: I have tried adding the path to my anaconda environments to "python-envs.workspaceSearchPaths" but that didnt work and I have tried empty braces as you have above but neither worked.

Gorluxor:
@JTMarsh556 check what solved it for me as fix here https://github.com/microsoft/vscode-python-environments/issues/1335#issuecomment-4023856252

I don't know if this works on Windows, or just on Linux (which I tested). But it solved the problem on 3 different remote machines. I don't know if the default is somehow still used even if overridden (I just assumed that changing it would solve it). 

Please do tell if downgrading go 1.16 and those config changes does help in your case as well. An alternative would be to go before the timeout limits are introduced.

danielkza:
I also observe that the `configure` step reliably fails when any workspace folders are large and/or in network storage. Overriding `python-envs.workspaceSearchPaths` solves the issue, which otherwise means no interpreter can be selected which completely breaks the Python extension.

PR #1389, opened 2026-03-24
enhance issue reporter with title and required description prompts

eleanorjboyd:


src/extension.ts
```diff
@@ -447,12 +447,38 @@ export async function activate(context: ExtensionContext): Promise<PythonEnviron
         ),
         commands.registerCommand('python-envs.reportIssue', async () => {
             try {
+                // Prompt for issue title
+                const title = await window.showInputBox({
+                    title: 'Report Issue - Title',
+                    prompt: 'Enter a brief title for the issue',
+                    placeHolder: 'e.g., Environment not detected, activation fails, etc.',
```
StellaHuang95: LGTM. Does this need to be localized?

eleanorjboyd: good point thank you!

src/extension.ts
```diff
@@ -447,12 +447,38 @@ export async function activate(context: ExtensionContext): Promise<PythonEnviron
         ),
         commands.registerCommand('python-envs.reportIssue', async () => {
             try {
+                // Prompt for issue title
+                const title = await window.showInputBox({
+                    title: l10n.t('Report Issue - Title'),
+                    prompt: l10n.t('Enter a brief title for the issue'),
+                    placeHolder: l10n.t('e.g., Environment not detected, activation fails, etc.'),
+                    ignoreFocusOut: true,
+                });
```
edvilme, line 456:
```suggestion
                }).trim();
```

Will remove leading and trailing whitespaces, which is useful. Especially in the below condition, it can catch string such as `"   "`

src/extension.ts
```diff
@@ -447,12 +447,38 @@ export async function activate(context: ExtensionContext): Promise<PythonEnviron
...
+                    return;
+                }
+
+                // Prompt for issue description
+                const description = await window.showInputBox({
+                    title: l10n.t('Report Issue - Description'),
+                    prompt: l10n.t('Describe the issue in more detail'),
+                    placeHolder: l10n.t('Provide additional context about what happened...'),
+                    ignoreFocusOut: true,
+                });
```
edvilme:
```suggestion
                }).trim();
```

Same as above

edvilme:
Oh didn't realize it was a thennable. Perhaps we should be doing the stripping in the if statements?

src/extension.ts
```diff
@@ -447,12 +447,38 @@ export async function activate(context: ExtensionContext): Promise<PythonEnviron
         ),
         commands.registerCommand('python-envs.reportIssue', async () => {
             try {
+                // Prompt for issue title
+                const title = await window.showInputBox({
+                    title: l10n.t('Report Issue - Title'),
+                    prompt: l10n.t('Enter a brief title for the issue'),
+                    placeHolder: l10n.t('e.g., Environment not detected, activation fails, etc.'),
+                    ignoreFocusOut: true,
+                }).trim();
```
edvilme:
```suggestion
                const title = (await window.showInputBox({
                    title: l10n.t('Report Issue - Title'),
                    prompt: l10n.t('Enter a brief title for the issue'),
                    placeHolder: l10n.t('e.g., Environment not detected, activation fails, etc.'),
                    ignoreFocusOut: true,
                })).trim();
```

src/extension.ts
```diff
@@ -447,12 +447,38 @@ export async function activate(context: ExtensionContext): Promise<PythonEnviron
...
+                    return;
+                }
+
+                // Prompt for issue description
+                const description = await window.showInputBox({
+                    title: l10n.t('Report Issue - Description'),
+                    prompt: l10n.t('Describe the issue in more detail'),
+                    placeHolder: l10n.t('Provide additional context about what happened...'),
+                    ignoreFocusOut: true,
+                }).trim();
```
edvilme:
```suggestion
                const description = (await window.showInputBox({
                    title: l10n.t('Report Issue - Description'),
                    prompt: l10n.t('Describe the issue in more detail'),
                    placeHolder: l10n.t('Provide additional context about what happened...'),
                    ignoreFocusOut: true,
                })).trim();
```

PR #1401, opened 2026-03-26
Choose project for terminal

### LINKED ISSUE #291: 'Python: Create Terminal' doesn't consider multiroot option
I have a workspace with two folders each with their own Python virtual environment. 

Invoking the `Python: Create Terminal` command automatically creates a terminal for the first project in the workspace. We should unify this with the Python environments extension. If there is more than one Python project in the workspace ask which project to create a terminal for.

edvilme:
Fixes #291 

image

This pull request updates the logic for creating terminal commands related to Python environments, focusing on how Python projects are retrieved and passed to the `pickProject` function. The changes improve consistency and clarity in how project lists are handled.

**Improvements to project selection logic:**

* The list of Python projects is now retrieved once and reused in both the undefined context and tree item context branches, ensuring consistency and avoiding redundant API calls. [[1]](diffhunk://#diff-aed4c8c8194da21783959fe3536c5b86f48d4c0755e2ef4092501dbc5e849419L609-R611) [[2]](diffhunk://#diff-aed4c8c8194da21783959fe3536c5b86f48d4c0755e2ef4092501dbc5e849419L644-R645)
* The condition for prompting the user to pick a project has been updated to check if the list of Python projects is non-empty, which prevents unnecessary prompts when there are no projects available.

### DIFF
--- a/src/features/envCommands.ts
+++ b/src/features/envCommands.ts
@@ -30,7 +30,6 @@ import { removePythonProjectSetting, setEnvironmentManager, setPackageManager } 
 
 import { executeCommand } from '../common/command.api';
 import { clipboardWriteText } from '../common/env.apis';
-import { } from '../common/errors/utils';
 import { Pickers } from '../common/localize';
 import { pickEnvironment } from '../common/pickers/environments';
 import {


package-lock.json
```diff
@@ -792,6 +792,7 @@
             "integrity": "sha512-D7DbgGFtsqIPIFMPJwCad9Gfi/hC0PWErRRHFnaCWoEDYi5tQUDiJCTmGUbBiLzjqAck4KcXt9Ayj0CNlIrF+w==",
             "dev": true,
             "license": "BSD-2-Clause",
+            "peer": true,
```
eleanorjboyd: any idea why this had changes to the package lock file?

edvilme: Oh, you're right, I think it is because of the npm version I am using

eleanorjboyd (approved):

PR #1408, opened 2026-03-27
feat: implement fast path resolution for Python environment managers

eleanorjboyd:
- Added fast path resolution logic in SysPythonManager, VenvManager, CondaEnvManager, PipenvManager, and PyEnvManager to directly resolve persisted environments without waiting for full initialization.
- Enhanced error handling with logging for failed resolutions in the fast path.
- Introduced unit tests to validate the fast path behavior across different environment managers, ensuring correct resolution and background initialization handling.

### DIFF
--- a/src/managers/common/fastPath.ts
+++ b/src/managers/common/fastPath.ts
@@ -66,15 +66,21 @@ export async function tryFastPathGet(opts: FastPathOptions): Promise<FastPathRes
         deferred = createDeferred<void>();
         opts.setInitialized(deferred);
         const deferredRef = deferred;
-        Promise.resolve(opts.startBackgroundInit()).then(
-            () => deferredRef.resolve(),
-            (err) => {
-                traceError(`[${opts.label}] Background initialization failed: ${err}`);
-                // Allow subsequent get()/initialize() calls to retry after a background init failure.
-                opts.setInitialized(undefined);
-                deferredRef.resolve();
-            },
-        );
+        try {
+            Promise.resolve(opts.startBackgroundInit()).then(
+                () => deferredRef.resolve(),
+                (err) => {
+                    traceError(`[${opts.label}] Background initialization failed:`, err);
+                    // Allow subsequent get()/initialize() calls to retry after a background init failure.
+                    opts.setInitialized(undefined);
+                    deferredRef.resolve();
+                },
+            );
+        } catch (syncErr) {
+            traceError(`[${opts.label}] Background initialization threw synchronously:`, syncErr);
+            opts.setInitialized(undefined);
+            deferredRef.resolve();
+        }
     }
 
     const fsPath = opts.getProjectFsPath(opts.scope);
@@ -87,9 +93,7 @@ export async function tryFastPathGet(opts: FastPathOptions): Promise<FastPathRes
                 return { env: resolved };
             }
         } catch (err) {
-            traceWarn(
-                `[${opts.label}] Fast path resolve failed for '${persistedPath}', falling back to full init: ${err}`,
-            );
+            traceWarn(`[${opts.label}] Fast path resolve failed for '${persistedPath}', falling back to full init:`, err);
         }
     }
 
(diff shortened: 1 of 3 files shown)


dmitrivMS (dismissed):

src/test/managers/fastPath.get.unit.test.ts
```diff
@@ -0,0 +1,649 @@
+import * as assert from 'assert';
```
StellaHuang95, line 4: Copilot generated:
**Consider parameterizing the integration tests.** The 5 manager suites (VenvManager, CondaEnvManager, SysPythonManager, PyEnvManager, PipenvManager) each repeat ~7 near-identical test cases (fast path success, no persisted env, resolve throws, scope undefined, already initialized, in-progress init, no second deferred). A table-driven approach with a factory per manager would reduce the 649 lines significantly and make it easier to add new managers or test cases. Not blocking, but a maintenance concern.

src/managers/builtin/sysPythonManager.ts
```diff
@@ -145,6 +146,22 @@ export class SysPythonManager implements EnvironmentManager {
     }
 
     async get(scope: GetEnvironmentScope): Promise<PythonEnvironment | undefined> {
+        const fastResult = await tryFastPathGet({
```
StellaHuang95, line 149: Copilot generated:
**Duplicated `getProjectFsPath` lambda.** The expression `(s) => this.api.getPythonProject(s)?.uri.fsPath ?? s.fsPath` is identical across all 5 managers. Consider extracting it as a shared helper (e.g., `getProjectFsPathForScope(api, scope)`) or making it the default behavior inside `tryFastPathGet` by accepting `api` directly.

src/managers/common/fastPath.ts
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * Attempts fast-path resolution for manager.get(): if full initialization hasn't completed yet
+ * and there's a persisted environment for the workspace, resolve it directly via nativeFinder
+ * instead of waiting for full discovery.
+ *
+ * Returns the resolved environment (with an optional new deferred) if successful, or undefined
+ * to fall through to the normal init path.
+ */
+export async function tryFastPathGet(opts: FastPathOptions): Promise<FastPathResult | undefined> {
+    if ((!opts.initialized || !opts.initialized.completed) && opts.scope instanceof Uri) {
```
StellaHuang95: Copilot generated:
**Race condition on concurrent `get()` calls (High).** Two callers entering when `opts.initialized` is `undefined` both pass the guard, both `await` the resolve, then both create separate deferreds and fire separate `startBackgroundInit()`. The second caller's `newDeferred` overwrites the first in the manager, orphaning the first deferred. For Conda/Pipenv/PyEnv, both inits write to `this.collection`, so one overwrites the other's results.

Fix: create and register the deferred *before* any `await`, so a concurrent caller sees it as "in-progress" rather than "not started". One approach: accept a `setInitialized: (d: Deferred<void>) => void` callback in options and call it synchronously before yielding. Alternatively, re-check `opts.initialized` (or have callers check `this._initialized`) after the fast path returns but before assigning.

src/managers/common/fastPath.ts
```diff
@@ -0,0 +1,79 @@
...
+                const resolved = await opts.resolve(persistedPath);
+                if (resolved) {
+                    let newDeferred: Deferred<void> | undefined;
+                    // Ensure full init is running in background (may already be in progress)
+                    if (!opts.initialized) {
+                        newDeferred = createDeferred();
+                        const deferred = newDeferred;
+                        Promise.resolve(opts.startBackgroundInit()).then(
+                            () => deferred.resolve(),
+                            (err) => {
```
StellaHuang95: Copilot generated:
**Silent permanent degradation on background init failure (High).** When `startBackgroundInit` throws, the error handler calls `deferred.resolve()`, marking the manager as "initialized." Subsequent `get()` calls see `_initialized.completed === true`, skip the fast path, call `await this.initialize()` which no-ops (deferred already resolved), then query an empty `fsPathToEnv` map → return `undefined`. The user's environment silently disappears until VS Code restarts.

Consider either: (1) not resolving the deferred on failure (allowing `initialize()` to retry), (2) rejecting so callers can handle it, or (3) resetting `_initialized` to `undefined` on failure so a fresh init attempt is made on the next `get()` call.

Issue #378, opened 2025-05-02
Extension API: How to map it to hatch?


flying-sheep:
Hi! I started implementing https://github.com/flying-sheep/vscode-hatch

Hatch’s env model is:

1. there are a few pre-defined one and you can define more
2. you can create, destroy, and sync any defined env for a project, but none that aren’t pre-defined or defined in Hatch’s config

How should we map this to the extension API?

- `create` only makes sense for defined envs. So show a picker dialog for defined envs?
- `remove` ditto
- `set` & `get` doesn’t really make sense for Hatch. Would be best if these were optional to implement and this extension didn’t have to track the active env
- `refresh` … I don‘t know what it’s for. Is this supposed to re-fetch the list of *defined* envs or *created* envs?
- `getEnvironments` ditto. I think for Hatch it would make more sense if it returned the list of defined envs and we could respond to `onTryingToSetEnvironment` event or so to create it
- `clearCache` I also don’t know what this is for

I think in general it would be nice to have some examples when these are called. Like what UI action leads to `clearCache` being called?

cc @karthiknadig

karthiknadig:
Thanks for working on the Hatch extension. I can provide some guidance here ( I may have to break it into individual comments).

### Real-World Scenarios these APIs Support
The extension needs to handle a wide range of scenarios, including:

- **Single Workspace, Single Environment (1:1)**: Straightforward and very common there’s one obvious Python to use. `get` returns it, `set` is a no-op.

- **Single Workspace, Multiple Environments (1:N)**: This includes test/prod splits or monorepos with many envs (e.g., Azure SDK with 390+). Here, users need to switch between envs for different workflows. `set` is essential to reflect user intent in the IDE, if it is not already obvious from some other mechanism that `get` cannot disambiguate, or if `get` "picks" the wrong one.

- **Multi-Root Workspaces (M:N)**: Each project root might use a different Python. The extension must support selecting environments per workspace folder. This again requires `get` and `set` to coordinate tools like notebooks, language servers, and linters.

Remarks: Rather than designing for just the simplest case, the API enables flexible handling of general M:N environments, with optimizations for common 1:1 workflows.

### How the Extension Uses `get` and `set`

These APIs form the backbone of the Python experience with the extension. When the extension activates, it queries the registered environment manager using `get`. The result is used across tools like the language server, static analysis, formatters, linters, and debuggers.

[...]

PR #1437, opened 2026-04-03
docs: add @remarks to EnvironmentManager interface documenting when each method is called

StellaHuang95:
## Summary

Adds `@remarks` JSDoc sections to all methods on the `EnvironmentManager` interface, documenting **which user actions or internal events trigger each method**. Changes are documentation-only — no functional changes.

## Motivation

Raised in #378: third-party extension authors implementing `EnvironmentManager` know *what* each method does (the existing JSDoc covers that), but not *when* or *why* it gets called. For example, `get()` is described as "retrieves the current Python environment" — but there's no mention that it's called at extension startup, during terminal activation, before script execution, etc.

This makes it hard for authors to reason about their implementation. They end up guessing at lifecycle behavior, which leads to bugs (e.g., not caching in `get()` because they didn't realize it's called frequently).

## What's added

`@remarks` sections on 8 methods of `EnvironmentManager` in both `src/api.ts` and `pythonEnvironmentsApi/src/main.ts`:

| Method | Trigger summary |
|--------|----------------|
| `create` | "Create Environment" command, "+" button in tree view, new package project wizard |
| `remove` | Right-click → "Delete Environment" in tree view |
| `refresh` | Refresh button in the Python Environments view title bar |
| `getEnvironments` | Tree view expansion, environment picker, internally after `refresh` |
| `set` | Environment picker selection, checkmark button in tree view, auto-select after create, startup caching, project removal |
| `get` | Startup, after `set`, terminal activation, script execution, picker display, auto-discovery |
| `resolve` | File picker interpreter selection, `defaultInterpreterPath` resolution, pre-run resolution |
| `clearCache` | "Python: Clear Cache" command |

Each remark was verified by tracing every call site through `InternalEnvironmentManager` → `envManagers.ts` → `envCommands.ts` → `extension.ts` command registrations and `package.json` menu contributions.

## Who this helps

Extension authors building environment managers for tools like Hatch, Pixi, uv, etc. They can now hover over any `EnvironmentManager` method in their IDE and see exactly what user actions will invoke it, making implementation decisions much clearer.

## Files changed

- `src/api.ts` — internal API copy (+45 lines, documentation only)
- `pythonEnvironmentsApi/src/main.ts` — published API package (+45 lines, documentation only)

Refs #378

### DIFF
--- a/pythonEnvironmentsApi/src/main.ts
+++ b/pythonEnvironmentsApi/src/main.ts
@@ -455,8 +455,12 @@ export interface EnvironmentManager {
      * - Selects an environment in the environment picker or clicks the checkmark button in the tree view.
      * - Creates a new environment and the extension auto-selects it.
      *
-     * Also called at extension startup during initial environment selection to cache the active
-     * environment, and when a Python project is removed (with `environment` set to `undefined`).
+     * Also called at extension startup during initial environment selection to initialize or
+     * reconcile the active environment state. This startup call may pass the already-persisted or
+     * previously-selected environment; implementations should treat it as idempotent and avoid
+     * unnecessary work or side effects, including firing change events, unless the selected
+     * environment actually changes. Also called when a Python project is removed (with
+     * `environment` set to `undefined`).
      */
     set(scope: SetEnvironmentScope, environment?: PythonEnvironment): Promise<void>;
 
--- a/pythonEnvironmentsApi/src/main.ts
+++ b/pythonEnvironmentsApi/src/main.ts
@@ -343,6 +343,14 @@ export interface QuickCreateConfig {
 
 /**
  * Interface representing an environment manager.
+ *
+ * @remarks
+ * Methods on this interface are invoked both by the Python Environments extension itself
+ * (in response to UI actions, startup, terminal activation, script execution, and so on)
+ * and directly by other extensions that consume the published API. Any "called when…"
+ * notes on individual methods list representative triggers only — they are not
+ * exhaustive, and the precise set of call sites may evolve over time. Implementations
+ * should focus on the documented contract rather than any specific caller.
  */
 export interface EnvironmentManager {
     /**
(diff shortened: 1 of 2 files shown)


eleanorjboyd:
will review - sorry for the delay

StellaHuang95:
done addressing the comments

pythonEnvironmentsApi/src/main.ts
```diff
@@ -396,27 +396,46 @@ export interface EnvironmentManager {
...
      */
     create?(scope: CreateEnvironmentScope, options?: CreateEnvironmentOptions): Promise<PythonEnvironment | undefined>;
 
     /**
      * Removes the specified Python environment.
      * @param environment - The Python environment to remove.
      * @returns A promise that resolves when the environment is removed.
+     *
+     * @remarks
+     * Called when the user right-clicks an environment in the Python Environments tree view and selects
```
eleanorjboyd: this can also be called by another extension or elsewhere in the extension tho? I am worried about tying these so closely to their current implementations they will become stale quickly and then lead future agents astray

StellaHuang95: That's good point. You're right that (a) these methods are public API that other extensions can call directly, and (b) enumerating current UI call sites will drift as the extension evolves. So here's the improvements I made:

Added an interface-level @remarks on `EnvironmentManager` noting that the methods are called by both this extension and external API consumers, and that any trigger notes on individual methods are representative, not exhaustive.

Replaced the per-method "called when the user clicks X" bullet lists with a short description of what the method is for, followed by "typical triggers include…" framed as examples rather than a spec. No specific buttons, menus, or command IDs are named, so the docs stay accurate if the UI is rewired. Let me know if you have any more feedback. Thanks for reviewing!

eleanorjboyd (approved):

PR #1455, opened 2026-04-14
Cross-session cache for discovered system Python path

eleanorjboyd:
Persist the discovered system Python path (e.g., `/usr/bin/python3` or `C:\Python312\python.exe`) to `globalState`. On next startup, use the cached path immediately and verify it's still valid in the background.

Potential risk: The cache could become stale if the user uninstalls/upgrades Python. Background verification mitigates this — the cached env is used immediately, and if verification fails, a re-scan is triggered. The user would see a brief env switch in the rare case the cache is wrong.


eleanorjboyd:
@copilot resolve the merge conflicts in this pull request

benvillalobos (approved):

PR #1456, opened 2026-04-14
 Don't block on global scope during ext startup

eleanorjboyd:
When at least one workspace folder resolved successfully, don't await the global scope resolution. Run it as a background task instead of blocking startup on it.

For more than 3/4ths of slow sessions where the workspace env resolves fast (via `envPreResolved`), this should make the status bar update near-instantly.

Files opened outside the workspace folder wouldn't have a Python env immediately available. They would resolve moments later when the background task completes.

### DIFF
--- a/src/features/interpreterSelection.ts
+++ b/src/features/interpreterSelection.ts
@@ -407,12 +407,26 @@ export async function applyInitialEnvironmentSelection(
 /**
  * Notify the user when their configured settings could not be applied.
  * Shows a warning message with an option to open settings.
+ * Tracks already-warned settings to avoid duplicate dialogs (e.g., when
+ * the same user-level misconfiguration is hit by both workspace and
+ * deferred global scope resolution).
  */
+const warnedSettings = new Set<string>();
+
+export function resetSettingWarnings(): void {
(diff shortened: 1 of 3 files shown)


src/features/interpreterSelection.ts
```diff
@@ -336,38 +339,69 @@ export async function applyInitialEnvironmentSelection(
...
+    } else {
+        // No workspace folder resolved — global scope is the primary fallback, must await.
+        const globalErrors = await resolveGlobalScope();
+        allErrors.push(...globalErrors);
     }
 
-    // Notify user if any settings could not be applied
+    // Notify user if any settings could not be applied (workspace + global when awaited)
     if (allErrors.length > 0) {
         await notifyUserOfSettingErrors(allErrors);
```
StellaHuang95: Could this be producing duplicate dialogs as the await notifyUserOfSettingErrors(globalErrors) at the same time?

eleanorjboyd: great point- added in a small check as it is a smaller edge case

StellaHuang95 (approved):

Issue #1454, opened 2026-04-14
Conda Environment Audit Findings


StellaHuang95:
**Recently, we’ve received a few user reports about being unable to select certain Conda environments. Without reliable repro steps, it’s been difficult for us to narrow down the root cause, so I asked Copilot to perform a full audit of all Conda-related code paths to help identify potential issues. I’m planning to go through the findings one by one.**

Audit of all conda environment and active environment code paths in vscode-python-environments, motivated by user reports of being unable to select certain conda Python environments as the active interpreter.

**Files audited**: `condaEnvManager.ts`, `condaUtils.ts`, `condaSourcingUtils.ts`, `condaStepBasedFlow.ts`, `condaPackageManager.ts`, `envManagers.ts`, `envCommands.ts`, `fastPath.ts`, `pathUtils.ts`, `deferred.ts`

---

## Bug 1 (Critical): `set()` for global scope does not update in-memory `globalEnv`

**File**: `src/managers/conda/condaEnvManager.ts`, lines 313–314

**What happens**: When a user selects a conda environment as the global interpreter, `set(undefined, environment)` persists the selection to disk via `setCondaForGlobal()`, but **never updates `this.globalEnv`** in memory. Subsequent calls to `get(undefined)` (line 305: `return this.globalEnv`) return the **stale old** environment.

**Comparison with venvManager.ts** (lines 407–414), which correctly does:

```ts
const before = this.globalEnv;
this.globalEnv = environment;
await setVenvForGlobal(environment?.environmentPath.fsPath);
```

**Repro steps**:

1. Open VS Code with no workspace folders (so the scope is global).
2. Select a conda environment as the Python interpreter.
3. The status bar briefly shows the new environment (from the orchestrator's event), but any subsequent call to `getEnvironment(undefined)` — e.g. opening a new terminal, running a script, or the status bar refreshing — returns the **old** environment.
4. The selection appears to **revert** or **flicker** back to the previous environment.

**Why it's a bug**: The orchestrator's `getEnvironment()` method (in `envManagers.ts` line 526) calls `manager.get(scope)`, which returns `this.globalEnv`. Since the in-memory field was never updated, the persisted selection is effectively invisible until VS Code is restarted (when `loadEnvMap` re-reads from persistent state).

**Proposed fix**: Update the in-memory state before persisting, matching the venv manager pattern:

```ts
if (scope === undefined) {
    this.globalEnv = checkedEnv;
[...]

PR #1457, opened 2026-04-14
Fix `set()` for global scope not updating `globalEnv` in memory

StellaHuang95:
Part of https://github.com/microsoft/vscode-python-environments/issues/1454

## Bug

When a user selects a conda environment as the global Python interpreter, `CondaEnvManager.set(undefined, environment)` persists the selection to disk via `setCondaForGlobal()`, but never updates `this.globalEnv` in memory. Subsequent calls to `get(undefined)` return the stale old environment because `get()` returns `this.globalEnv` directly (line 305).

## Why it's a bug

The orchestrator's `getEnvironment()` method calls `manager.get(scope)`, which returns the in-memory `this.globalEnv`. Since the field is never updated after `set()`, the persisted selection is effectively invisible until VS Code is restarted (when `loadEnvMap` re-reads from persistent state). This causes the selection to appear to "revert" or "flicker" back to the previous environment.

## Repro steps

1. Open VS Code with no workspace folders (so the scope is global).
2. Select a conda environment as the Python interpreter.
3. The status bar briefly shows the new environment (from the orchestrator's event), but any subsequent query for the active environment returns the **old** environment.
4. The selection appears to revert on its own.

## Fix

Update `this.globalEnv` in memory before persisting to disk, matching the pattern already used by `venvManager.ts` (lines 407–414):

```ts
if (scope === undefined) {
    this.globalEnv = checkedEnv;
    await setCondaForGlobal(checkedEnv?.environmentPath?.fsPath);
}
```
Before:

After:

## Why this fix is correct

- The venv manager already uses this pattern and works correctly.
- The `checkedEnv` variable already accounts for no-python environments (it's `undefined` if the user declines Python installation), so setting `this.globalEnv = checkedEnv` correctly handles both the set and unset cases.
- No downstream code assumes `globalEnv` is only updated during initialization.

## Tests added

- `condaEnvManager.setGlobal.unit.test.ts`: 4 test cases covering set, persist, clear, and no-python decline scenarios.

### DIFF
--- a/src/test/managers/conda/condaEnvManager.setGlobal.unit.test.ts
+++ b/src/test/managers/conda/condaEnvManager.setGlobal.unit.test.ts
@@ -5,7 +5,7 @@ import { PythonEnvironmentApi } from '../../../api';
 import { CondaEnvManager } from '../../../managers/conda/condaEnvManager';
 import * as condaUtils from '../../../managers/conda/condaUtils';
 import { NativePythonFinder } from '../../../managers/common/nativePythonFinder';
-import { makeMockPythonEnvironment as makeEnv } from '../../mocks/pythonEnvironment';
+import { makeMockCondaEnvironment as makeEnv } from '../../mocks/pythonEnvironment';
 
 function createManager(): CondaEnvManager {
     const manager = new CondaEnvManager(
--- a/src/test/managers/conda/condaEnvManager.setGlobal.unit.test.ts
+++ b/src/test/managers/conda/condaEnvManager.setGlobal.unit.test.ts
@@ -1,29 +1,11 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import assert from 'assert';
 import * as sinon from 'sinon';
-import { Uri } from 'vscode';
-import { PythonEnvironment, PythonEnvironmentApi } from '../../../api';
-import { PythonEnvironmentImpl } from '../../../internal.api';
+import { PythonEnvironmentApi } from '../../../api';
 import { CondaEnvManager } from '../../../managers/conda/condaEnvManager';
 import * as condaUtils from '../../../managers/conda/condaUtils';
 import { NativePythonFinder } from '../../../managers/common/nativePythonFinder';
-
-function makeEnv(name: string, envPath: string, version: string = '3.12.0'): PythonEnvironment {
-    return new PythonEnvironmentImpl(
-        { id: `${name}-test`, managerId: 'ms-python.python:conda' },
-        {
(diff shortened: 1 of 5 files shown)


src/test/managers/conda/condaEnvManager.setGlobal.unit.test.ts
```diff
@@ -0,0 +1,110 @@
...
+import assert from 'assert';
+import * as sinon from 'sinon';
+import { Uri } from 'vscode';
+import { PythonEnvironment, PythonEnvironmentApi } from '../../../api';
+import { PythonEnvironmentImpl } from '../../../internal.api';
+import { CondaEnvManager } from '../../../managers/conda/condaEnvManager';
+import * as condaUtils from '../../../managers/conda/condaUtils';
+import { NativePythonFinder } from '../../../managers/common/nativePythonFinder';
+
+function makeEnv(name: string, envPath: string, version: string = '3.12.0'): PythonEnvironment {
```
eleanorjboyd: can you see if we have a helper function like this elsewhere and if not move this to a common place which many testing files could use

StellaHuang95: Yes the exact `makeEnv(name, envPath, version?)` body was duplicated across three conda unit test files, so I refactored the three conda test files to import `makeMockPythonEnvironment as makeEnv` from the shared location instead of defining it locally:
  - `src/test/managers/conda/condaEnvManager.findEnvironmentByPath.unit.test.ts`
  - `src/test/managers/conda/condaEnvManager.setEvents.unit.test.ts`
  - `src/test/managers/conda/condaEnvManager.setGlobal.unit.test.ts`

src/test/mocks/pythonEnvironment.ts
```diff
@@ -0,0 +1,80 @@
...
+ * Shared across manager and view tests so they agree on the shape of a mock
+ * environment. Only fields that tests commonly need are populated; extend this
+ * helper if additional fields become required.
+ */
+export function createMockPythonEnvironment(options: MockPythonEnvironmentOptions): PythonEnvironment {
+    const {
+        name = 'test-env',
+        envPath,
+        version = '3.12.0',
+        managerId = 'ms-python.python:conda',
```
eleanorjboyd, line 42: this will always be conda right? maybe update the function to specify that so there isn't confusion

StellaHuang95: good point, just renamed

eleanorjboyd (approved):

PR #1459, opened 2026-04-14
Fix wrong Python executable path on Linux/macOS after creating conda environments

StellaHuang95:
Part of https://github.com/microsoft/vscode-python-environments/issues/1454

## Bug

In `createNamedCondaEnvironment` and `createPrefixCondaEnvironment`, the Python executable path is computed as:

```ts
const bin = os.platform() === 'win32' ? 'python.exe' : 'python';
// Then: path.join(envPath, bin)
```

On Linux/macOS, this produces `/path/to/env/python`, but conda installs the Python executable at `/path/to/env/bin/python`.

## Why it's a bug

The environment appears in the VS Code picker and can be selected. However, the internal `execInfo.run.executable` points to a non-existent path. Any operation that invokes Python — running scripts, linting, getting version info — fails silently. Users see a valid-looking conda environment that they can select, but nothing works after selecting it.

The `quickCreateConda` function (line 1225) already had the correct path:
```ts
os.platform() === 'win32' ? path.join(prefix, 'python.exe') : path.join(prefix, 'bin', 'python')
```

## Repro steps

1. On Linux or macOS, create a conda environment through the VS Code UI (either named or prefix type).
2. The environment appears in the picker — select it.
3. Try running a Python script or checking the Python version.
4. Operations fail because the executable path `/path/to/env/python` does not exist.

## Fix

Changed both `createNamedCondaEnvironment` (line 1085) and `createPrefixCondaEnvironment` (line 1169) to use:

```ts
const bin = os.platform() === 'win32' ? 'python.exe' : path.join('bin', 'python');
```

This matches the correct pattern already used in `quickCreateConda`.

## Tests added

- `condaUtils.pythonExePath.unit.test.ts`: 3 test cases verifying correct executable paths for both platform types and both environment types.

### DIFF
--- a/src/managers/conda/condaUtils.ts
+++ b/src/managers/conda/condaUtils.ts
@@ -1082,7 +1082,7 @@ export async function createNamedCondaEnvironment(
         },
         async () => {
             try {
-                const bin = os.platform() === 'win32' ? 'python.exe' : 'python';
+                const bin = os.platform() === 'win32' ? 'python.exe' : path.join('bin', 'python');
                 const output = await runCondaExecutable(runArgs);
                 log.info(output);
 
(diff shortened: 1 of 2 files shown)


eleanorjboyd:
@karthiknadig could you give this a quick once-over since it is windows related

eleanorjboyd (approved):

Issue #1375, opened 2026-03-22
Automated terminal activation kills all running tasks in terminals


MaddyGuthridge:
When I run `uv sync` on a freshly-cloned project, VS Code detects the new virtual environment, and activates it in all terminals.

However, in order to do this, it sends an interrupt signal in all terminals, causing most running tasks to exit. This is incredibly annoying in some projects, especially as the action can sometimes occur randomly up to 2 minutes after the virtual environment is created.

## To reproduce

1. `git clone` a Python project
2. Start up the project, eg by running `uv run [program]`
3. UV creates the virtual environment and starts the program
4. VS Code then immediately kills the freshly-started program

## Why this is incredibly disruptive to my workflow

I work as a university tutor. I am currently assignment marking, and to do so I must `uv run` large numbers of student projects. Every time I do this, VS Code kills their project part-way through testing.

anthonykim1:
Could you try switching to shell startup? Then it wont send any activation command visible, but rather terminal would be ready from the start. 

Please set `python-envs.terminal.autoActivationType` to `shellStartup` in your settings.
Also add `  "python.useEnvironmentsExtension": true,` to your settings.json.

MaddyGuthridge:
Indeed that does work. Could it be a good idea to make that option be the default, as it (a) performs better and (b) is less disruptive.

joneuhauser:
> Please set python-envs.terminal.autoActivationType to shellStartup in your settings.

On an SSH session, that config seems to be disabled. Instead i set    

```
"python.terminal.activateEnvironment": false
```

eleanorjboyd:
🤖 Thanks for the report and for confirming the workaround. Command-based activation is intentionally the default because it can activate environments in any shell; `shellStartup` is the non-disruptive opt-in for supported shells, and `off` is available where no automatic activation is wanted. The same Ctrl+C behavior was subsequently reviewed and closed as designed in https://github.com/microsoft/vscode-python-environments/issues/1570. Please continue using `"python-envs.terminal.autoActivationType": "shellStartup"` for this workflow.

Issue #962, opened 2025-11-06
Experimental roll out shell startup as default


anthonykim1:
Shell startup is the ideal way of activation we want to push for.
We should start rolling it out as default as this is the only reliable way of activation for copilot, etc.

Need to refactor, make sure there are good test coverage, and ensure reliability much as possible.

PR #1462, opened 2026-04-15
Clear the current line when activating in terminal

edvilme:
Ensures that the current terminal line is cleared before the command for activation is passed to the terminal. Before, it would interrupt user input, resulting in corrupted commands like: git pu<activation command>\\n, which not only produces the wrong command, but can lead to unsafe execution.

This occurs on the legacy (de)activation, due to users:
- Having setting 	erminal.integrated.shellIntegration.enabled = false
- Using a shell that does not support shell integration
- Using slower machines, or remote SSH, and the shell integration timeout elapsing before the handshake.

### DIFF
--- a/src/features/terminal/runInTerminal.ts
+++ b/src/features/terminal/runInTerminal.ts
@@ -1,12 +1,12 @@
 import { Terminal, TerminalShellExecution } from 'vscode';
 import { PythonEnvironment, PythonTerminalExecutionOptions } from '../../api';
+import { traceLog } from '../../common/logging';
 import { createDeferred } from '../../common/utils/deferred';
 import { onDidEndTerminalShellExecution } from '../../common/window.apis';
 import { ShellConstants } from '../common/shellConstants';
 import { identifyTerminalShell } from '../common/shellDetector';
 import { quoteArgs } from '../execution/execUtils';
-import { normalizeShellPath } from './shells/common/shellUtils';
-import { traceLog } from '../../common/logging';
+import { getClearLineSequence, normalizeShellPath } from './shells/common/shellUtils';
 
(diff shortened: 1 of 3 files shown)


edvilme:
Ref #1375

edvilme:
Would be superseded by #962 but this consider the case for using the legacy activation

eleanorjboyd:
going to defer this to @anthonykim1 to make a decision since this is his area of expertise. I am inclined to say we hold on this as we get the terminal experiments out and the stuff anthony has been working on since that should address the same problem

anthonykim1:
Thanks for the PR :) 

I don't think we should do this, as clearing line is pretty aggressive compared to `^c` keyboard interrupt which is part of shell integration's executeCommand.
Even with the keyboard interrupt there has been much complains about how "aggressive" the behavior is.

Big downside of clearing line is that user may be typing something but clearing before the explicit activation command would result in that typing experience "disappearing". 
Aside from user typing, we would also have to assume that other extensions may be trying to send command upon launch of terminal so clearing would be effectively "undo"-ing their behavior as well. 

Keyboard interrupt behavior is more expected as it shows what user was typing (or extension sending command) before the `^C` is inserted. 

As far as shell integration, majority of our users are on it, and it's been the default ever since its introduction.
It's what also powers copilot terminal experience, so we'd want to focus on shell integration enabled experience.

In case auto-injection fails, we do provide way for manual installation here : https://code.visualstudio.com/docs/terminal/shell-integration#_manual-installation 

It's great that you called out https://github.com/microsoft/vscode-python-environments/issues/962 since we're planning roll shell startup as the default once env extension is fully rolled out.

Issue #997, opened 2025-11-20
[Python Environments] Virtual environment activation is slow


cecheta:
Type: <b>Performance Issue</b>

Activating the virtual environment seems to take a long time (about 5 seconds) after opening a new shell. This is quite annoying, as the `source` command that gets executed in the shell will interrupt any other command that you are currently trying to write.

![Image](https://github.com/user-attachments/assets/8c2fd2ad-2b6c-4875-96c3-f8a7f099f79c)

<details>
<summary>Environment Information</summary>

```
Extension Version: 1.12.0
Python Extension Version: 2025.18.0

Registered Environment Managers (4):
  - ms-python.python:system (Global)
  - ms-python.python:venv (venv)
  - ms-python.python:pyenv (PyEnv)
  - ms-python.python:poetry (Poetry)

Total Available Environments: 6
Environment Details:
  1. Python 3.13.9 (homebrew) (3.13.9.final.0) - /home/linuxbrew/.linuxbrew/bin/python3.13
  2. Python 3.12.3 (3.12.3.final.0) - /bin/python3
  3. Python 3.12.3 (3.12.3.final.0) - /usr/bin/python3
  4. .venv (3.13.9) (3.13.9) - /home/cecheta/Code/playground/.venv/bin/python
  5. pyenv (3.13.9) (3.13.9) - /home/cecheta/.pyenv/versions/3.13.9
  6. pyenv (3.12.9) (3.12.9) - /home/cecheta/.pyenv/versions/3.12.9

Python Projects (1):
  1. /home/cecheta/Code/playground
     Environment: .venv (3.13.9)

Extension Settings:
  Default Environment Manager: ms-python.python:venv
  Default Package Manager: ms-python.python:pip
Auto-activation is "command". Activation based on first 'py-env.terminal.autoActivationType' setting which is 'command' and 'python.terminal.activateEnvironment' if the first is undefined which is 'true'.

```

</details>

Extension version: 1.12.0
VS Code version: Code 1.106.2 (1e3c50d64110be466c0b4a45222e81d2c9352888, 2025-11-19T16:56:50.023Z)
OS version: Windows_NT x64 10.0.26200
Modes:
[...]

PR #1501, opened 2026-05-01
fix: skip shell-integration wait for shells without SI support (fixes #1391)

### LINKED ISSUE #1391: Terminal activation is slow on nushell
#997 says to use the `"python-envs.terminal.autoActivationType": "shellStartup"` mode, which isn’t available on nushell.

But I’m sure things can be better. It takes 4 seconds or more until VS Code executes my shell startup command. Why?

StellaHuang95:
## Summary

Fixes #1391 — terminal activation is slow on nushell (and any other shell VS Code doesn't provide shell integration for).

When a Python terminal opens in `command` activation mode, `waitForShellIntegration()` is awaited before sending the activation command. For shells VS Code does **not** provide shell integration for — `nu`, `cmd`, `csh`, `tcsh`, `ksh`, `xonsh` — the `onDidChangeTerminalShellIntegration` event never fires, the prompt-pattern heuristic is unreliable, and the wait runs out the full 5 s default timeout (per `getShellIntegrationTimeout()` when `terminal.integrated.shellIntegration.enabled` is true). Once the wait completes, activation falls back to `terminal.sendText` anyway — so the delay is pure overhead and visibly disrupts the user's typing in the new terminal.

## Root cause

`waitForShellIntegration()` in [`src/features/terminal/utils.ts`](src/features/terminal/utils.ts) doesn't consult the codebase's existing source of truth for SI-capable shells (`shellIntegrationSupportedShells` in [`src/features/terminal/shells/common/shellUtils.ts`](src/features/terminal/shells/common/shellUtils.ts)). For `nu`, that means it always times out.

## Fix

After the existing `terminal.shellIntegration` early-return in `waitForShellIntegration`, identify the shell via `identifyTerminalShell()` and return `false` immediately when the shell type is known and **not** in `shellIntegrationSupportedShells`. Wrapped in `try/catch` so any detection failure (or `'unknown'` result) falls through to the existing `Promise.race` body — strictly additive: best case faster, worst case identical to today.

This:

- Reuses the codebase's single source of truth (`shellIntegrationSupportedShells`), so adding a new SI-capable shell automatically benefits both this code and `shouldUseProfileActivation`.
- Returns the same value (`false`) the function would have eventually returned — semantics unchanged for all callers, only latency differs.
- Preserves all existing behavior for supported shells (`pwsh`, `bash`, `gitbash`, `fish`, `zsh`).
- Defensive `try/catch` and explicit `'unknown'` check make the change non-regressing even if shell detection fails.

For nu users, this turns the ~5 s freeze before `overlay use ...activate.nu` runs into an imperceptible delay.

### Manual

On `main` with `nu` configured as default profile and `python-envs.terminal.autoActivationType: "command"`:
- Open new terminal → ~5 s gap between prompt appearing and activation command running.

On this branch, same setup:
- Open new terminal → activation command runs immediately after the prompt.
- `Python Environments` output channel logs: `Shell 'nu' does not support shell integration; skipping wait.`

Verified the fix does **not** alter behavior for `pwsh` / `bash` (they still wait for SI to come up, which they do quickly).

### Automated

8 new unit tests in `Terminal Utils - waitForShellIntegration` cover:

- `undefined` terminal → `false` immediately
- `terminal.shellIntegration` already set → `true` immediately, no shell detection
- `nu` → `false` immediately, no event listeners registered
- `cmd` → `false` immediately
- `csh` / `tcsh` / `ksh` / `xonsh` → all `false` immediately
- `bash` → falls through to event race, resolves `true` when SI event fires
- `unknown` shell → falls through to race
- `identifyTerminalShell` throws → falls through to race (defensive)

```
npm run lint           # clean
npm run compile-tests  # clean
npm run unittest       # 992 passing (including the 8 new tests), 2 pending
```

## Related

- #997 — original "Virtual environment activation is slow" tracking issue, where the maintainer explicitly noted: *"In case shell does not support dynamic evals like `nu`, it will fall back to command activation."* This PR makes that command-activation fallback fast.

## Risk

Low. Fully covered by tests. The change is gated by `try/catch` and an explicit allow-list, so the worst case (detection fails or returns `'unknown'`) is identical to current behavior.

### DIFF
--- a/src/features/terminal/utils.ts
+++ b/src/features/terminal/utils.ts
@@ -1,10 +1,13 @@
 import * as path from 'path';
 import { Disposable, env, tasks, Terminal, TerminalOptions, Uri } from 'vscode';
 import { PythonEnvironment, PythonProject, PythonProjectEnvironmentApi, PythonProjectGetterApi } from '../../api';
+import { traceVerbose } from '../../common/logging';
 import { timeout } from '../../common/utils/asyncUtils';
 import { createSimpleDebounce } from '../../common/utils/debounce';
 import { onDidChangeTerminalShellIntegration, onDidWriteTerminalData } from '../../common/window.apis';
 import { getConfiguration, getWorkspaceFolders } from '../../common/workspace.apis';
+import { identifyTerminalShell } from '../common/shellDetector';
+import { shellIntegrationSupportedShells } from './shells/common/shellUtils';
 
 export const SHELL_INTEGRATION_TIMEOUT = 500; // 0.5 seconds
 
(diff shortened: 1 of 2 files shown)


StellaHuang95:
@anthonykim1 I made some changes around shell integration, would you mind taking a look as well? Thanks!

eleanorjboyd (approved):

bpasero (approved):

anthonykim1:
@StellaHuang95 This looks good, thank you

Issue #1491, opened 2026-04-28
“venv” enviroment manager steals environment managed by other managers


flying-sheep:
related but not the same as #1471

## Environment data

-   Python Environments extension version: 1.28.0
-   Python extension (`ms-python.python`) version: 2026.4.0
-   VS Code version (Help → About): 1.117.0
-   OS and version: irrelevant
-   Python version (& distribution if applicable, e.g. Anaconda): 3.14
-   Environment manager in use (`venv` / `conda` / `pyenv` / `poetry` / `pipenv` / `system` / `uv` / other): hatch
-   Shell (bash / zsh / fish / pwsh / cmd / other): irrelevant
-   Remote / container scenario (none / WSL / SSH Remote / Dev Container / Codespaces): none
-   Workspace type (single folder / multi-root / mono-repo): single folder
-   Is this a regression? If yes, last known working extension version: no

## Repro Steps

1. Install hatch environment extension from VSIX: https://github.com/flying-sheep/hatch-code/releases/tag/v0.1.1
2. open a project and a python file
3. select a hatch environment using the python-environments extension, e.g. `default`
4. run the `python.setInterpreter` command and look at the offered venvs

the reason I need to use that command is #1485, i.e. `ruff` and `ty` don’t see the environment selected for the workspace, so I need top set both workspace and global environment separately. Really bad UX, please help how to fix that!

## Expected behavior

the currently selected enviroment is listed as an environment provided by the hatch env manager

## Actual behavior

the currently selected environment is listed as provided by the “venv” env manager

## Logs

After selecting the `hatch-test.py3.14-stable` environment for the workspace, I cleared the logs, then ran the `python.setInterpreter` command. These are the logs generated by that action:

```
2026-04-28 09:47:39.248 [info] Refreshing Pyenv Environments
2026-04-28 09:47:39.248 [info] Refreshing pyenv environments
2026-04-28 09:47:39.248 [info] Refreshing Pipenv Environments
2026-04-28 09:47:39.248 [info] Refreshing pipenv environments
2026-04-28 09:47:39.248 [info] Refreshing Poetry Environments
2026-04-28 09:47:39.248 [info] Refreshing poetry environments
2026-04-28 09:47:39.248 [info] Refreshing Conda Environments
2026-04-28 09:47:39.248 [info] Refreshing conda environments (hardRefresh=true)
2026-04-28 09:47:39.265 [info] Discovered manager: (Poetry) /usr/bin/poetry
[...]

Issue #1485, opened 2026-04-26
What is the global environment for?


flying-sheep:
Hi, using my hatch environment extension, I ran into an usability issue: there are extensions that use the global Python environment such `ruff` and `ty`.

When setting a workspace environment, these extensions still use the global env, for which there is no clear indicator in the UI. How should this be fixed:
- should my extension just also set the global environment whenever setting a project-scoped environment?
- should these extensions be changed so they stop using `api.environments.onDidChangeActiveEnvironmentPath`, which seems to track the global environment?

Maybe I just don’t really know what the `'global'` scope is for … any guidance?

Issue #1471, opened 2026-04-21
regular venv environment manager duplicates Hatch-managed environments


flying-sheep:
See here: the recommended environment on top (which is the currently selected one) comes from my Hatch extension’s env manager, but the “venv” env manager shows the same environment. The only difference I can see is that it adds `bin/python` to the path.

Should I set the path of the environment to the binary in my extension to prevent this?

If so, which one is used to deduplicate? Are they deduplicated at all or will the `venv` env manager just pick up and duplicate everything it ever sees?

https://github.com/microsoft/vscode-python-environments/blob/3c072d1abc3e94ff4f96c8130344bde9716d8797/src/api.ts#L178

https://github.com/microsoft/vscode-python-environments/blob/3c072d1abc3e94ff4f96c8130344bde9716d8797/src/api.ts#L188

https://github.com/microsoft/vscode-python-environments/blob/3c072d1abc3e94ff4f96c8130344bde9716d8797/src/api.ts#L208

<img width="829" height="280" alt="Image" src="https://github.com/user-attachments/assets/47ebcea9-b31d-4c0d-a08f-5f6c3ee2cd6e" />

eleanorjboyd:
@StellaHuang95 or @flying-sheep is this resolved?

StellaHuang95:
@eleanorjboyd Yes, this should be resolved, the duplication had 2 independent casues:
1. There's fix made on PET side such that PET now classifies `Hatch` envs as `Hatch`, not `Venv`.
2. `VenvManager.resolve()` no longer mutates state.

Going to close this out, please reopen it if the issue still persists in the latest version. Thanks.

PR #1507, opened 2026-05-05
fix: stop resolve() from leaking foreign envs (e.g. hatch) into the venv collection

StellaHuang95:
### Context

@flying-sheep reported three related bugs — #1471, #1485, #1491. The two venv-related ones both traced to the same root cause in `VenvManager.resolve()`.

Hatch envs structurally look like venvs (they have a `pyvenv.cfg`), so until very recently PET classified them as `Venv` and they passed through `resolveVenvPythonEnvironmentPath`. Every successful `resolve()` call also mutated `this.collection` via `addEnvironment(resolved, true)`. So whenever anything called `resolve()` for a hatch path — `Python: Select Interpreter`, Pylance startup, a test runner asking for execInfo, another extension querying the env — the hatch env got persisted under venv.

The original side effect dates back to commit [9363bc1](https://github.com/microsoft/vscode-python-environments/commit/9363bc1301b6fcfc7dbec924534f680568bc1bf5) (Oct 2024), when the extension only had `system` and `venv` managers and the author assumed any successful `resolve()` was a newly discovered venv worth caching. That assumption stopped being true once the API supported third-party managers.

### Upstream fix has now landed in PET

As of [microsoft/python-environment-tools#460](https://github.com/microsoft/python-environment-tools/pull/460) (commit [d581272](https://github.com/microsoft/python-environment-tools/commit/d58127285a6571fe4e0589a5c4f9220e559207b5)), PET now has a dedicated `pet-hatch` locator:

- New `LocatorKind::Hatch` and `PythonEnvironmentKind::Hatch`.
- Registered **before** the `Venv` locator, so it claims hatch envs first.
- Matches hatch's real storage layout (`<data_dir>/env/virtual/<project_name>/<project_id>/<venv_name>`), honors `HATCH_DATA_DIR`, and reads `[tool.hatch.dirs.env].virtual` from `pyproject.toml` / `hatch.toml`.

That fixes the structural source of the hatch-as-venv misclassification.

### What this PR does

This PR is a complementary, defense-in-depth correction in `src/managers/builtin/venvManager.ts`:

```diff
 if (resolved) {
     if (resolved.envId.managerId === `${PYTHON_EXTENSION_ID}:venv`) {
-        // This is just like finding a new environment or creating a new one.
-        // Add it to collection, and trigger the added event.
-        this.addEnvironment(resolved, true);
-
         // We should only return the resolved env if it is a venv.
         return resolved;
     }
 }
 return undefined;
```

`resolve()` is now a pure query: *"given a path, are you the manager for it?"* It still returns the resolved env (every consumer of `resolve()` keeps working), it just stops mutating state. Cached venvs continue to land in `this.collection` through the normal discovery and creation paths, which is the only path that should be writing to it.

### Why keep this PR alongside the PET fix

Even with PET correctly classifying hatch envs, `VenvManager.resolve()` mutating `this.collection` is an invariant violation worth correcting:

- **General cross-manager overlap.** Anything that produces a `pyvenv.cfg`-shaped env (tox, nox, custom bootstrap scripts, future tools) would re-introduce the same class of leak. With `resolve()` as a pure query, no future overlap can silently grow the venv collection.
- **Locator-ordering regressions.** If PET's locator priority ever shifts again, this change keeps the venv manager from quietly absorbing whatever lands in its lap.

### DIFF
--- a/src/managers/builtin/venvManager.ts
+++ b/src/managers/builtin/venvManager.ts
@@ -336,14 +336,15 @@ export class VenvManager implements EnvironmentManager {
                     environment: env,
                 }));
 
-                this.collection = (await findVirtualEnvironments(
-                    hardRefresh,
-                    this.nativeFinder,
-                    this.api,
-                    this.log,
-                    this,
-                    scope ? [scope] : undefined,
-                )) ?? [];
+                this.collection =
+                    (await findVirtualEnvironments(
+                        hardRefresh,
+                        this.nativeFinder,
+                        this.api,
+                        this.log,
+                        this,
+                        scope ? [scope] : undefined,
+                    )) ?? [];
                 await this.loadEnvMap();
 
                 const added = this.collection.map((env) => ({ environment: env, kind: EnvironmentChangeKind.add }));
@@ -500,10 +501,6 @@ export class VenvManager implements EnvironmentManager {
         );
         if (resolved) {
             if (resolved.envId.managerId === `${PYTHON_EXTENSION_ID}:venv`) {
-                // This is just like finding a new environment or creating a new one.
-                // Add it to collection, and trigger the added event.
-                this.addEnvironment(resolved, true);
-
                 // We should only return the resolved env if it is a venv.
                 // Fall through an return undefined if it is not a venv
                 return resolved;


eleanorjboyd:
From what I can tell the PR is a valid defensive fix, but the root cause is that PET has no concept of Hatch environments. Since hatch envs have pyvenv.cfg, the Venv locator here claims them. We would need to add Hatch as supported in PET.

This fix is still useful as defense-in-depth (resolve shouldn't mutate state), let me look at this more to see if we still want this PR then, in addition to any PET changes.

flying-sheep:
> the root cause is that PET has no concept of Hatch environments. […] We would need to add Hatch as supported in PET.

Hatch environments are usually regular venvs that exist in specific locations, but not always. Hatch can use any kind of environment via plugin, e.g. you could make Hatch create conda envs instead, or docker images.

So I’d phrase that more generally: PET has no concept of env managers that manage a specific subset of envs of a known type – in other words, project-aware env managers like Hatch should be able to claim certain envs with a higher specificity than more generic / base env managers (which aren’t project-aware).

PS: don’t you have the same issue with Poetry already? It also just creates regular venvs, right?

eleanorjboyd:
@flying-sheep  the following is merged in the PET extension which should help make sure Hatch environments do not get picked up: [https://github.com/microsoft/python-environment-tools/commit/d58127285a6571fe4e0589a5c4f9220e559207…](https://github.com/microsoft/python-environment-tools/commit/d58127285a6571fe4e0589a5c4f9220e559207b5). @StellaHuang95 could you go through this PR once more now and just update it and the description given what we discussed about PET needing to recognize the env type but then about this change being a good additional stop-gate.

StellaHuang95:
> @flying-sheep the following is merged in the PET extension which should help make sure Hatch environments do not get picked up: [https://github.com/microsoft/python-environment-tools/commit/d58127285a6571fe4e0589a5c4f9220e559207…](https://github.com/microsoft/python-environment-tools/commit/d58127285a6571fe4e0589a5c4f9220e559207b5). @StellaHuang95 could you go through this PR once more now and just update it and the description given what we discussed about PET needing to recognize the env type but then about this change being a good additional stop-gate.

@eleanorjboyd Sure, now that fix on the PET side is in, hatch envs get classified correctly at the source, so I've reframed the description of the PR to be more generic, basically `resolve()` just stops mutating `this.collection`. Same one-line change, but it's there to catch anything else that ends up looking like a venv (tox, nox, hatch plugin envs, random custom scripts) rather than being the fix for hatch specifically.

eleanorjboyd (approved):

PR #1518, opened 2026-05-10
Export ESM module

flying-sheep:
Allow bundlers to pick the best version, do tree-shaking, …

I checked that the below works with typescript, but I didn’t check if npmjs.com is smart enough to add the little “TS” icon. I assume so according to their docs, but if not, we can re-add the old `"types"` field.

<details>
<summary>Details about <code>exports</code></summary>

According to https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-7.html#packagejson-exports-imports-and-self-referencing:

> if you write an import from an ES module, it will look up the import field, and from a CommonJS module, it will look at the require field. If it finds them, it will look for a corresponding declaration file. If you need to point to a different location for your type declarations, you can add a "types" import condition.
> […]
> It’s important to note that the CommonJS entrypoint and the ES module entrypoint each needs its own declaration file, even if the contents are the same between them.

and according to npmjs.com’s [RFC 46](https://github.com/npm/rfcs/blob/6f61e5a2394040fa5843024d799db4ac675378d5/implemented/0046-add-types-metadata-to-package-json.md):

> If there is not the explicit fields: Resolving the main with both TypeScript and Flow's extra file resolvers (e.g. when distribution/danger.js look for the files ./distribution/danger.js.flow and ./distribution/danger.d.ts)

So in summary, TypeScript 4.7+ should find the `.d.ts` files, and npmjs.com should therefore resolve it to, but others had problems with that: https://github.com/browserslist/browserslist-useragent-regexp/issues/1545

</details>

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -11,7 +11,7 @@
     // Turn off tsc task auto detection since we have the necessary tasks as npm scripts
     "typescript.tsc.autoDetect": "off",
     "editor.formatOnSave": true,
-    "[typescript]": {
+    "[typescript][json][jsonc]": {
         "editor.defaultFormatter": "esbenp.prettier-vscode",
         "editor.codeActionsOnSave": {
             "source.organizeImports": "explicit"
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -11,7 +11,7 @@
     // Turn off tsc task auto detection since we have the necessary tasks as npm scripts
     "typescript.tsc.autoDetect": "off",
     "editor.formatOnSave": true,
-    "[typescript][json][jsonc]": {
+    "[typescript]": {
         "editor.defaultFormatter": "esbenp.prettier-vscode",
         "editor.codeActionsOnSave": {
             "source.organizeImports": "explicit"
--- a/api/package.json
+++ b/api/package.json
@@ -36,6 +36,7 @@
     },
     "scripts": {
         "prepublishOnly": "echo \"⛔ Can only publish from a secure pipeline ⛔\" && node -e \"process.exitCode = 1\"",
+        "preinstall": "npm run all:publish",
         "prepack": "npm run all:publish",
         "all:publish": "git clean -xfd . && npm install && npm run compile",
         "compile": "npm run compile:cjs && npm run compile:esm",
--- a/api/package.json
+++ b/api/package.json
@@ -36,7 +36,6 @@
     },
     "scripts": {
         "prepublishOnly": "echo \"⛔ Can only publish from a secure pipeline ⛔\" && node -e \"process.exitCode = 1\"",
-        "preinstall": "npm run all:publish",
         "prepack": "npm run all:publish",
         "all:publish": "git clean -xfd . && npm install && npm run compile",
         "compile": "npm run compile:cjs && npm run compile:esm",
--- a/api/package.json
+++ b/api/package.json
@@ -14,11 +14,11 @@
     "exports": {
         "import": {
             "types": "./out/esm/main.d.ts",
-            "default": "./out/esm/main.js"
+            "default": "./out/esm/main.mjs"
         },
         "require": {
             "types": "./out/cjs/main.d.ts",
-            "default": "./out/cjs/main.js"
+            "default": "./out/cjs/main.cjs"
         }
     },
     "engines": {
@@ -38,14 +38,15 @@
         "prepublishOnly": "echo \"⛔ Can only publish from a secure pipeline ⛔\" && node -e \"process.exitCode = 1\"",
         "prepack": "npm run all:publish",
         "all:publish": "git clean -xfd . && npm install && npm run compile",
-        "compile": "npm run compile:cjs && npm run compile:esm",
-        "compile:cjs": "tsc -b ./tsconfig.cjs.json",
-        "compile:esm": "tsc -b ./tsconfig.esm.json",
+        "compile": "npm run compile:esm && npm run compile:cjs",
+        "compile:esm": "tsc -b ./tsconfig.esm.json && renamer 'out/esm/**' -e ext -f js -r mjs",
+        "compile:cjs": "tsc -b ./tsconfig.cjs.json && renamer 'out/cjs/**' -e ext -f js -r cjs",
         "clean": "node -e \"const fs = require('fs'); fs.rmSync('./out', { recursive: true, force: true });\""
     },
     "devDependencies": {
         "@types/node": "^22.0.0",
         "@types/vscode": "^1.99.0",
+        "renamer": "^5.0.3",
         "typescript": "^5.1.3"
     }
 }
--- a/api/package.json
+++ b/api/package.json
@@ -11,8 +11,11 @@
         "API",
         "Environments"
     ],
-    "main": "./out/main.js",
-    "types": "./out/main.d.ts",
+    "exports": {
+        "import": "./out/esm/main.js",
+        "require": "./out/cjs/main.js",
+        "types": "./out/types/main.d.ts"
+    },
     "engines": {
         "node": ">=22.21.1",
         "vscode": "^1.110.0"
@@ -30,7 +33,9 @@
         "prepublishOnly": "echo \"⛔ Can only publish from a secure pipeline ⛔\" && node -e \"process.exitCode = 1\"",
         "prepack": "npm run all:publish",
         "all:publish": "git clean -xfd . && npm install && npm run compile",
-        "compile": "tsc -b ./tsconfig.json",
+        "compile": "npm run compile:cjs && npm run compile:esm",
+        "compile:cjs": "tsc -b ./tsconfig.cjs.json",
+        "compile:esm": "tsc -b ./tsconfig.esm.json",
         "clean": "node -e \"const fs = require('fs'); fs.rmSync('./out', { recursive: true, force: true });\""
     },
     "devDependencies": {
--- a/api/package.json
+++ b/api/package.json
@@ -39,14 +39,14 @@
         "prepack": "npm run all:publish",
         "all:publish": "git clean -xfd . && npm install && npm run compile",
         "compile": "npm run compile:esm && npm run compile:cjs",
-        "compile:esm": "tsc -b ./tsconfig.esm.json && renamer 'out/esm/**' -f .js -r .mjs --force",
-        "compile:cjs": "tsc -b ./tsconfig.cjs.json && renamer 'out/cjs/**' -f .js -r .cjs --force",
+        "compile:esm": "tsc -b ./tsconfig.esm.json && mve out/esm/main.js out/esm/main.mjs",
+        "compile:cjs": "tsc -b ./tsconfig.cjs.json && mve out/cjs/main.js out/cjs/main.cjs",
         "clean": "node -e \"const fs = require('fs'); fs.rmSync('./out', { recursive: true, force: true });\""
     },
     "devDependencies": {
         "@types/node": "^22.0.0",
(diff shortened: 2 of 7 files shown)


api/package.json
```diff
@@ -11,8 +11,16 @@
...
+    "exports": {
+        "import": {
+            "types": "./out/esm/main.d.ts",
+            "default": "./out/esm/main.js"
+        },
+        "require": {
+            "types": "./out/cjs/main.d.ts",
+            "default": "./out/cjs/main.js"
+        }
+    },
```
StellaHuang95, line 23: **🔴 The ESM build is missing a `"type": "module"` signal — Node loads it via the typeless-reparse fallback.**

The `exports` field tells Node *which* file to open, but not *how* to interpret it. Node decides interpretation from the file extension or the nearest `package.json`'s `"type"` field. Here:

- `out/esm/main.js` contains real ESM syntax (verified locally: `import { extensions, } from 'vscode';` and `export var ...`)
- The extension is `.js` (ambiguous), and there's no `"type": "module"` anywhere
- Default for `.js` without a `"type"` is CJS

I built and loaded the package locally with Node 22.18 and got:

```
[MODULE_TYPELESS_PACKAGE_JSON] Warning: Module type of .../out/esm/main.js
is not specified and it doesn't parse as CommonJS.
Reparsing as ES module because module syntax was detected.
This incurs a performance overhead.
```

Why this matters:
1. It currently works **only by accident**, via Node's fallback. The package emits a runtime warning every time it's loaded.
2. Stricter consumers will outright reject it — notably TypeScript with `"moduleResolution": "node16"`/`"nodenext"` and some bundlers in strict mode. Those are precisely the modern ESM consumers this PR is trying to help.
3. The reparse has a measurable per-load cost.

**Suggested fix (smallest):** emit a tiny `out/esm/package.json` containing `{ "type": "module" }` (and optionally `out/cjs/package.json` with `{ "type": "commonjs" }`) as part of the build. A one-line postcompile step is enough, e.g.:

```json
"compile:esm": "tsc -b ./tsconfig.esm.json && node -e \"require('fs').writeFileSync('out/esm/package.json', '{\\\"type\\\":\\\"module\\\"}')\""
```

Alternatively, switch `tsconfig.esm.json` to `"module": "node16"`/`"nodenext"` and let TS emit `.mjs` files (see related comment on `tsconfig.esm.json`).

api/tsconfig.esm.json
```diff
@@ -0,0 +1,7 @@
+{
+    "extends": "./tsconfig.base.json",
+    "compilerOptions": {
+        "module": "esnext",
```
StellaHuang95, line 4: **🟡 `"module": "esnext"` is the wrong target for a package shipped to Node — it's a footgun for future contributors.**

`esnext` emits ESM syntax but does **not** enforce Node's ESM resolution rules. The most important rule TypeScript skips under `esnext` is that **relative imports must include the file extension**:

```ts
// Compiles fine under "module": "esnext"
import { thing } from './helper';

// Required by real Node ESM at runtime
import { thing } from './helper.js';
```

Today this happens to work because `main.ts` has zero relative imports — the only import is `'vscode'`. But the next contributor who splits a helper into a separate file and writes `import { x } from './helper'` will:

- Compile cleanly ✅
- Pass review ✅
- Get published ✅
- **Crash at runtime in any pure-Node ESM consumer** with `Cannot find module './helper'` ❌

`"module": "node16"` (or `"nodenext"`) makes TypeScript enforce the `.js` extension at compile time, catching the bug immediately. It also makes TS emit `.mjs` files, which — bonus — eliminates the `"type": "module"` issue I flagged on `package.json` since the extension itself signals ESM to Node.

Recommend `"module": "node16"` here.

.vscode/settings.json
```diff
@@ -11,7 +11,7 @@
     // Turn off tsc task auto detection since we have the necessary tasks as npm scripts
     "typescript.tsc.autoDetect": "off",
     "editor.formatOnSave": true,
-    "[typescript]": {
+    "[typescript][json][jsonc]": {
```
StellaHuang95: **🟡 Unrelated scope creep — please split this out of the PR.**

This change makes Prettier the default formatter for **all JSON/JSONC** files in the repo, not just TypeScript. It's the reason the diff also contains the cosmetic reformat of the main `package.json` (`python-envs.workspaceSearchPaths.default` getting expanded across multiple lines), which has nothing to do with shipping an ESM build.

A few concerns:

1. **Unrelated to the PR's stated goal.** The PR is "Export ESM module" — reviewers and future git-archaeologists won't expect editor/formatting policy changes here.
2. **Imposes a team-wide preference unilaterally.** Repo-level `.vscode/settings.json` applies to every contributor. A change to the default JSON formatter deserves its own discussion rather than riding along in a packaging PR.
3. **Causes ongoing diff churn.** Once merged, every contributor with format-on-save will reformat JSON files according to Prettier's rules, adding noise to unrelated future diffs.

Could you drop this hunk (and the resulting reformat in `package.json`) and propose it as a separate PR if you'd like to discuss it?

StellaHuang95 (commented):
Thanks for tackling this — the dual CJS+ESM publishing direction is the right move and will help modern bundler consumers via tree-shaking.

I built and loaded the package locally before reviewing. Left three inline comments:

1. **🔴 ESM build isn't signaled as ESM to Node** (`api/package.json`) — Node loads `out/esm/main.js` only via the `MODULE_TYPELESS_PACKAGE_JSON` reparse fallback. Stricter consumers (TS `node16`/`nodenext`, some bundlers) will reject it. Easiest fix: emit `out/esm/package.json` with `{"type":"module"}` during the build.
2. **🟡 `"module": "esnext"` is a footgun** (`api/tsconfig.esm.json`) — works today only because there are no relative imports. The first `import './helper'` a future contributor adds will compile fine and crash at runtime in Node ESM. Recommend `"module": "node16"` so TS enforces the `.js` extension.
3. **🟡 Scope creep in `.vscode/settings.json`** — the formatter change (and resulting `package.json` reformat) is unrelated to ESM publishing and changes team-wide editor behavior. Please split it out.

Two minor non-blockers: the two emitted `.d.ts` files are byte-identical (43 KB shipped twice — could be deduped via a single shared declaration), and the two builds run sequentially when they could run in parallel.

Happy to re-review once the above is addressed.

api/tsconfig.esm.json
flying-sheep: Unfortunately that’s not true, typescript only ever creates `.js` files (and `.d.ts` files): https://github.com/microsoft/TypeScript/issues/18442#issuecomment-329387971

To switch to `.mjs` (and maybe also `.cjs`), we would need a rename step as described in the linked comment.

Would you prefer that rename step to adding the nested `package.json`? There is a dedicated tool to do the latter: https://github.com/azu/tsconfig-to-dual-package

I think I’m tending towards using that tool because of the aforementioned footgun about not being allowed to reuse the same `.d.ts` file.

api/package.json
flying-sheep: got it! note that if we switch to `.mjs`/`.cjs`, it’s tempting to use the same declaration for both, but [that’s wrong](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-7.html#packagejson-exports-imports-and-self-referencing):

> It’s important to note that the CommonJS entrypoint and the ES module entrypoint each needs its own declaration file, even if the contents are the same between them. Every declaration file is interpreted either as a CommonJS module or as an ES module, based on its file extension and the `"type"` field of the `package.json`, and this detected module kind must match the module kind that Node will detect for the corresponding JavaScript file for type checking to be correct. Attempting to use a single `.d.ts` file to type both an ES module entrypoint and a CommonJS entrypoint will cause TypeScript to think only one of those entrypoints exists, causing compiler errors for users of the package.

.vscode/settings.json
flying-sheep: will do!

flying-sheep:
OK done. As said, `"module": "nodenext" | "node16"` does *not* make typescript emit `.mjs` or `.cjs` files, so I just rename the files after compiling.

I chose `mve` despite it being unmaintained, as it is very simple, works flawlessly, and it has much fewer dependencies than `renamer` or `move-file-cli`.

api/tsconfig.esm.json
StellaHuang95: I'm not an expert on this, so thanks for pointing this out🙂 My understanding is there are two concerns here.

**At build time**: `tsc` doesn't enforce extensions on relative imports under `module: "esnext"`, so it compiles fine but might break at runtime in Node's strict ESM resolver. 

Switching `tsconfig.esm.json` to `module: "nodenext"` would [catch this](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-7.html#ecmascript-module-support-in-nodejs:~:text=Relative%20import%20paths%20need%20full%20extensions) at compile time, but it also needs `"type": "module"` added to `api/package.json` so that `tsc` detects `src/main.ts` as an ESM source ([otherwise nodenext would emit CJS into the ESM build](https://www.typescriptlang.org/docs/handbook/modules/reference.html#module-format-detection)).

**At consume time**: the `.d.ts` file paired with the ESM entrypoint is classified as CJS (since the nearest ancestor `package.json` has no `"type":"module"`), mismatching the `.mjs` file which Node treats as ESM at runtime. The fix would be either:
 
  - Rename `.d.ts → .d.mts/.d.cts` (same mve step you already use for .js), or
  - Use `tsconfig-to-dual-package` to add sentinel `package.json` files: 
1. `out/esm/main.d.ts` → walk up → `out/esm/package.json ("type":"module")`
2. `out/cjs/main.d.ts` → walk up → `out/cjs/package.json ("type":"commonjs")` 
 
Is this what you're thinking too? No preference on which one to pick.

flying-sheep: Yeah, so `"module": "nodenext" | "node16"` [flexibly emit ESM or CJS code](https://www.typescriptlang.org/tsconfig/#node16node18node20nodenext) based on `"type": "module" | "commonjs"` in `package.json`, and it changes the module resolution TypeScript uses when type checking / building.

I think this context-dependent mode would be confusing, and instead opted for the explict `"module"` values in the `tsconfig.<variant>.json` files, that result in the following code being emitted:

- the ESM variant has `import { extensions, } from 'vscode';`
- the CJS variant has `const vscode_1 = require("vscode");` and then uses `vscode_1.extensions`.

This looks ideal, so we have build time settled. Regarding consume time, there are two technologies that must be satisfied:

1. node.js is satisfied if we either use `.mjs`/`.cjs` or if we use `.js` and the closest `package.json` has the appropriate `"type"`
2. TypeScript is satisfied if the resolved file has an associated `.d.ts`

I forgot to address 1. properly at first, but opted for renaming file suffixes, so we don’t have to use extra `package.json`s. Using the current combination of file endings and `"imports"` map, both technologies should be satisfied explicitly and cleanly:

- at type check time, users need to configure TypeScript to resolve the node.js package (including its import map), where both `import * as api from "@vscode/python-environments"` or `require("@vscode/python-environments")` will lead it to either `out/esm/main.d.ts` or `out/cjs/main.d.ts` based on the users’ settings.
- at build time, the same applies
- at node.js runtime (or when using a bundler), the same import map will be used to resolve to either `out/esm/main.mjs` or `out/cjs/main.cjs`, which node.js will be interpret as the correct module type based on file extension

StellaHuang95 (approved):

StellaHuang95 (approved):

StellaHuang95:
@eleanorjboyd looks like 2 approvals are required now so added you here:)

edvilme (approved):

Issue #1490, opened 2026-04-28
Deactivation may be triggered in restored terminal session where "deactivate" is not defined


Stasshe:
## Environment data

-   Python Environments extension version: latest (reproduced on current main)
-   Python extension (`ms-python.python`) version: latest
-   VS Code version (Help → About): VS Code Insiders (recent build)
-   OS and version: Windows 11 + WSL2 (Ubuntu)
-   Python version: 3.12 (system)
-   Environment manager in use: venv
-   Shell: zsh (likely shell-agnostic; depends on function-based activation)
-   Remote / container scenario: WSL
-   Workspace type: single folder
-   Is this a regression? Unknown

## Screenshot

<img width="682" height="168" alt="Image" src="https://github.com/user-attachments/assets/5ce55340-325b-4d9b-8c78-616575a0cb7d" />

## Repro Steps

1. Open a WSL workspace in VS Code Insiders.
2. Ensure persistent terminal sessions are enabled.
3. Create and activate a virtual environment:

```zsh
rm -rf .venv
python3 -m venv --without-pip .venv
source .venv/bin/activate
```

4. Confirm that the environment is active and that `deactivate` is defined:

```zsh
echo $VIRTUAL_ENV
typeset -f deactivate | head
```

5. Remove only the shell function while keeping the environment variables:

> Note: This step intentionally creates an inconsistent shell state for reproducibility.

```zsh
unset -f deactivate
echo $VIRTUAL_ENV
typeset -f deactivate | head
```

At this point:
- `VIRTUAL_ENV` is still set
- `deactivate` function no longer exists

6. Close VS Code without killing the terminal.
7. Reopen VS Code (e.g., from the app icon so that it restores the previous session).

## Expected behavior

It should avoid sending `deactivate` when the shell no longer has it defined.

It should either:
- verify that the command/function exists before calling it, or
- skip deactivation when terminal state is uncertain after restore.

## Actual behavior

The restored terminal may execute a deactivation command and produce:

```zsh
zsh: command not found: deactivate
```

## Logs

(No additional logs captured yet — behavior is observable directly in terminal output.)

## Additional context

- This issue does not depend on workspace files or `.vscode/settings.json`.
- It can be reproduced in a minimal temporary workspace.
- It appears related to terminal session restore, not environment creation itself.
- The issue does not reproduce when opening the same workspace explicitly via CLI (`code-insiders .`, or `code .`).
[...]

PR #1529, opened 2026-05-19
Fix `deactivate: command not found` when deactivating restored or half-activated terminals

StellaHuang95:
### TL;DR

In some scenarios a terminal ends up with `VIRTUAL_ENV` still set but the `deactivate` shell function undefined. When the extension then issues a deactivate against that terminal, the shell prints `deactivate: command not found`. This PR makes the extension's deactivation command **safe to send even when the function isn't there**: it now checks first, and only calls `deactivate` if it actually exists.

This is a small, contained fix that removes the user-visible error in #1490. It does not redesign the underlying activation tracking — that's a separate, larger piece of work outlined at the bottom of this description.

---

### Background — why this bug happens

The immediate cause of the error is simple: **the `deactivate` shell function is not defined in the terminal when the extension tries to call it.** The shell then prints `command not found`.

How does that happen? When the extension activates a Python venv, two different things get set up inside the shell:

1. An **environment variable** called `VIRTUAL_ENV` is set, and the venv's `bin`/`Scripts` directory is prepended to `PATH`. These are **exported**, so they're carried into child processes.
2. A **shell function** called `deactivate` is defined. It knows how to undo step 1 — restore the old `PATH`, unset `VIRTUAL_ENV`, fix the prompt, and remove itself. **Shell functions are not exported** and don't carry across process boundaries the way env vars do.

That asymmetry means there are several real-world scenarios in which `VIRTUAL_ENV` is still set but `deactivate` has gone missing — the shell looks "activated" but can't undo it. Known triggers include:

- The user (or a script) ran `unset -f deactivate`, which is the easiest way to reproduce the bug deliberately.
- The user ran a subshell (`bash` inside `zsh`, etc.). The subshell inherits exported env vars but not the parent's function definitions.
- A VS Code terminal restore path that actually re-spawns the shell process (for example, certain remote / dev-container reconnection scenarios). VS Code's normal persistent-terminal feature keeps the shell process alive across window reloads, in which case the function *is* preserved; the buggy paths are the ones where a fresh shell process is started but `VIRTUAL_ENV` is reintroduced via the new shell's environment.

In every one of those scenarios, the resulting state is the same: `VIRTUAL_ENV` set, `PATH` modified, prompt still shows `(.venv)`, but `deactivate` is gone. The wrap added by this PR fixes the user-visible error in **all** of them — it doesn't need to know which trigger fired, only whether `deactivate` is callable right now.

---

### What this PR changes

This PR fixes the symptom at the smallest possible code surface: the **one function** the extension uses to generate the deactivation command before sending it to the terminal.

**Files changed:**

- `src/features/terminal/shells/common/shellUtils.ts` — adds `wrapDeactivationCommand(shell, command)`.
- `src/features/common/activation.ts` — calls the wrap from `getDeactivationCommand`.
- `src/test/features/terminal/shells/common/shellUtils.unit.test.ts` — 21 new unit tests covering every supported shell and command shape.

**What `wrapDeactivationCommand` does:**

For shells where the deactivation command is the bare token `deactivate` (bash, zsh, fish, pwsh, gitbash), the wrap rewrites the command into a **guarded call**: check whether `deactivate` exists, and only run it if it does. For example, for bash/zsh/gitbash the command becomes:

```bash
 command -v deactivate >/dev/null 2>&1 && deactivate
```

(The leading space is preserved so the command stays out of shell history, just like before.)

For shells where the deactivation command is something else — `conda deactivate`, `pyenv shell --unset`, `deactivate.bat` on cmd, fish's `overlay hide`, or anything we don't recognize — the wrap returns the command **unchanged**. We only protect bare-token cases where we're confident a missing function is the problem.

**Where the wrap hooks in:**

In exactly one place: `getDeactivationCommand` in `src/features/common/activation.ts`. That function is the single chokepoint called by both deactivation paths in `terminalActivationState.ts` (`deactivateLegacy` and `deactivateUsingShellIntegration`), so every extension-initiated deactivation now flows through the wrap. No other code path is touched. Activation paths are completely unaffected.

**Result:**

- On `main`, deactivating a half-activated terminal produces a visible `bash: deactivate: command not found` error.
- With this PR, the same scenario silently no-ops — no error message, no state change.

---

### What this PR does NOT fix

This is important to call out this is a "surgical" fix so reviewers and users have the right expectations.

The PR removes the visible error but **does not clean up the half-activated state**. After deactivating a terminal that was in the broken state:

- `$VIRTUAL_ENV` is still set.
- `$PATH` still has the venv directory prepended.
- The prompt still shows `(.venv)`.
- The extension's internal tracker considers the terminal "deactivated," so subsequent deactivate commands are no-ops.

The terminal is in a recoverable but mildly confusing state. The user can recover by closing the terminal and opening a new one (which auto-activates cleanly), or by manually unsetting `VIRTUAL_ENV` and resetting `PS1`.

The reason we don't address this in the same PR is that fully cleaning up the half-state requires changes that are **fundamentally bigger and riskier** — the entire section below explains why.

---

### Why a structural fix is needed (and why it's a separate piece of work)

The wrap removes the visible error, but it doesn't prevent the underlying half-activated state from arising. To prevent it, the rc snippet that the extension installs in `shellStartup` mode needs to be smarter than it is today.

Today's snippet uses an exported environment variable, `VSCODE_PYTHON_AUTOACTIVATE_GUARD`, as a boolean "I've already activated, skip" flag. The design has two practical problems:

- **The guard is exported, so it propagates to places the `deactivate` function does not.** A child shell or a re-spawned shell can inherit the guard, decide it's "already activated," and skip activation — including the part that defines `deactivate`. That's one way the half-state can arise on its own, without the user doing anything.
- **The guard tracks a synthetic boolean rather than the real state.** Even if `deactivate` is missing for an unrelated reason (a plugin unset it, the user ran `unset -f`, an `exec` replaced the process), the guard still says "activated" and the snippet declines to re-define the function.

So while the guard isn't the *only* way `deactivate` can go missing, it's a design that lets the half-state stick around even when the shell has a fresh chance to fix itself.

#### Why this is a separate, larger PR

There are three real costs that make this not a casual one-line change:

**1. It changes behavior in interactive subshells.**
Today, opening `bash` inside an activated `zsh` silently inherits the guard and skips activation. After the structural fix, the subshell would re-run `activate` because it doesn't see `VIRTUAL_ENV` and `deactivate` together. That means the venv's `bin` directory gets prepended to `PATH` a second time, leading to visible PATH duplication, and `deactivate` in the subshell only partially undoes things. Manageable but real.

**2. Existing users wouldn't get the fix without a migration.**
The code that installs the rc snippet — `editUtils.ts hasStartupCode` — only checks for region markers and an environment-key string. It does not compare the script version. So if a user already has the old snippet in their `.bashrc`/`.zshrc`/etc., the extension thinks setup is done and never rewrites it. The version comment in the snippet today is for humans, not code.

Shipping a new snippet without a migration path means **only new users get the fix**. Everyone who already enabled `shellStartup` mode would still have the broken guard in their rc file. A full structural fix needs:

- A parseable version line inside the managed region.
- A `getStartupCodeVersion` helper that reads it.
- A `setupStartup` path that detects a version mismatch and rewrites the block.
- Atomic write-with-rename to avoid corrupting rc files mid-migration.
- Careful preservation of everything outside the managed markers.

Editing user rc files automatically is a sensitive operation. It needs its own PR with its own review and its own tests.

That's a multi-PR effort — appropriate for the root-cause fix, inappropriate to bolt onto a small bug-fix PR.

---

### DIFF
--- a/src/features/common/activation.ts
+++ b/src/features/common/activation.ts
@@ -4,6 +4,7 @@ import {
     getShellActivationCommand,
     getShellCommandAsString,
     getShellDeactivationCommand,
+    wrapDeactivationCommand,
 } from '../terminal/shells/common/shellUtils';
 import { identifyTerminalShell } from './shellDetector';
 
@@ -28,7 +29,7 @@ export function getDeactivationCommand(terminal: Terminal, environment: PythonEn
     const shell = identifyTerminalShell(terminal);
     const command = getShellDeactivationCommand(shell, environment);
     if (command) {
-        return getShellCommandAsString(shell, command);
+        return wrapDeactivationCommand(shell, getShellCommandAsString(shell, command));
     }
     return undefined;
 }
--- a/src/features/terminal/shells/common/shellUtils.ts
+++ b/src/features/terminal/shells/common/shellUtils.ts
@@ -50,6 +50,68 @@ export function getShellCommandAsString(shell: string, command: PythonCommandRun
     return commandStr;
 }
 
+// Shells whose bare `deactivate` command is a shell function/alias defined by venv's
+// `activate` script. These can disappear (e.g. on session restore where VIRTUAL_ENV
+// persists but the shell function does not), so we must guard the call with an
+// existence check before sending it to the terminal — otherwise the shell prints
+// `deactivate: command not found`.
+const bareDeactivateGuardByShell = new Map<string, (cmd: string) => string>([
+    // POSIX-family shells: `command -v <name>` is the portable existence check.
+    [ShellConstants.BASH, (cmd) => `command -v deactivate >/dev/null 2>&1 && ${cmd.trimStart()}`],
+    [ShellConstants.SH, (cmd) => `command -v deactivate >/dev/null 2>&1 && ${cmd.trimStart()}`],
+    [ShellConstants.ZSH, (cmd) => `command -v deactivate >/dev/null 2>&1 && ${cmd.trimStart()}`],
+    [ShellConstants.KSH, (cmd) => `command -v deactivate >/dev/null 2>&1 && ${cmd.trimStart()}`],
+    [ShellConstants.GITBASH, (cmd) => `command -v deactivate >/dev/null 2>&1 && ${cmd.trimStart()}`],
+    // fish uses `functions -q` for function existence.
+    [ShellConstants.FISH, (cmd) => `functions -q deactivate; and ${cmd.trimStart()}`],
+    // PowerShell: Get-Command returns silently if not found with -ErrorAction SilentlyContinue.
+    [ShellConstants.PWSH, (cmd) => `if (Get-Command deactivate -ErrorAction SilentlyContinue) { ${cmd.trimStart()} }`],
+]);
+
+/**
+ * Returns the bare `deactivate` token if and only if `command` represents a single,
+ * bare invocation of a shell function/alias literally named `deactivate` — meaning
+ * it is safe and meaningful to gate on the existence of that function in the shell.
+ *
+ * Returns `undefined` for anything else (full paths like `path/to/deactivate.bat`,
+ * multi-token forms like `conda deactivate`, `pyenv shell --unset`, `overlay hide ...`,
+ * etc.), since those have different failure modes that should not be silently swallowed.
+ *
+ * The token is normalized to lowercase so the generated guard is consistent across
+ * shells (notably PowerShell, which is case-insensitive).
+ */
+function bareDeactivateInvocation(command: string): string | undefined {
+    const trimmed = command.trim();
+    return trimmed.toLowerCase() === 'deactivate' ? 'deactivate' : undefined;
+}
+
+/**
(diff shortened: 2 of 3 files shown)


StellaHuang95:
Looping the terminal expert @anthonykim1  in for an opinion.

For https://github.com/microsoft/vscode-python-environments/issues/1490 I went with a surgical workaround: the extension's deactivation command is now safe to send even when the deactivate function isn't defined. It checks first and only calls deactivate if it actually exists. That removes the visible "command not found" error.

While digging into it, though, it's clear the activation system is pretty involved, especially around how state persists across terminal sessions and child processes. I wrote down my understanding of what a long-term fix would look like at the bottom of the PR description, but I think we should probably discuss whether we want to invest in it since it's complicated and require careful rewrite.

For now, feel free to review the current fix if you think it's good to ship. Thanks!

eleanorjboyd (approved):

PR #1530, opened 2026-05-20
Distinguish between direct and transitive packages

### LINKED ISSUE #524: Unclear which packages were installed as dependencies of another
This came up in a user study that if we have the ability to delete packages but a user goes to delete one that was installed as a dependency of another it could break the program. Could we provide some indicator or even a warning if they try to delete it that it will cause issues with the dependent package? 

Something like "Deleting X package will cause Y package to not perform properly due to required dependencies. Are you sure you want to delete?"

edvilme:
This pull request attempts to identify transitive packages in the user environment, and show indicators in the UI. 

## Problem
image

All installed packages, regardless of hierarchy are displayed equally in the sidebar. However, the relationships between them is not entirely obvious through the UI. This may cause (less experienced) users to get confused when they see packages they haven't explicitly installed, or to modify/delete transitive packages, potentially affecting their direct packages. 

Hence, there needs to be a way to clearly distinguish between them, and provide guardrails to prevent unintended behaviour. 

## Proposal
Direct packages are detected through the built in commands of the package managers
| Package Manager | Command | Description |
| -------------------  | ----------- | ------------- |
| pip                        |  `pip list --not-required --format=json` | Lists packages not required (direct, non-transitive) by other packages in a JSON format. |
| uv                         | `uv pip tree --depth 0` | Lists a dependency tree of all the packages with depth 0 (i.e., doesn't include transitive packages). |
| poetry                   | `poetry show --top-level --no-ansi` | Lists the top level packages |
| conda                   | `conda env export --from-history` * | Lists explicitly installed packages (note: this does not differenciate between direct/transitive) |

Packages shown are clearly identified as "Direct packages" or "Transitive" in the UI. Controls for uninstalling transitive packages are hidden to avoid unwanted behaviors. 

image

Closes #524

### DIFF
--- a/api/src/main.ts
+++ b/api/src/main.ts
@@ -694,6 +694,13 @@ export interface PackageManager {
 
     /**
      * Fetches the names of direct (non-transitive) packages for the specified Python environment.
+     *
+     * **Caveat:** Most package managers cannot track user install intent. For pip, this uses
+     * `pip list --not-required` which returns packages with no installed dependents (leaf packages),
+     * not necessarily packages the user explicitly installed. For example, if a user runs
+     * `pip install flask werkzeug`, werkzeug will still be reported as transitive because flask
+     * depends on it. This is a best-effort approximation.
+     *
      * @param environment - The Python environment for which to fetch direct package names.
      * @returns A promise that resolves to a set of package name strings, or undefined if not supported.
      */
--- a/api/src/main.ts
+++ b/api/src/main.ts
@@ -688,6 +688,13 @@ export interface PackageManager {
     onDidChangePackages?: Event<DidChangePackagesEventArgs>;
 
     /**
+     * Fetches the names of direct (non-transitive) packages for the specified Python environment.
+     * @param environment - The Python environment for which to fetch direct package names.
+     * @returns A promise that resolves to an array of package name strings, or undefined if not supported.
+     */
+    fetchDirectPackageNames?(environment: PythonEnvironment): Promise<Set<string> | undefined>;
+
+    /**
      * Clears the package manager's cache.
      * @returns A promise that resolves when the cache is cleared.
      */
--- a/api/src/main.ts
+++ b/api/src/main.ts
@@ -675,9 +675,9 @@ export interface PackageManager {
     /**
      * Refreshes the package list for the specified Python environment.
      * @param environment - The Python environment for which to refresh the package list.
-     * @returns A promise that resolves when the refresh is complete.
+     * @returns A promise that resolves with the refreshed list of packages, or undefined.
      */
-    refresh(environment: PythonEnvironment): Promise<void>;
+    refresh(environment: PythonEnvironment): Promise<Package[] | undefined>;
 
     /**
      * Retrieves the list of packages for the specified Python environment.
@@ -1041,9 +1041,9 @@ export interface PythonPackageGetterApi {
      * Refresh the list of packages in a Python Environment.
      *
      * @param environment The Python Environment for which the list of packages is to be refreshed.
-     * @returns A promise that resolves when the list of packages has been refreshed.
+     * @returns A promise that resolves with the refreshed list of packages, or undefined.
      */
-    refreshPackages(environment: PythonEnvironment): Promise<void>;
+    refreshPackages(environment: PythonEnvironment): Promise<Package[] | undefined>;
 
     /**
      * Get the list of packages in a Python Environment.
--- a/api/src/main.ts
+++ b/api/src/main.ts
@@ -578,6 +578,11 @@ export interface PackageInfo {
      * The URIs associated with the package.
      */
     readonly uris?: readonly Uri[];
+
+    /**
+     * Whether the package is a transitive dependency.
+     */
+    readonly isTransitive?: boolean;
 }
 
 /**
@@ -690,9 +695,9 @@ export interface PackageManager {
     /**
      * Fetches the names of direct (non-transitive) packages for the specified Python environment.
      * @param environment - The Python environment for which to fetch direct package names.
-     * @returns A promise that resolves to an array of package name strings, or undefined if not supported.
+     * @returns A promise that resolves to a set of package name strings, or undefined if not supported.
      */
-    fetchDirectPackageNames?(environment: PythonEnvironment): Promise<Set<string> | undefined>;
+    getDirectPackageNames?(environment: PythonEnvironment): Promise<Set<string> | undefined>;
 
     /**
      * Clears the package manager's cache.
--- a/src/api.ts
+++ b/src/api.ts
@@ -576,7 +576,7 @@ export interface PackageInfo {
     /**
      * Whether the package is a transitive dependency.
      */
-    isTransitive?: boolean;
+    readonly isTransitive?: boolean;
 }
 
 /**
@@ -689,7 +689,7 @@ export interface PackageManager {
     /**
      * Fetches the names of direct (non-transitive) packages for the specified Python environment.
      * @param environment - The Python environment for which to fetch direct package names.
-     * @returns A promise that resolves to an array of package name strings, or undefined if not supported.
+     * @returns A promise that resolves to a set of package name strings, or undefined if not supported.
      */
     getDirectPackageNames?(environment: PythonEnvironment): Promise<Set<string> | undefined>;
 
--- a/src/api.ts
+++ b/src/api.ts
@@ -691,7 +691,7 @@ export interface PackageManager {
      * @param environment - The Python environment for which to fetch direct package names.
      * @returns A promise that resolves to an array of package name strings, or undefined if not supported.
      */
-    fetchDirectPackageNames?(environment: PythonEnvironment): Promise<Set<string> | undefined>;
+    getDirectPackageNames?(environment: PythonEnvironment): Promise<Set<string> | undefined>;
 
     /**
      * Clears the package manager's cache.
--- a/src/api.ts
+++ b/src/api.ts
@@ -572,6 +572,11 @@ export interface PackageInfo {
      * The URIs associated with the package.
      */
     readonly uris?: readonly Uri[];
+
+    /**
+     * Whether the package is a transitive dependency.
+     */
+    isTransitive?: boolean;
 }
 
 /**
@@ -680,6 +685,13 @@ export interface PackageManager {
      * Event that is fired when packages change.
      */
     onDidChangePackages?: Event<DidChangePackagesEventArgs>;
+
+    /**
+     * Fetches the names of direct (non-transitive) packages for the specified Python environment.
+     * @param environment - The Python environment for which to fetch direct package names.
+     * @returns A promise that resolves to an array of package name strings, or undefined if not supported.
+     */
+    fetchDirectPackageNames?(environment: PythonEnvironment): Promise<Set<string> | undefined>;
 
     /**
      * Clears the package manager's cache.
--- a/src/api.ts
+++ b/src/api.ts
@@ -669,9 +669,9 @@ export interface PackageManager {
     /**
      * Refreshes the package list for the specified Python environment.
      * @param environment - The Python environment for which to refresh the package list.
-     * @returns A promise that resolves when the refresh is complete.
+     * @returns A promise that resolves with the refreshed list of packages, or undefined.
      */
-    refresh(environment: PythonEnvironment): Promise<void>;
+    refresh(environment: PythonEnvironment): Promise<Package[] | undefined>;
 
     /**
      * Retrieves the list of packages for the specified Python environment.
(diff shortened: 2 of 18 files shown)


eleanorjboyd:
do you have a screenshot of what this would look like? and how do you see it working for other pkg managers?

edvilme:
Depends on #1538
Will re-target and rebase when merged

src/features/envCommands.ts
```diff
@@ -305,6 +305,14 @@ export async function removeEnvironmentCommand(context: unknown, managers: Envir
 
 export async function handlePackageUninstall(context: unknown, em: EnvironmentManagers) {
     if (context instanceof PackageTreeItem || context instanceof ProjectPackage) {
+        // Ask for user confirmation if the package is transitive
+        if (context.pkg.isTransitive) {
+            const message = `The package "${context.pkg.name}" is a transitive dependency. Uninstalling it may break other packages that depend on it. Are you sure you want to uninstall it?`;
```
eleanorjboyd: do you need this msg if the delete option is gone?

src/managers/builtin/pipListUtils.ts
```diff
@@ -4,6 +4,13 @@ export interface PipPackage {
     displayName: string;
     description: string;
 }
+export function parseUvTree(data: string): string[] {
+    return data
+        .split('\n')
+        .map((line) => line.trim())
+        .map((line) => line.split(/\s+/, 1)[0])
+        .filter((name) => !!name);
+}
```
eleanorjboyd, line 15: can you confirm there isn't a way to get the tree as a json or smth easier to parse?

src/managers/poetry/poetryPackageManager.ts
```diff
@@ -263,6 +263,21 @@ export class PoetryPackageManager implements PackageManager, Disposable {
         // Convert to Package objects using the API
         return poetryPackages.map((pkg) => this.api.createPackageItem(pkg, environment, this));
     }
+
+    async getDirectPackageNames(_environment: PythonEnvironment): Promise<Set<string> | undefined> {
```
eleanorjboyd, line 269: same thing here- any args to make it more easily parsable

src/api.ts
```diff
@@ -681,6 +686,13 @@ export interface PackageManager {
      */
     onDidChangePackages?: Event<DidChangePackagesEventArgs>;
 
+    /**
+     * Fetches the names of direct (non-transitive) packages for the specified Python environment.
+     * @param environment - The Python environment for which to fetch direct package names.
+     * @returns A promise that resolves to an array of package name strings, or undefined if not supported.
+     */
+    getDirectPackageNames?(environment: PythonEnvironment): Promise<Set<string> | undefined>;
+
```
eleanorjboyd, line 702: feel like this could also be an option addition, so like get pkgs with an optional arg for direct only

src/test/managers/common/packageChanges.unit.test.ts
```diff
@@ -169,5 +169,101 @@ suite('packageChanges', () => {
...
+                { name: 'charset-normalizer', version: '3.0.0' } as Package,
+            ];
+            getPackagesStub.resolves(after);
+            const getDirectPackageNamesStub = sinon.stub().resolves(new Set(['requests']));
+            (packageManager as unknown as Record<string, unknown>).getDirectPackageNames = getDirectPackageNamesStub;
+            const onChanges = sinon.stub();
+
+            await updatePackagesAndNotify(packageManager, environment, undefined, onChanges);
+
+            assert.strictEqual(after[0].isTransitive, true, 'urllib3 should be transitive');
```
eleanorjboyd: would add a test here which confirms error handling works if the output is returned in the wrong format or something

edvilme:
Addressed all review comments in the latest commit. See individual replies on each comment thread for details.

src/features/envCommands.ts
```diff
@@ -305,6 +305,21 @@ export async function removeEnvironmentCommand(context: unknown, managers: Envir
...
+                    context.pkg.name,
+                ),
+                { modal: true },
+                l10n.t('Uninstall'),
+                l10n.t('Cancel'),
+            );
+            if (confirm !== l10n.t('Uninstall')) {
+                return;
+            }
+        }
```
edvilme, line 321:
```suggestion
        if (context.pkg.isTransitive) {
            return;
        }
```

src/managers/builtin/pipListUtils.ts
edvilme: No, didn't find one for UV

src/features/envCommands.ts
```diff
@@ -305,6 +305,9 @@ export async function removeEnvironmentCommand(context: unknown, managers: Envir
 
 export async function handlePackageUninstall(context: unknown, em: EnvironmentManagers) {
     if (context instanceof PackageTreeItem || context instanceof ProjectPackage) {
+        if (context.pkg.isTransitive) {
```
StellaHuang95, line 308: Does this mean silent no-op uninstall on transitive packages from command palette? Can the ux flow be improved here?

api/src/main.ts
```diff
@@ -687,6 +692,13 @@ export interface PackageManager {
      */
     onDidChangePackages?: Event<DidChangePackagesEventArgs>;
 
+    /**
+     * Fetches the names of direct (non-transitive) packages for the specified Python environment.
+     * @param environment - The Python environment for which to fetch direct package names.
+     * @returns A promise that resolves to a set of package name strings, or undefined if not supported.
+     */
+    getDirectPackageNames?(environment: PythonEnvironment): Promise<Set<string> | undefined>;
```
StellaHuang95, line 707: Is it implemented for conda? Seems not covered

src/managers/builtin/utils.ts
```diff
@@ -235,14 +235,34 @@ export async function refreshPipPackages(
...
+            ['pip', 'tree', '--python', environment.execInfo.run.executable, '--depth=0'],
+            undefined,
+            log,
+            undefined,
+            PIP_LIST_TIMEOUT_MS,
+        );
+        return parseUvTree(treeOutput);
+    }
+    const data = await execPipList(environment, log, ['--not-required']);
+    const packages = parsePipListJson(data);
```
StellaHuang95, line 270: Copilot generated: `pip list --not-required` returns packages with no installed dependents (leaves of the graph), not packages a user directly installed. Example: `pip install flask werkzeug` — `werkzeug` is required by `flask`, so it won't appear in `--not-required` and will be shown as `(transitive)` even though the user installed it explicitly. pip doesn't track install intent, so this is the closest proxy available but isn't equivalent. Worth documenting the caveat on `getDirectPackageNames` and possibly surfacing it in the transitive tooltip so users aren't misled.

src/managers/common/packageChanges.ts
```diff
@@ -48,6 +48,21 @@ export async function updatePackagesAndNotify(
...
+
+    // Handle transitive dependencies (best-effort, don't break package refresh on failure)
+    let afterDirectDependenciesNames: Set<string> | undefined;
+    try {
+        afterDirectDependenciesNames = await packageManager.getDirectPackageNames?.(environment);
+    } catch {
+        // If direct package detection fails, leave isTransitive undefined rather than breaking refresh
+    }
+    if (afterDirectDependenciesNames && afterDirectDependenciesNames.size > 0) {
+        for (const pkg of after) {
```
StellaHuang95: Copilot generated: This comparison is case- and separator-sensitive, but `pip list` / `uv pip tree` / `poetry show` and `Package.name` can disagree (`PyYAML` vs `pyyaml`, `typing_extensions` vs `typing-extensions`, `ruamel.yaml` vs `ruamel-yaml`). Consider normalizing both sides per PEP 503 before comparing, e.g. `name.toLowerCase().replace(/[-_.]+/g, '-')`. Without this, real direct packages can be silently misclassified as transitive.

src/managers/poetry/poetryPackageManager.ts
```diff
@@ -263,6 +263,21 @@ export class PoetryPackageManager implements PackageManager, Disposable {
...
+        try {
+            const topLevelResult = await runPoetry(['show', '--no-ansi', '--top-level'], undefined, this.log);
+            const names = topLevelResult
+                .split('\n')
+                .map((line) => line.trim())
+                .map((line) => line.match(/^([a-zA-Z0-9_-]+)/)?.[1] ?? '')
```
StellaHuang95: Copilot generated: The character class `[a-zA-Z0-9_-]` doesn't include `.`, so dotted package names like `zope.interface`, `ruamel.yaml`, and `backports.zoneinfo` get truncated to `zope` / `ruamel` / `backports` and won't match the names returned by `getPackages()`. Add `\.` to the class (and ideally normalize per PEP 503 before comparing).

src/managers/common/packageChanges.ts
StellaHuang95: Copilot generated: `isTransitive` is declared `readonly` on both the public `PackageInfo` and `PythonPackageImpl`, but this writes to it through a structural cast — that bypasses the contract just introduced. It also depends on `getPackages()` returning the same cached object references each call, which isn't a documented guarantee. Preferred: compute `isTransitive` inside each manager's `refresh()` and pass it through `api.createPackageItem(...)` so the field is genuinely immutable and the cast can go away.

src/features/views/envManagersView.ts
```diff
@@ -252,9 +252,17 @@ export class EnvManagerView implements TreeDataProvider<EnvTreeItem>, Disposable
             const views: EnvTreeItem[] = [];
 
             if (pkgManager) {
-                const packages = await pkgManager.getPackages(environment);
+                let packages = await pkgManager.getPackages(environment);
+                if (!packages || packages.length === 0) {
```
StellaHuang95: Copilot generated: For environments that legitimately have zero packages, this will trigger a full `refresh()` on every expansion (and `refresh()` now costs two subprocesses with the transitive detection added). `refresh()` also fires `onDidChangePackages`, which is wired into tree refresh — worth double-checking this can't loop. Consider a sentinel: `undefined` cache = never refreshed, `[]` = refreshed-and-empty. Also: the same block is duplicated in `projectView.ts` — would be cleaner as a shared helper.

src/managers/common/packageChanges.ts
```diff
@@ -48,6 +48,21 @@ export async function updatePackagesAndNotify(
     onChanges: PackageChangesCallback,
 ): Promise<void> {
     const after = (await packageManager.getPackages(environment, { skipCache: true })) ?? [];
+
+    // Handle transitive dependencies (best-effort, don't break package refresh on failure)
+    let afterDirectDependenciesNames: Set<string> | undefined;
+    try {
```
StellaHuang95: Copilot generated: `getPackages` and `getDirectPackageNames` are awaited sequentially, so every refresh now pays for two subprocess spawns in series. On Windows in particular, spawn cost dominates. Minimum fix: run them with `Promise.all`. Better: fold direct-name detection into each manager's `refresh()` so it can be combined into a single subprocess (e.g., one `uv pip list` + `uv pip tree` pass) and the field can be set immutably at construction time.

api/src/main.ts
```diff
@@ -687,6 +692,13 @@ export interface PackageManager {
      */
     onDidChangePackages?: Event<DidChangePackagesEventArgs>;
 
+    /**
+     * Fetches the names of direct (non-transitive) packages for the specified Python environment.
+     * @param environment - The Python environment for which to fetch direct package names.
+     * @returns A promise that resolves to a set of package name strings, or undefined if not supported.
```
StellaHuang95, line 705: Copilot generated: Two questions on this signature:

1. `Set<string>` is unusual on our public API surface; `readonly string[]` would be more consistent with the rest of `PackageManager` and easier for consumers to serialize/transport.
2. The three observable states — method not implemented, method returns `undefined`, method returns an empty `Set` — all need documented semantics. The current consumer treats empty-set as "no info" (via a `size > 0` guard), which silently drops legitimately-empty results. Likewise, consumers of `PackageInfo.isTransitive` can't distinguish `undefined` from `false`; please document that `undefined` means "unknown".

src/managers/common/packageChanges.ts
```diff
@@ -48,6 +48,21 @@ export async function updatePackagesAndNotify(
     onChanges: PackageChangesCallback,
 ): Promise<void> {
     const after = (await packageManager.getPackages(environment, { skipCache: true })) ?? [];
+
+    // Handle transitive dependencies (best-effort, don't break package refresh on failure)
```
StellaHuang95: Copilot generated: The extra subprocess on every refresh is noticeable on slow machines / large environments, and some users may not want the classification at all (see the `pip list --not-required` accuracy caveat). Consider gating this behind a setting like `python-envs.detectTransitiveDependencies` (default `true`) so users can opt out.

src/features/envCommands.ts
edvilme: Should we show a confirmation message: "This is a transitive package, uninstalling it may have unintended consequences"?

StellaHuang95: I think right now it's just a no-op so that sound confusing and not consistent with other paths?

api/src/main.ts
edvilme: Conda doesn't have a direct mechanism for getting direct packages, only `conda env export --from-history` which isn't strictly the same as direct or transitive packages, and can lead to more confusion

src/managers/common/packageChanges.ts
edvilme: I think the Package class should have a `get normalizedName` attribute that resolves to this, so we can do comparisons in a single gesture :)

edvilme: Added helper method to normalize both sides

eleanorjboyd (dismissed):

aeschli (approved):

DonJayamanne (approved):

PR #1548, opened 2026-05-29
feat: add PET version and build ID telemetry properties to enhance performance tracking

eleanorjboyd:
goes along with https://github.com/microsoft/python-environment-tools/pull/470 to support getting PET versions as telemetry to track performance regression and improvements across specific version numbers


StellaHuang95:
This looks good, I am going to add the commit hash as another property to the telemetry and combine all the changes into a single PR.

Issue #1581, opened 2026-06-15
Cross-session cache misses on every fresh remote (SSH / WSL / dev container / codespace)


eleanorjboyd:
## Problem

The cross-session cache for the discovered global system Python (PR #1455) lives in `context.globalState`. In remote scenarios, `context.globalState` is scoped to the remote machine, not the local one. As a result, the cache never primes on a fresh remote and every first connection to a new SSH host, dev container, codespace, attached container, or WSL distro starts cold and falls through to a full PET refresh on the foreground env-selection path.

This also affects the local case any time `globalState` is wiped (e.g. extension reinstall, profile reset).

## Current behavior

- [`getSystemEnvForGlobal()` / `setSystemEnvForGlobal()`](../../src/managers/builtin/cache.ts) read/write `SYSTEM_GLOBAL_KEY` via [`getGlobalPersistentState()`](../../src/common/persistentState.ts), which wraps `context.globalState`.
- `getGlobalPersistedPath` is wired into `tryFastPathGet` ([fastPath.ts](../../src/managers/common/fastPath.ts)) for the system-Python path. When `globalState` is empty the fast path returns `undefined` and registration falls through to the slow path.
- Workspace-scoped paths use `setSystemEnvForWorkspace` / `getSystemEnvForWorkspace`, which write to `context.workspaceState` (keyed on workspace folder fsPath) — that path is unaffected.

## Why it matters

Every time a user connects VS Code to a new remote (or reopens after the remote `globalState` is reset), the foreground env-selection path waits on a fresh PET refresh instead of validating a cached path. This is the dominant remaining hang scenario on Linux based on our most recent investigation, and ssh-remote, WSL, and dev-containers are all affected.

## Acceptance criteria

- [ ] On a workspace where the global system Python has been resolved at least once, opening the same workspace folder in a fresh VS Code session on the same remote hits the cache without a PET refresh.
- [ ] On a brand-new workspace folder on the same remote, behavior is unchanged from today (cache miss → PET refresh).
- [ ] On a brand-new remote with a brand-new workspace folder, behavior is unchanged from today (option 1 doesn't claim to fix this case — option 2/3 would).
- [ ] No regression to existing telemetry: `global_env.cache` continues to emit `hit` / `miss` / `stale` correctly.
- [ ] No regression to the workspace-scoped fast-path (`getSystemEnvForWorkspace` users).
[...]

PR #1598, opened 2026-06-19
Coalesce concurrent native finder refreshes with the same key

### LINKED ISSUE #1587: Environment discovery is taking a VERY long time
I opened the code at https://github.com/pytorch/pytorch.git (after cloning it) and got this output in the enviroments extension:

```
2026-06-16 10:53:57.678 [warning] [Pipenv] Environment discovery took 199.2s (found 0 environments). If this is causing problems, please report it: https://github.com/microsoft/vscode-python-environments/issues/new
2026-06-16 10:54:52.264 [info] Refreshing poetry environments
2026-06-16 10:54:52.272 [warning] [Poetry] Environment discovery took 253.8s (found 0 environments). If this is causing problems, please report it: https://github.com/microsoft/vscode-python-environments/issues/new
```

I have 4 or 5 pythons installed on this machine and a single venv in the pytorch folder.

StellaHuang95:
## Summary

Fixes #1587.

At startup, several environment managers (venv, system, pyenv, pipenv, poetry, conda) each call `nativeFinder.refresh(true, undefined)` concurrently — typically triggered when the Python extension calls `api.refreshEnvironments(undefined)` via its `triggerRefresh` on activation. Every one of those calls resolves to the same cache key (`'all'`) but is queued as a distinct task on the native finder's worker pool, which runs with `concurrency = 1`. Neither the worker pool nor `doRefresh` deduplicates identical submissions, so N parallel callers become N sequential full PET discoveries — same input, same output, N-1 wasted scans.

The reporter (pytorch workspace) saw the result of this stacking:

```
[Pipenv] Environment discovery took 199.2s (found 0 environments).
[Poetry] Environment discovery took 253.8s (found 0 environments).
```

plus the `Setup appears hung during stage: managerRegistration` watchdog firing, and the same `Discovered env: <path>` lines repeating 4-5× in the log — the smoking gun that PET ran the same scan multiple times back-to-back.

## Fix

Add an in-flight tracking map (`Map<string, Promise<NativeInfo[]>>`) in `NativePythonFinderImpl` and a guard at the top of `handleHardRefresh`: if a refresh for the same key is already running, return the existing promise instead of queueing another task. The slot is cleared in `.finally` so a rejected refresh does not poison future callers — sequential refreshes still each run a fresh PET scan.

`handleSoftRefresh` already falls through to `handleHardRefresh` on cache miss, so it picks up the dedup automatically — a concurrent soft + hard for the same key now share one PET scan.

## Why this layer

- All duplicate-refresh paths converge at `handleHardRefresh` (manager refreshes, conda's `getConda` fallback via `native.refresh(false)`, the Python extension's `triggerRefresh`, the `Refresh All Environment Managers` command, external API consumers). Fixing here covers every caller in one place.
- No public API change, no observable semantic change for sequential refreshes, no test breakage.
- Mirrors the existing `inFlight: Map<string, Promise<…>>` idiom in `src/features/inlineScriptLazyDetector.ts`.

## Expected impact for #1587

| Metric | Before | After |
|---|---:|---:|
| PET scans triggered at startup | ~6-7 | **1** |
| Pipenv discovery warning | 106-199 s | ~25-35 s |
| Poetry discovery warning | 121-253 s | ~25-35 s |
| `Setup appears hung during stage: managerRegistration` | fires | should not fire |
| Total startup-to-ready | ~150 s | **~40-50 s** (~3× wall-clock) |

The fix eliminates duplicate-scan stacking; it does not speed up a single PET scan (intrinsic ~25-35 s on pytorch on the reporter''s machine — separate from this change).

## Verification

- ✅ Webpack production build succeeds.
- ✅ ESLint clean on the changed file.
- ✅ Unit tests: 1215 pass (same 3 unrelated pre-existing failures on `main`).
- ✅ Built a VSIX with this branch and a VSIX from `main`, diffed the minified `handleHardRefresh` body — the fix VSIX has the new structure (`inFlightRefreshes.get → guard → addToQueue.then.finally`), `main` has the old (`await pool.addToQueue`).

Tested manually by inspecting the minified output of both builds; would appreciate eyes from someone with the pytorch repro to confirm wall-clock improvement on a real run.

## What this does NOT address (intentionally)

- Single PET scan latency on large workspaces (separate, PET-side).
- Cross-session cache misses on remote (issue #1581, separate fix in flight).
- The Python extension''s `triggerRefresh` on every activation (worth a follow-up — could skip when our cache is non-empty).
- The double-init in `InternalEnvironmentManager.refresh`''s tail-call to `getEnvironments(''all'')` (separate small follow-up; produces the "phantom" lowercase-only refresh log lines).

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/src/managers/common/nativePythonFinder.ts
+++ b/src/managers/common/nativePythonFinder.ts
@@ -326,6 +326,11 @@ class NativePythonFinderImpl implements NativePythonFinder {
     private connection: rpc.MessageConnection;
     private readonly pool: WorkerPool<NativePythonEnvironmentKind | Uri[] | undefined, NativeInfo[]>;
     private cache: Map<string, NativeInfo[]> = new Map();
+    /**
+     * Tracks in-flight hard refreshes by cache key so concurrent callers share a
+     * single PET scan instead of queueing duplicate work.
+     */
+    private inFlightRefreshes: Map<string, Promise<NativeInfo[]>> = new Map();
     private startDisposables: Disposable[] = [];
     private proc: ChildProcess | undefined;
     private processExited: boolean = false;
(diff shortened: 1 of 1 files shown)


StellaHuang95:
Rich verified he's no longer seeing something like this in the logs: `2026-06-18 15:53:31.276 [warning] [Pipenv] Environment discovery took 106.2s (found 0 environments). If this is causing problems, please report it: https://github.com/microsoft/vscode-python-environments/issues/new`

eleanorjboyd (approved):

PR #1621, opened 2026-07-05
Introduce package manager command classes

edvilme:
## Summary

Introduces the reusable command-object layer for package management while leaving existing package-manager call sites unchanged.

## Scope

- Adds the `PackageManagerCommand` base class and shared execution options.
- Adds abstract templates for install, uninstall, list, version, available versions, and direct package names.
- Adds concrete Pip/UV, Conda, and Poetry implementations.
- Adds command factories, exports, and command-local execution helpers required by those implementations.
- Preserves the existing executables, flags, environment targeting, prerelease defaults, and list-command timeouts.

## Non-goals

This PR does not add command tests or migrate package managers to the new classes. Those changes are isolated in follow-up PRs #1677 and #1678.

## PR stack

1. This PR: command classes and implementations, based on `main`.
2. #1677: command smoke tests, based on this branch.
3. #1678: adoption and cleanup, based on #1677.

## Testing

- `npm run lint`
- `npm run compile-tests`
- `npm run unittest` (1,448 passing, 4 pending)

### DIFF
--- a/src/managers/base/commands/availableVersions.ts
+++ b/src/managers/base/commands/availableVersions.ts
@@ -1,0 +1,31 @@
+import type { Pep440Version } from '@renovatebot/pep440';
+import { explain as parsePep440Version } from '@renovatebot/pep440';
+import { BaseExecuteArgs, PackageManagerCommand } from './packageManagerCommand';
+
+/**
+ * Arguments for available versions command execution (change per execution).
+ */
+export interface AvailableVersionsExecuteArgs extends BaseExecuteArgs {
+    packageName: string;
+    pythonVersion: string;
+    includePrerelease?: boolean;
+}
+
+/**
+ * Template class for availableVersions commands.
+ * Subclasses implement concrete package-manager-specific logic.
+ */
+export abstract class AvailableVersionsCommand extends PackageManagerCommand {
+    protected static readonly configSection = 'availableVersionsCommandArgs';
+    protected abstract buildCommand(executeArgs: AvailableVersionsExecuteArgs): string[];
+    protected parseVersions(versions: string[], includePrerelease?: boolean): Pep440Version[] {
+        let parsed = Array.from(new Set(versions))
+            .map((version) => parsePep440Version(version.trim()))
+            .filter((version): version is Pep440Version => version !== null);
+        if (includePrerelease === false) {
+            parsed = parsed.filter((version) => !/[ab]|rc|dev/i.test(version.public));
+        }
+        return parsed;
+    }
+    abstract execute(executeArgs: AvailableVersionsExecuteArgs): Promise<Pep440Version[]>;
+}
--- a/src/managers/base/commands/availableVersions.ts
+++ b/src/managers/base/commands/availableVersions.ts
@@ -23,7 +23,7 @@ export abstract class AvailableVersionsCommand extends PackageManagerCommand {
             .map((version) => parsePep440Version(version.trim()))
             .filter((version): version is Pep440Version => version !== null);
         if (includePrerelease === false) {
-            parsed = parsed.filter((version) => !/[ab]|rc|dev/i.test(version.public));
+            parsed = parsed.filter((version) => !version.is_prerelease);
         }
         return parsed;
     }
--- a/src/managers/base/commands/packageManagerCommand.ts
+++ b/src/managers/base/commands/packageManagerCommand.ts
@@ -1,0 +1,46 @@
+import { CancellationToken, LogOutputChannel, WorkspaceConfiguration } from 'vscode';
+import { getConfiguration } from '../../../common/workspace.apis';
+
+/**
+ * Base interface for all command execute arguments.
+ * Provides optional cancellation token that all commands can use.
+ */
+export interface BaseExecuteArgs {
+    cancellationToken?: CancellationToken;
+}
+
+/**
+ * Constructor options shared by all package manager commands.
+ */
+export interface CommandConstructorOptions {
+    pythonExecutable: string;
+    cwd?: string;
+    log?: LogOutputChannel;
+}
+
+/**
+ * Base class for all package manager commands.
+ * Provides common properties and minimal interface for subclasses.
+ */
+export abstract class PackageManagerCommand {
+    protected static readonly configSection?: string;
+
+    protected pythonExecutable: string;
+    protected cwd?: string;
+    protected log?: LogOutputChannel;
+    protected timeout: number = 300000;
+    protected config?: WorkspaceConfiguration;
+
+    constructor(options: CommandConstructorOptions) {
+        this.pythonExecutable = options.pythonExecutable;
+        this.cwd = options.cwd;
+        this.log = options.log;
+        const configSection = (this.constructor as typeof PackageManagerCommand).configSection;
+        this.config = configSection ? getConfiguration(`python-envs.packageManager.${configSection}`) : undefined;
+    }
+
+    /**
+     * Subclasses implement to build the command arguments.
+     */
+    protected abstract buildCommand(executeArgs: BaseExecuteArgs): string[];
+}
--- a/src/managers/builtin/commands/availableVersions.ts
+++ b/src/managers/builtin/commands/availableVersions.ts
@@ -1,0 +1,93 @@
+import type { Pep440Version } from '@renovatebot/pep440';
+import { AvailableVersionsCommand, type AvailableVersionsExecuteArgs } from '../../base/commands/index';
+import { runPython, runUV } from '../helpers';
+
+/**
+ * Pip available versions command.
+ * Parsed command: `python -m pip index versions <package> --json --python-version <version>`
+ * Official documentation: https://pip.pypa.io/en/stable/cli/pip_index/
+ */
+export class PipAvailableVersionsCommand extends AvailableVersionsCommand {
+    protected buildCommand(executeArgs: AvailableVersionsExecuteArgs): string[] {
+        return [
+            '-m',
+            'pip',
+            'index',
+            'versions',
+            executeArgs.packageName,
+            '--json',
+            '--python-version',
+            executeArgs.pythonVersion,
+        ];
+    }
+
+    async execute(executeArgs: AvailableVersionsExecuteArgs): Promise<Pep440Version[]> {
+        const output = await runPython(
+            this.pythonExecutable,
+            this.buildCommand(executeArgs),
+            undefined,
+            this.log,
+            executeArgs.cancellationToken,
+            this.timeout,
+        );
+        const match = output.match(/{[\s\S]*}/);
+        if (!match) {
+            return [];
+        }
+
+        try {
+            const parsed = JSON.parse(match[0]) as { versions?: string[] };
+            return this.parseVersions(
+                Array.isArray(parsed.versions) ? parsed.versions : [],
+                executeArgs.includePrerelease,
+            );
+        } catch {
+            return [];
+        }
+    }
+}
+
+/**
+ * UV available versions command.
+ * Parsed command: `uv tool run pip index versions <package> --json --python-version <version>`
+ * Official documentation: https://docs.astral.sh/uv/pip/
+ */
+export class UvAvailableVersionsCommand extends AvailableVersionsCommand {
+    protected buildCommand(executeArgs: AvailableVersionsExecuteArgs): string[] {
+        return [
+            'tool',
+            'run',
+            'pip',
+            'index',
+            'versions',
+            executeArgs.packageName,
+            '--json',
+            '--python-version',
+            executeArgs.pythonVersion,
+        ];
+    }
+
+    async execute(executeArgs: AvailableVersionsExecuteArgs): Promise<Pep440Version[]> {
+        const output = await runUV(
+            this.buildCommand(executeArgs),
+            undefined,
+            this.log,
+            executeArgs.cancellationToken,
+            this.timeout,
+        );
+        const match = output.match(/{[\s\S]*}/);
+        if (!match) {
+            return [];
+        }
+
+        try {
+            const parsed = JSON.parse(match[0]) as { versions?: string[] };
+            return this.parseVersions(
+                Array.isArray(parsed.versions) ? parsed.versions : [],
+                executeArgs.includePrerelease,
+            );
+        } catch {
+            return [];
+        }
+    }
+}
--- a/src/managers/builtin/commands/install.ts
+++ b/src/managers/builtin/commands/install.ts
@@ -1,0 +1,52 @@
+import { InstallCommand, type InstallExecuteArgs } from '../../base/commands/index';
+import { runPython, runUV } from '../helpers';
+import { processEditableInstallArgs } from '../utils';
+
+/**
+ * Pip install command.
+ * Parsed command: `python -m pip install [--upgrade] <package>`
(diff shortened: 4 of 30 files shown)


src/managers/poetry/commands/remove.ts
```diff
@@ -0,0 +1,26 @@
+import { CommandConstructorOptions, UninstallCommand, type UninstallExecuteArgs } from '../../base/commands/index';
+import { runPoetry } from '../poetryPackageManager';
+
```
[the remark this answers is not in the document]
edvilme: Fixed in [b4f9ae5](https://github.com/microsoft/vscode-python-environments/pull/1621/commits/b4f9ae53cc6ef8454bc794c01673c43c8c010345)

src/managers/poetry/commands/showTopLevel.ts
```diff
@@ -0,0 +1,38 @@
+import { CommandConstructorOptions, ListDirectNamesCommand, type BaseExecuteArgs } from '../../base/commands/index';
+import { runPoetry } from '../poetryPackageManager';
+
```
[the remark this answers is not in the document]
edvilme: Fixed in [b4f9ae5](https://github.com/microsoft/vscode-python-environments/pull/1621/commits/b4f9ae53cc6ef8454bc794c01673c43c8c010345)

src/managers/conda/condaPackageManager.ts
```diff
@@ -175,25 +164,12 @@ export class CondaPackageManager implements PackageManager, Disposable {
...
-                return Array.from(uniqueVersions.values()).sort((a: Pep440Version, b: Pep440Version) =>
-                    rcompare(a.public, b.public),
-                );
-            }
-            return undefined;
+            const availableVersionsCmd = new CondaAvailableVersionsCommand({
+                pythonExecutable: 'conda',
+                log: this.log,
+            });
+            const versionStrings = await availableVersionsCmd.execute({ packageName, pythonVersion: '' });
```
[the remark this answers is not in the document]
edvilme: Fixed in 00cc947

src/managers/poetry/poetryPackageManager.ts
```diff
@@ -230,78 +213,26 @@ export class PoetryPackageManager implements PackageManager, Disposable {
...
-        }
-
-        // Convert to Package objects using the API
-        return poetryPackages.map((pkg) => this.api.createPackageItem(pkg, environment, this));
+        const showCmd = new PoetryShowCommand({
+            pythonExecutable: poetry,
+            log: this.log,
+        });
+        const data = await showCmd.execute();
+        return (data ?? []).map((pkg) => this.api.createPackageItem(pkg, environment, this));
```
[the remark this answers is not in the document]
edvilme: Fixed in 186524a

src/managers/conda/condaPackageManager.ts
```diff
@@ -175,25 +172,15 @@ export class CondaPackageManager implements PackageManager, Disposable {
         packageName: string,
```
StellaHuang95, line 175: Copilot generated:
Conda available-versions ordering is inverted (verified). The old code sorted `rcompare(a.public, b.public)` (descending / newest-first); the new code sorts `.sort((a, b) => compare(a.public, b.public))` (ascending / oldest-first). The pip path in this same PR correctly uses `compare(b.public, a.public)` — conda is the odd one out, so the arg order was almost certainly not swapped when moving `rcompare`→`compare`. This is a user-visible regression in the "Manage Package Version" picker. Fix: `.sort((a, b) => compare(b.public, a.public))`.

[verified]

src/managers/builtin/commands/availableVersions.ts
```diff
@@ -0,0 +1,121 @@
...
+ * Official Documentation: https://docs.astral.sh/uv/pip/
+ * The `uv pip index versions` command lists all available versions of a package.
+ * The `--python-version` flag filters versions compatible with the specified Python version.
+ * The `--json` flag outputs results in JSON format for structured parsing.
+ */
+export class UvAvailableVersionsCommand extends AvailableVersionsCommand {
+    constructor(options: CommandConstructorOptions) {
+        super(options);
+    }
+
```
StellaHuang95: Copilot generated:
UV command shape changed from `uv tool run pip index versions ...` to `uv pip index versions ...`. The original deliberately ran real pip via `uv tool run` because uv's own pip shim historically does not implement `index`. If that's still true, every uv user's available-versions lookup fails, `execute` catches and returns `[]`, and the picker is silently empty. [unverified] — please run `uv pip index versions requests --json` against a current uv before merge; if it errors, restore `uv tool run pip`.

[verified]

src/managers/base/commands/packageManagerCommand.ts
```diff
@@ -0,0 +1,77 @@
...
+export interface BaseExecuteArgs {
+    cancellationToken?: CancellationToken;
+    showProgress?: boolean;
+}
+
+/**
+ * Constructor options shared by all package manager commands.
+ */
+export interface CommandConstructorOptions {
+    pythonExecutable: string;
```
StellaHuang95, line 16: Copilot generated:
Review rule `readonly-by-default`: `pythonExecutable`, `log`, `timeout`, and `config` are each assigned exactly once in the constructor and never reassigned, but none are `readonly`. Declare all four `readonly` (e.g. `protected readonly pythonExecutable: string;`). Per the review-rule floor this is non-downgradeable.

[verified]

src/managers/builtin/commands/install.ts
```diff
@@ -0,0 +1,96 @@
...
+import { CommandConstructorOptions, InstallCommand, type InstallExecuteArgs } from '../../base/commands/index';
+import { runPython, runUV, shouldUseUv } from '../helpers';
+import { processEditableInstallArgs } from '../utils';
+
+/**
+ * Pip install command.
+ *
+ * Parsed Command: `python -m pip install [--upgrade] [--index-url <url>] <package>`
+ *
+ * Official Documentation: https://pip.pypa.io/en/stable/cli/pip_install/
```
StellaHuang95: Copilot generated:
Review rule `readonly-by-default` (two violations): (1) `private indexUrl?: string;` in both `PipInstallCommand` and `UvInstallCommand` is assigned once in the constructor and never reassigned — make it `private readonly indexUrl?: string;`. (2) `let args = ['-m', 'pip', 'install'];` (and the UV `let args = ['pip', 'install', '--python', this.pythonExecutable];`) are only mutated via `.push()`, never reassigned — use `const`. Non-downgradeable per the rule floor.

[verified]

src/managers/builtin/utils.ts
```diff
@@ -22,13 +13,22 @@ import {
...
 } from '../common/nativePythonFinder';
 import { shortenVersionString, sortEnvironments } from '../common/utils';
-import { runPython, runUV, shouldUseUv } from './helpers';
-import { parsePipListJson, parseUvTree, PipPackage } from './pipListUtils';
 
 const PIXI_EXTENSION_ID = 'renan-r-santos.pixi-code';
 const PIXI_RECOMMEND_DONT_ASK_KEY = 'pixi-extension-recommend-dont-ask';
 let pixiRecommendationShown = false;
 
+/**
```
StellaHuang95: Copilot generated:
Review rule `readonly-by-default`: `parsePackageSpecs(packageStrings: string[])` only reads its parameter (via `.map`) and never mutates it, so the parameter should be `readonly string[]` (`ReadonlyArray<string>`). Non-downgradeable per the rule floor.

[verified]

src/managers/base/commands/packageManagerCommand.ts
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * Base class for all package manager commands.
+ * Provides common properties and minimal interface for subclasses.
+ */
+export abstract class PackageManagerCommand {
+    protected static readonly configSection?: string;
+
+    protected pythonExecutable: string;
+    protected log?: LogOutputChannel;
+    protected timeout: number = 300000;
```
eleanorjboyd, line 31: worried about this as a constant at the "base" level

src/managers/base/commands/packageManagerCommand.ts
```diff
@@ -0,0 +1,77 @@
...
+    ): Promise<T> {
+        if (!executeArgs?.showProgress) {
+            return this.execute(executeArgs) as Promise<T>;
+        }
+
+        return Promise.resolve(
+            window.withProgress(
+                {
+                    location: ProgressLocation.Notification,
+                    title: title ?? l10n.t('Running package manager command'),
```
eleanorjboyd: would be nice if this could be a bit more specific? that may be too hard tho idk

src/managers/builtin/commands/availableVersions.ts
```diff
@@ -0,0 +1,121 @@
+import {
+    AvailableVersionsCommand,
+    CommandConstructorOptions,
+    type AvailableVersionsExecuteArgs,
+} from '../../base/commands/index';
+import { runPython, runUV, shouldUseUv } from '../helpers';
```
eleanorjboyd: what do you think about the pip stuff being called builtin? I always found it a bit confusing but this is your ownership area so defer to your prefernce

src/managers/builtin/pipPackageManager.ts
```diff
@@ -125,7 +141,15 @@ export class PipPackageManager implements PackageManager, Disposable {
 
     async getPackages(environment: PythonEnvironment, options?: GetPackagesOptions): Promise<Package[] | undefined> {
         if (options?.skipCache || !this.packages.has(environment.envId.id)) {
-            const data = await refreshPipPackages(environment, this.log);
+            const pythonExecutable = environment.execInfo?.run?.executable;
```
eleanorjboyd: does this behavior match how that function did it? What do you think about checking and unpacking the env to the executable before going into the build function?

src/managers/builtin/pipPackageManager.ts
```diff
@@ -161,39 +178,38 @@ export class PipPackageManager implements PackageManager, Disposable {
...
+            const availableVersionsCmd = await BuiltinAvailableVersionsCommandFactory({
+                pythonExecutable,
+                log: this.log,
+            });
+
+            // For pip < 21.2.0, check version first
+            if (availableVersionsCmd instanceof PipAvailableVersionsCommand) {
+                const pipVersion = await this.getVersion(environment);
+                if (!pipVersion || compare(pipVersion.public, '21.2.0') < 0) {
+                    // pip <= 20.3.4 - version picking is undefined; no reliable machine-readable API exists.
```
eleanorjboyd: what pip version are we on? is it worth supporting this older one?

package.json
```diff
@@ -144,6 +144,156 @@
                     "description": "%python-envs.packageWatchers.description%",
                     "default": true,
                     "scope": "machine"
+                },
```
eleanorjboyd: what do you see as the use cases for these commands?

eleanorjboyd (commented):
very cool! Great idea getting some comphrehensive structure around the packaging commands! For the real PR would suggest splitting this up- maybe the base and then each package manager on top of that or smth since above 1k is a lot of lines for a PR. Thanks!!

src/managers/base/commands/packageManagerCommand.ts
edvilme: It is overrideable by the children commands

src/managers/base/commands/packageManagerCommand.ts
edvilme: The title can be passed as an argument. This is the fallback message to be shown

src/managers/builtin/commands/availableVersions.ts
edvilme: I find it a bit confusing too, would not mind changing it (although maybe that can be a follow-up PR?)

src/managers/builtin/pipPackageManager.ts
edvilme: Yes it does the same thing AFAIK

src/managers/builtin/pipPackageManager.ts
edvilme: Yes, we don't support that version, but keeping it because the command does not exist on previous versions, so it would break. Plus we already have the mechanism to get the version

package.json
edvilme: You are right, these are not needed rn

src/managers/poetry/commands/add.ts
```diff
@@ -0,0 +1,27 @@
...
+                }
+                return pkg.packageName;
+            }),
+        );
+
+        return args;
+    }
+
+    async execute(executeArgs: InstallExecuteArgs): Promise<void> {
+        await runPoetry(this.buildCommand(executeArgs), undefined, this.log, executeArgs.cancellationToken);
```
StellaHuang95: The `runPoetry` is accepting `cwd` as an optional arguments, but both add and remove are not passing it, do we need to worry about multi-root workspace scenarios?

src/managers/base/commands/availableVersions.ts
```diff
@@ -0,0 +1,27 @@
+import { BaseExecuteArgs, PackageManagerCommand } from './packageManagerCommand';
+
+/**
+ * Arguments for available versions command execution (change per execution).
+ */
+export interface AvailableVersionsExecuteArgs extends BaseExecuteArgs {
+    packageName: string;
+    pythonVersion: string;
+    includePrerelease?: boolean;
```
StellaHuang95, line 11: Is this used in extended classes such as conda?  CondaAvailableVersionsCommand doesn't seem to handle the prerelease flag.

src/managers/poetry/commands/add.ts
edvilme: Yes, you are right. Should be fixed now :)

edvilme: Addressed across the stack: `cwd` is now persistent command constructor state, all Poetry commands pass it to `runPoetry`, and the adoption layer resolves the project directory associated with the target environment for multi-root workspaces.

src/managers/base/commands/availableVersions.ts
edvilme: You are right. Fixed so that it now handles the prerelease option (it manually filters it)

edvilme:
Adding "Skip Tests" label since tests are added in a followup PR for simplicity (this PR only introduces the classes, it does not wire them up still)

edvilme:
Should also fix #1650

StellaHuang95 (approved):

StellaHuang95:
Did you also run some manual testing? Looks good to me, but would be nice to run some sanity tests with the big refactoring

edvilme:
Yes, did some manual testing and added some unit tests in #1677 too

src/managers/builtin/commands/install.ts
```diff
@@ -0,0 +1,89 @@
...
+        args.push(...processedArgs);
+
+        return args;
+    }
+
+    async execute(executeArgs: InstallExecuteArgs): Promise<void> {
+        const args = this.buildCommand(executeArgs);
+
+        await runPython(this.pythonExecutable, args, undefined, this.log, executeArgs.cancellationToken, this.timeout);
+    }
```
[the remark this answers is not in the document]
eleanorjboyd: did you get this resolved?

eleanorjboyd (approved):

PR #1610, opened 2026-06-24
Add empty InlineScriptEnvManager skeleton behind internal flag (PEP 723 PR 4/16)

StellaHuang95:
> Part of #1602 (PEP 723 inline script env support). Design doc: #1601.

### Roadmap context — where this PR sits

This is **PR 4 of 16** in the PEP 723 inline-script roadmap. The full plan lives in #1602; here's a one-line summary of where each PR sits relative to this one:

| Phase | PR | Status |
|---|---|---|
| **Phase 1 — Foundation** | PR 1: cache key hash utility | _in progress_ |
| | PR 2: cache layout + `meta.json` sidecar | _in progress_ |
| | PR 3: `requires-python` → interpreter selection | _in progress_ |
| **Phase 2 — Manager** | **PR 4: `InlineScriptEnvManager` skeleton** | **this PR** |
| | PR 5: `create()` happy path | not started (needs 1, 2, 3, 4) |
| | PR 6: `create()` uv-install fallback | not started (needs 3, 5) |
| | PR 7: persistence — `get` / `set` + Memento | not started (needs 4) |
| | PR 8: activation-time discovery | not started (needs 2, 4, 7) |
| **Phase 3 — Routing** | PR 9: route PEP 723 scripts to inline manager | not started (needs 4, 7) |
| | PR 10: per-script project registration | not started (needs 9) |
| **Phase 3.5 — Cross-repo** | PR 17 (pyrx), PR 18 (pyrx), PR 19 (vscode-python) | not started |
| **Phase 4 — UX** | PR 11: picker item, PR 12: bulk command | not started |
| **Phase 5 — Lifecycle & polish** | PR 13: clear cache, PR 14: TTL, PR 15: telemetry, PR 16: status bar | not started |

After PRs 1–4 merge, **PR 7** unlocks the largest downstream wave (PR 8, 9, 13, 17, 19) — see [the ordering comment on #1602](https://github.com/microsoft/vscode-python-environments/issues/1602#issuecomment-4794340391) for the full timeline.

### Why a skeleton

The roadmap rolls out across 16 PRs. Landing each one behind a feature gate keeps `main` shippable at every step and lets reviewers see one concern at a time. This PR is the **smallest possible mount point** for that gate: an `EnvironmentManager` implementation that satisfies the interface contract and registers cleanly, but otherwise does nothing.

Every subsequent PR (5–8) replaces one of the no-ops in this skeleton with the real thing.

### What this PR does

[...]

Issue #1602, opened 2026-06-22
Add PEP 723 inline script env support


StellaHuang95:
See https://github.com/microsoft/vscode-python-environments/pull/1601 for the design document.

### Phase 1 — Foundation (parallelizable, no behavior change)

- [x] **PR 1 — Cache key hash utility** — Pure functions for dependency-list normalization (sort, lowercase, strip whitespace) and SHA-256 truncation. Unit tests only. _Depends on: —_
- [x] **PR 2 — Cache layout + `meta.json` sidecar helpers** — Resolve `<globalStorage>/script-envs-v1/<hash>`; typed `MetaJson` interface; atomic read/write; pure TTL-eviction helper (given a list of entries, return the paths to delete). _Depends on: —_
- [x] **PR 3 — `requires-python` → interpreter selection** — Filter `api.getEnvironments('global')` via `matchesPythonVersion`; extract a lower-bound version (`">=3.13"` → `"3.13"`) for the uv-install fallback. _Depends on: —_

> All three can be developed in parallel.

### Phase 2 — Manager (internal only)

- [x] **PR 4 — `InlineScriptEnvManager` skeleton** — Class implementing `EnvironmentManager`; `displayName = "Inline script environments"`; registered in `extension.ts`. `getEnvironments` returns `[]`. No `create` / `set` yet. Smoke check: appears as an empty section in the picker. _Depends on: —_
- [x] **PR 5 — `create()` happy path** — Given `(scriptUri, metadata)`: pick a compatible installed Python (PR 3), compute the hash (PR 1), build via existing `createWithProgress`, install dependencies, and write `meta.json` (PR 2). No persistence. No uv-install fallback. _Depends on: 1, 2, 3, 4_
- [x] **PR 6 — `create()` uv-install fallback** — Extend the `promptInstallPythonViaUv` trigger union with `'inlineScript'`; thread the `requires-python` lower bound to `installPythonWithUv(version)`; wire it into `create()` for the no-compatible-interpreter case. _Depends on: 3, 5_
- [x] **PR 7 — Persistence: `get` / `set` + Memento** — New `INLINE_SCRIPT_ENVS_KEY`; per-URI `fsPathToEnv` map; cache-hit re-verification of `requires-python` (Q4 step 3). Mirrors the `VenvManager` pattern. _Depends on: 4_
[...]

PR #1634, opened 2026-07-14
Add inline-script cache key hash utility (PEP 723 PR 1/16)

StellaHuang95:

### Roadmap context

This is **PR 1 of 16** in the PEP 723 inline-script roadmap. The full plan lives in #1602.

| Phase | PR | Status |
|---|---|---|
| **Phase 1: Foundation** | **PR 1: cache key hash utility** | **this PR** |
| | PR 2: cache layout + `meta.json` sidecar | _in progress_ |
| | PR 3: `requires-python` to interpreter selection | _in progress_ |

### Why this PR

The cache key decides whether a PEP 723 script reuses an existing environment or builds a fresh one. Under the dependency-keyed design from Q4 of #1601, scripts using the same dependencies and interpreter should share an environment.

Trivial metadata edits such as changing package-name casing, separator style, whitespace, dependency order, or extras order should not fragment the cache.

This PR isolates that logic as pure functions with no filesystem access, global storage dependency, or extension activation behavior. The on-disk cache layout remains separate in PR 2.

### What this PR does

Adds `src/common/inlineScriptCacheKey.ts` with four exports:

1. **`normalizeDependency(dep)`**

   Canonicalizes common variants of the same requirement:

   - Applies PEP 503 normalization to project names and extras: lowercase and collapse runs of `.`, `_`, and `-` to a single `-`.
   - Sorts and deduplicates extras after normalization.
   - Removes whitespace around version comparison operators.
   - Preserves meaningful version and marker content.
   - Drops empty dependency entries when keys are computed.

   Examples:

   - `Django` and `django` normalize identically.
   - `Flask-Login`, `flask_login`, and `flask.login` normalize identically.
   - `requests[Socks,security]` becomes `requests[security,socks]`.
   - `requests < 3` becomes `requests<3`.

   This is intentionally not a complete PEP 508 parser. Inputs it cannot meaningfully parse remain deterministic without attempting semantic normalization.

2. **`normalizeInterpreterPath(interpreterPath)`**

   Uses the existing `normalizePath` helper so Windows path casing and separator differences do not fragment the cache. POSIX path casing remains unchanged.

3. **`computeCacheKey({ dependencies, interpreterPath })`**

   - Normalizes, deduplicates, and sorts dependencies.
   - Normalizes the interpreter path.
   - Builds a versioned, labelled payload.
   - Hashes the payload with SHA-256.
   - Returns the first 16 lowercase hexadecimal characters.

   The resulting key is deterministic and safe to use as a directory name under the cache root introduced by PR 2.

4. **`CACHE_KEY_HEX_LENGTH`**

   Exposes the key length so callers and tests do not duplicate the value `16`.

### Caller contract

`interpreterPath` must be absolute, trimmed, and already resolved through symlinks, for example with `fs.realpath()`.

The utility deliberately performs no I/O. Two different path strings that resolve to the same executable will therefore produce different keys unless the caller canonicalizes them first. The creation flow in PR 5 will own that resolution.

### Cache behavior

The script path and raw `requires-python` value are intentionally not included directly in the key.

- Two scripts with equivalent dependencies and the same selected interpreter share an environment.
- Moving or renaming a script does not invalidate its environment.
- Adding, removing, or repinning a dependency produces a new key.
- Selecting a different interpreter produces a new key.
- A later PR re-verifies `requires-python` on cache hits in case the constraint changes without changing the selected interpreter.

This supports the design's two outcomes: reuse an unchanged cache entry or build a fresh environment. There is no in-place dependency synchronization path.

### Tests

Adds `src/test/common/inlineScriptCacheKey.unit.test.ts` with 37 unit tests covering:

- PEP 503 project-name normalization.
- Extras normalization, sorting, deduplication, and whitespace handling.
- Version comparison operators and multi-clause specifiers.
- Empty and duplicate dependency entries.
- Dependency ordering and deterministic output.
- Inputs that should and should not change the key.
- Windows and POSIX interpreter-path behavior.
- Fixed-length lowercase hexadecimal output.
- Filesystem-safe output.
- Pinned caller-contract behavior such as trailing whitespace in interpreter paths.

Platform behavior is stubbed through `platformUtils`, keeping the suite independent of the host operating system.

### User impact

**Zero.**

Nothing in the extension calls these helpers yet. This PR adds no settings, commands, UI, logging, filesystem writes, or activation behavior. The utility is wired into environment creation in PR 5.

### Design context

This implements the cache-key portion of Q4 in #1601: a pipx-style cache keyed by normalized dependencies and the selected interpreter.

PEP 503 normalization for project names and extras was added in response to review feedback on the design PR.

### DIFF
--- a/src/common/inlineScriptCacheKey.ts
+++ b/src/common/inlineScriptCacheKey.ts
@@ -1,0 +1,103 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import { createHash } from 'crypto';
+import { normalizePath } from './utils/pathUtils';
+
+/** Length, in hex chars, of the cache key returned by {@link computeCacheKey}. 16 = 64 bits of SHA-256; fixed-length and filesystem-safe. */
+export const CACHE_KEY_HEX_LENGTH = 16;
+
+/**
+ * Inputs that participate in the cache key.
+ *
+ * **Caller contract**: `interpreterPath` must be an absolute path that the
+ * caller has already resolved through symlinks (e.g. via `fs.realpath()`).
+ * This function does no I/O. Two string-distinct paths that point to the
+ * same physical executable will produce two different cache keys.
+ */
+export interface CacheKeyInputs {
+    readonly dependencies: readonly string[];
+    readonly interpreterPath: string;
+}
+
+function normalizePep503Name(name: string): string {
+    return name.replace(/[-_.]+/g, '-').toLowerCase();
+}
+
+function normalizeExtras(inner: string): string {
+    const items = inner
+        .split(',')
+        .map((e) => e.trim())
+        .filter((e) => e.length > 0)
+        .map((e) => normalizePep503Name(e));
+    const deduped = Array.from(new Set(items)).sort();
+    return deduped.length > 0 ? `[${deduped.join(',')}]` : '';
+}
+
+/**
+ * Canonicalize a PEP 723 dependency entry so common variants of the same
+ * requirement produce identical strings. Not a full PEP 508 parser — only
+ * folds the superficial edits users are likely to make:
+ *
+ *   "Requests"                       → "requests"
+ *   "Flask-Login"                    → "flask-login"
+ *   "requests <3"                    → "requests<3"
+ *   "requests <3, !=2.0"             → "requests<3,!=2.0"
+ *   "Requests[Security,Socks]"       → "requests[security,socks]"
+ *   "requests[socks,security]"       → "requests[security,socks]"
+ *   "requests[security,Security]"    → "requests[security]"
+ *
+ * Extras (per PEP 503) are lowercased, separator-folded ([._-]+ → -),
+ * deduplicated, and sorted alphabetically — same canonical form pip and
+ * uv use for the project name, applied to each extra.
+ */
+export function normalizeDependency(dep: string): string {
+    const trimmed = dep.trim();
+    if (trimmed.length === 0) {
+        return '';
+    }
+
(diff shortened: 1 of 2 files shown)


brettcannon:
The normalization approach SGTM.

src/common/inlineScriptCacheKey.ts
```diff
@@ -0,0 +1,103 @@
...
+ *   "requests <3, !=2.0"             → "requests<3,!=2.0"
+ *   "Requests[Security,Socks]"       → "requests[security,socks]"
+ *   "requests[socks,security]"       → "requests[security,socks]"
+ *   "requests[security,Security]"    → "requests[security]"
+ *
+ * Extras (per PEP 503) are lowercased, separator-folded ([._-]+ → -),
+ * deduplicated, and sorted alphabetically — same canonical form pip and
+ * uv use for the project name, applied to each extra.
+ */
+export function normalizeDependency(dep: string): string {
```
eleanorjboyd, line 51: wondering if all of these should live in a PEP 508 file so ai can more easily find and re-use these helpers at other points

eleanorjboyd (dismissed):

src/common/inlineScriptCacheKey.ts
StellaHuang95: Yes it makes sense to move the normalizePEP503Name to a util file for common use, I will do that. But I would keep the remaining dependency canonicalization there because it intentionally implements only cache-equivalent transformations, not full PEP 508 parsing, treating it as reuseable could lead to incorrect use.

eleanorjboyd (approved):

rzhao271 (approved):

PR #1635, opened 2026-07-15
Add inline-script cache layout, meta.json sidecar, and env-usability guard (PEP 723 PR 2/16)

StellaHuang95:

### Roadmap context

This is **PR 2 of 16** in the PEP 723 inline-script roadmap. The full plan lives in #1602.

| Phase | PR | Status |
|---|---|---|
| **Phase 1: Foundation** | PR 1: cache key hash utility | in review (#1634) |
| | **PR 2: cache layout + `meta.json` sidecar** | **this PR** |
| | PR 3: `requires-python` to interpreter selection | _in progress_ |

### Why this PR

Q2 and Q3 of #1601 place inline-script environments under `<globalStorageUri>/script-envs-v1/<cache-key>/` and give each environment a `.meta.json` sidecar for lifecycle bookkeeping. Those contracts need to exist before the manager can create, discover, reuse, or clean up cached environments.

This PR also implements the cache-hit usability guard from Q4 step 4. Hashing the selected interpreter path catches a switch to a different Python path, but it does not catch an interpreter being uninstalled from the same path. The guard detects that stale cache state before the extension attempts to reuse a broken environment.

The filesystem concerns are kept separate from PR 1's pure hashing logic and PR 3's interpreter-selection logic.

### What this PR does

Adds `src/common/inlineScriptCacheLayout.ts` with cache path helpers, typed sidecar I/O, stale-entry selection, and an environment-usability check.

#### Cache layout

- `getScriptEnvCacheRoot(globalStorageUri)` returns `<globalStorageUri>/script-envs-v1/`.
- `getScriptEnvDir(globalStorageUri, cacheKey)` returns the directory for one cached environment.
- `getMetaJsonPath(envDir)` returns `<envDir>/.meta.json`.
- `INLINE_SCRIPT_CACHE_DIR_NAME`, `META_JSON_FILENAME`, and `META_SCHEMA_VERSION` centralize the on-disk contract.

The cache directory and metadata schema are both versioned. An incompatible future format should bump the cache directory suffix and schema version together rather than migrate environments in place.

#### Typed `.meta.json` sidecar

`InlineScriptEnvMeta` contains the fields with concrete downstream consumers:

- `schemaVersion`: validates the sidecar format.
- `scriptFsPath`: identifies the owning script for lifecycle cleanup.
- `lastUsedAt`: drives TTL eviction.
- `requiresPython?`: supports compatibility revalidation on cache hits.

`readMetaJson(envDir)` never throws. It returns `undefined` with a warning for missing or non-regular files, files larger than 1 MiB, malformed JSON, unknown schema versions, invalid field types, and non-canonical ISO timestamps. Unknown properties are dropped when the validated object is constructed.

`writeMetaJson(envDir, meta)` ensures the directory exists, writes formatted JSON to a randomized sibling temporary file, and atomically renames it to `.meta.json`. If writing or renaming fails, it attempts to remove the temporary file and rethrows the original error.

#### TTL selection

`selectStaleEntries(entries, now, ttlMs)` is a pure selector used by the future opportunistic cleanup path.

- An entry is stale only when its age is strictly greater than the TTL.
- Entries exactly at the TTL boundary are retained.
- Future timestamps are retained.
- Entries without `lastUsedAt` are retained because the extension should not delete data it cannot classify confidently.

The function returns paths to delete; it performs no filesystem walk or deletion itself.

#### Cache-hit usability guard

`verifyEnvUsable(envDir)` checks whether the cached environment still has a usable base interpreter. It never throws; failures produce a warning and return `false`.

- **POSIX:** stats `<envDir>/bin/python`. Because `fs.stat` follows symlinks, a dead launcher symlink is detected.
- **Windows:** reads `home = ...` from `pyvenv.cfg` and stats `<home>/python.exe`. Checking `<envDir>/Scripts/python.exe` would be insufficient because that launcher can remain after the base Python is uninstalled.

This covers common removal paths such as `pyenv uninstall`, `uv python uninstall`, Homebrew or apt removal, and Windows Add/Remove Programs.

### Failure policy

The helpers distinguish recoverable cache state from creation-time failures:

- Metadata reads and usability checks return `undefined` or `false` rather than throwing, allowing callers to treat invalid cache state as a miss.
- Metadata writes rethrow because a caller must know that newly created state was not recorded successfully.
- The TTL selector is pure and cannot delete anything by itself.

### Tests

Adds `src/test/common/inlineScriptCacheLayout.unit.test.ts` with 44 focused unit tests covering:

- Cross-platform cache and sidecar path construction.
- Metadata write/read round trips and concurrent writers.
- Atomic write behavior and temporary-file cleanup.
- Missing, malformed, oversized, and structurally invalid sidecars.
- Schema-version and canonical-timestamp validation.
- TTL boundaries, future timestamps, and missing `lastUsedAt` values.
- POSIX launchers, live symlinks, dead symlinks, and non-file paths.
- Windows `pyvenv.cfg` parsing and missing base interpreters.

Required checks pass on this branch:

- `npm run lint`
- `npm run compile-tests`
- `npm run unittest`: 1418 passing, 0 failing, 4 pending

### User impact

**Zero.**

Nothing in the extension calls these helpers yet. This PR creates no environments, writes nothing to global storage during activation, deletes no cache entries, and adds no settings, commands, UI, or logging on the default execution path.

PR 5 will use the layout and sidecar helpers during creation. PRs 7 and 8 will use them during reuse and activation-time discovery. PRs 13 and 14 will use the cleanup helpers.

### Design context

This implements Q2 (disk location), Q3 (environment contents and sidecar metadata), Q4 step 4 (base-interpreter existence guard), and the pure selection portion of Q7 (TTL cleanup) from #1601.

The base-interpreter guard was added in response to review feedback on the design PR.

### DIFF
--- a/src/common/inlineScriptCacheLayout.ts
+++ b/src/common/inlineScriptCacheLayout.ts
@@ -1,0 +1,246 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import * as crypto from 'crypto';
+import * as fsapi from 'fs-extra';
+import * as path from 'path';
+import { Uri } from 'vscode';
+import { traceWarn } from './logging';
+import { isWindows } from './utils/platformUtils';
+
+/**
+ * Versioned name of the cache root under the extension's `globalStorageUri`.
+ *
+ * Bump the `-v1` suffix together with {@link META_SCHEMA_VERSION} on any
+ * incompatible on-disk change, so old envs sit unread and TTL out naturally
+ * instead of being migrated in place.
+ */
+export const INLINE_SCRIPT_CACHE_DIR_NAME = 'script-envs-v1';
+
+export const META_JSON_FILENAME = '.meta.json';
+
+/**
+ * Schema version embedded in every {@link InlineScriptEnvMeta}.
+ */
+export const META_SCHEMA_VERSION = 1 as const;
+export interface InlineScriptEnvMeta {
+    readonly schemaVersion: typeof META_SCHEMA_VERSION;
+    readonly scriptFsPath: string;
+    readonly lastUsedAt: string;
+    readonly requiresPython?: string;
+}
+
+export function getScriptEnvCacheRoot(globalStorageUri: Uri): Uri {
+    return Uri.joinPath(globalStorageUri, INLINE_SCRIPT_CACHE_DIR_NAME);
+}
+
+export function getScriptEnvDir(globalStorageUri: Uri, cacheKey: string): Uri {
+    return Uri.joinPath(getScriptEnvCacheRoot(globalStorageUri), cacheKey);
+}
+
+export function getMetaJsonPath(envDir: Uri): Uri {
+    return Uri.joinPath(envDir, META_JSON_FILENAME);
+}
+
+const MAX_META_JSON_BYTES = 1024 * 1024;
+
+export async function readMetaJson(envDir: Uri): Promise<InlineScriptEnvMeta | undefined> {
+    const metaPath = getMetaJsonPath(envDir).fsPath;
+
+    try {
+        const stat = await fsapi.stat(metaPath);
+        if (!stat.isFile()) {
+            traceWarn(`inline-script meta: not a regular file at ${metaPath}`);
+            return undefined;
+        }
+        if (stat.size > MAX_META_JSON_BYTES) {
+            traceWarn(`inline-script meta: refusing to read ${metaPath} (${stat.size} bytes > cap)`);
+            return undefined;
+        }
+    } catch (err) {
+        if (isFileNotFoundError(err)) {
+            traceWarn(`inline-script meta: not found at ${metaPath}`);
+        } else {
+            const code = (err as NodeJS.ErrnoException | undefined)?.code ?? 'unknown';
+            traceWarn(`inline-script meta: failed to stat ${metaPath} (code=${code}):`, err);
+        }
+        return undefined;
+    }
+
+    let raw: string;
+    try {
+        raw = await fsapi.readFile(metaPath, 'utf8');
+    } catch (err) {
+        const code = (err as NodeJS.ErrnoException | undefined)?.code ?? 'unknown';
+        traceWarn(`inline-script meta: failed to read ${metaPath} (code=${code}):`, err);
+        return undefined;
+    }
+
+    let parsed: unknown;
+    try {
+        parsed = JSON.parse(raw);
+    } catch (err) {
+        traceWarn(`inline-script meta: malformed JSON in ${metaPath}:`, err);
+        return undefined;
+    }
+
+    const validated = validateMeta(parsed);
+    if (!validated) {
+        traceWarn(`inline-script meta: invalid shape in ${metaPath}`);
+        return undefined;
+    }
+    return validated;
+}
+
+/**
+ * Atomically write the `.meta.json` sidecar via temp-file + rename.
+ */
+export async function writeMetaJson(envDir: Uri, meta: InlineScriptEnvMeta): Promise<void> {
+    await fsapi.ensureDir(envDir.fsPath);
+    const finalPath = getMetaJsonPath(envDir).fsPath;
+    const tmpSuffix = crypto.randomBytes(6).toString('hex');
+    const tmpPath = `${finalPath}.tmp-${tmpSuffix}`;
+    const payload = JSON.stringify(meta, undefined, 2);
+    try {
+        await fsapi.writeFile(tmpPath, payload, 'utf8');
+        await fsapi.rename(tmpPath, finalPath);
+    } catch (err) {
+        await fsapi.remove(tmpPath).catch(() => undefined);
+        throw err;
+    }
+}
+
+/**
+ * Snapshot of one cached entry, populated by the (separate, I/O-doing) disk
+ * walk.
+ */
+export interface CacheEntrySummary {
+    readonly envDirPath: string;
+    readonly lastUsedAt: Date | undefined;
+}
+
+/**
+ * Pure selector: returns the env-dir paths whose age exceeds `ttlMs`.
+ */
+export function selectStaleEntries(entries: ReadonlyArray<CacheEntrySummary>, now: Date, ttlMs: number): string[] {
+    const stale: string[] = [];
+    const nowMs = now.getTime();
+    for (const entry of entries) {
+        if (entry.lastUsedAt === undefined) {
+            continue;
+        }
+        const ageMs = nowMs - entry.lastUsedAt.getTime();
+        if (ageMs > ttlMs) {
+            stale.push(entry.envDirPath);
+        }
+    }
+    return stale;
+}
+
+/**
+ * Verify that a cached env's base interpreter still exists on disk.
+ */
+export async function verifyEnvUsable(envDir: Uri): Promise<boolean> {
+    if (isWindows()) {
+        return verifyWindowsBaseInterpreter(envDir);
+    }
+    return verifyPosixBaseInterpreter(envDir);
+}
+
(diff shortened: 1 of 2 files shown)


brettcannon (commented):
Conceptually SGTM.

src/common/inlineScriptCacheLayout.ts
```diff
@@ -0,0 +1,246 @@
...
+            stale.push(entry.envDirPath);
+        }
+    }
+    return stale;
+}
+
+/**
+ * Verify that a cached env's base interpreter still exists on disk.
+ */
+export async function verifyEnvUsable(envDir: Uri): Promise<boolean> {
```
eleanorjboyd: this just tests it exists? may be worth updating the name since we don't necessarily know it is usable from the existence right?

src/common/inlineScriptCacheLayout.ts
```diff
@@ -0,0 +1,246 @@
...
+    return Uri.joinPath(getScriptEnvCacheRoot(globalStorageUri), cacheKey);
+}
+
+export function getMetaJsonPath(envDir: Uri): Uri {
+    return Uri.joinPath(envDir, META_JSON_FILENAME);
+}
+
+const MAX_META_JSON_BYTES = 1024 * 1024;
+
+export async function readMetaJson(envDir: Uri): Promise<InlineScriptEnvMeta | undefined> {
```
eleanorjboyd, line 75: so are these in the folder / location we wrote? wondering if we come across a malformed one, like is it our responsibility to clean it up? or do we assume it could be made by the user?

eleanorjboyd (dismissed):

src/common/inlineScriptCacheLayout.ts
StellaHuang95: Yes I just renamed it and its tests.

src/common/inlineScriptCacheLayout.ts
StellaHuang95: yes, right now the reads are intentional non destructive, when it comes to the cache clean up implementation we can think about the deletion of the invalid ones.

eleanorjboyd (dismissed):

src/common/inlineScriptCacheLayout.ts
edvilme, line 1: Maybe we can move all of the constants and interfaces to the top?

src/common/inlineScriptCacheLayout.ts
```diff
@@ -0,0 +1,255 @@
...
+export const META_JSON_FILENAME = '.meta.json';
+
+/**
+ * Schema version embedded in every {@link InlineScriptEnvMeta}.
+ */
+export const META_SCHEMA_VERSION = 1 as const;
+export interface InlineScriptEnvMeta {
+    readonly schemaVersion: typeof META_SCHEMA_VERSION;
+    readonly scriptFsPath: string;
+    readonly lastUsedAt: string;
```
edvilme, line 39: Will this not be reused somewhere else? If so, maybe we can move it to its own place. 
Additionally, I think we could add more commments on what each of the members represent, as well as mention how the `lastUsedAt` date will be formatted (I am thinking maybe like an ISO date)

src/common/inlineScriptCacheLayout.ts
```diff
@@ -0,0 +1,255 @@
...
+ * Pure selector: returns the env-dir paths whose age exceeds `ttlMs`.
+ */
+export function selectStaleEntries(entries: ReadonlyArray<CacheEntrySummary>, now: Date, ttlMs: number): string[] {
+    const stale: string[] = [];
+    const nowMs = now.getTime();
+    for (const entry of entries) {
+        if (entry.lastUsedAt === undefined) {
+            continue;
+        }
+        const ageMs = nowMs - entry.lastUsedAt.getTime();
```
edvilme, line 151: The `CacheEntrySummary::lastUsedAt` is defined as type string. We should either define it as type date, or a string-like date representation. Otherwise, we risk trying to parse an invalid date.

src/common/inlineScriptCacheLayout.ts
```diff
@@ -0,0 +1,255 @@
...
+}
+
+/**
+ * Verify that a cached env's base interpreter still exists on disk.
+ */
+export async function verifyBaseInterpreterExists(envDir: Uri): Promise<boolean> {
+    if (isWindows()) {
+        return verifyWindowsBaseInterpreter(envDir);
+    }
+    return verifyPosixBaseInterpreter(envDir);
```
edvilme:
```suggestion
    return isWindows() ? verifyWindowsBaseInterpreter(envDir) : verifyPosixBaseInterpreter(envDir);
```

src/common/inlineScriptCacheLayout.ts
```diff
@@ -0,0 +1,255 @@
...
+    const home = parsePyvenvHome(raw);
+    if (home === undefined) {
+        traceWarn(`inline-script env: no 'home =' line in ${pyvenvPath}`);
+        return false;
+    }
+    const launcherPath = path.join(home, 'python.exe');
+    return statRegularFile(launcherPath, 'base interpreter');
+}
+
+async function statRegularFile(filePath: string, label: string): Promise<boolean> {
```
edvilme: Maybe rename so it is clear what it returns (bool)?

src/common/inlineScriptCacheLayout.ts
```diff
@@ -0,0 +1,255 @@
...
+
+    return {
+        schemaVersion: META_SCHEMA_VERSION,
+        scriptFsPath: obj.scriptFsPath,
+        lastUsedAt: obj.lastUsedAt,
+        requiresPython: obj.requiresPython,
+    };
+}
+
+function isCanonicalIsoTimestamp(value: unknown): value is string {
```
edvilme, line 253: Instead of converting from string to date and then comparing, maybe we could use a regex first?

edvilme (commented):
Few comments, mostly related to structure and handling of ISO dates

rchiodo:

rchiodo (dismissed):
Approved via Review Center.

src/common/inlineScriptCacheLayout.ts
StellaHuang95: done

src/common/inlineScriptCacheLayout.ts
StellaHuang95: I kept `InlineScriptEnvMeta` in this module because it defines the sidecar schema owned by `readMetaJson` and `writeMetaJson`, and it is already exported for reuse. I added documentation for the interface and each field, including that `lastUsedAt` is the canonical UTC string produced by `Date.toISOString()`.

src/common/inlineScriptCacheLayout.ts
StellaHuang95: done

src/common/inlineScriptCacheLayout.ts
StellaHuang95: done

src/common/inlineScriptCacheLayout.ts
StellaHuang95: Yeah the type of `CacheEntrySummary.lastUsedAt` is already Date | undefined, the string type if the serialized json value. the read function validates the string before returning it. I added some docs to clarify that.

src/common/inlineScriptCacheLayout.ts
StellaHuang95: The current validation is stronger than a regex, it verifies if the string represents a real timestamp and if it uses the exact format we want. any invalid dates will be rejected.

eleanorjboyd (approved):

rchiodo (approved):
Approved via Review Center.

edvilme (commented):
Thanks for the changes, lgtm

PR #1636, opened 2026-07-15
Add inline-script interpreter selection helpers (PEP 723 PR 3/16)

StellaHuang95:

### Roadmap context

This is **PR 3 of 16** in the PEP 723 inline-script roadmap. The full plan lives in #1602.

| Phase | PR | Status |
|---|---|---|
| **Phase 1: Foundation** | PR 1: cache key hash utility | in review (#1634) |
| | PR 2: cache layout + `meta.json` sidecar | in review (#1635) |
| | **PR 3: `requires-python` to interpreter selection** | **this PR** |

### Why this PR

Before the inline-script manager can create an environment, it needs to answer two questions from Q4 of #1601:

1. Which installed base interpreter is the newest one compatible with the script's `requires-python` declaration?
2. If none is compatible, what lower-bound version can the existing uv installation flow request?

Both decisions are pure functions over an environment list and a PEP 440 specifier. Keeping them outside the manager makes the filtering, ranking, and fallback rules independently testable without environment discovery, UI, filesystem access, or process execution.

### What this PR does

Adds `src/common/inlineScriptInterpreter.ts` with two exported helpers.

#### `pickCompatibleInterpreter(installed, requiresPython)`

Filters the supplied environments and returns the newest usable Python 3 interpreter satisfying the script's constraint.

A candidate is rejected when:

- `env.error` is set.
- `env.version` is missing, empty, or does not begin with numeric release segments.
- The interpreter is not Python 3.
- The version does not satisfy `requiresPython` according to the existing `matchesPythonVersion` helper.

An undefined, empty, or whitespace-only constraint is treated as unconstrained. If no candidate qualifies, the function returns `undefined` so the creation flow can offer the uv fallback.

Candidates are ranked by numeric release segments in descending order. A leading `v` is tolerated, and prerelease, development, and local suffixes are ignored for ranking. Ties preserve input order. The input array is copied before sorting and is never mutated.

Examples:

- No constraint across Python 3.10, 3.11, and 3.12 selects 3.12.
- `>=3.11,<3.13` selects the newest interpreter within that range.
- `==3.12.*` uses the existing wildcard-matching semantics.
- `~=3.12.4` honors the compatible-release upper bound.
- If only Python 2, broken environments, or incompatible versions are available, the result is `undefined`.

#### Base-interpreter caller contract

`installed` must contain base interpreters only: system Python, pyenv-installed Python, uv-installed Python, or conda `base`. Derived environments such as venvs, named conda environments, Poetry environments, and Pipenv environments must not be used as venv bases.

`api.getEnvironments('global')` is the intended source. The helper still defensively rejects errored, versionless, unparseable, and non-Python-3 entries.

#### `extractLowerBoundVersion(requiresPython)`

Extracts the tightest usable lower bound for a future `uv python install <version>` request.

Supported floor-producing forms include:

- `>=3.13` to `3.13`
- `~=3.12.4` to `3.12.4`
- `==3.12.7` to `3.12.7`
- `==3.12.*` to `3.12`
- `>=3.11,>=3.12,<3.14` to `3.12`

When multiple lower bounds are present, the numerically greatest one is returned.

Recognized operators that do not provide a safe integer floor (`>`, `<`, `<=`, `!=`, and `===`) are skipped without warning. Malformed clauses, invalid wildcard placement, and `~=` values without at least major and minor segments are skipped with `traceWarn`.

A skipped clause does not discard a valid floor from another clause. For example, `>=3.11,<3.13` returns `3.11`, while an upper-bound-only spec returns `undefined`.

The extracted value is an installation hint, not a full compatibility result. PR 6 will re-run `matchesPythonVersion` after uv installs an interpreter.

### Tests

Adds `src/test/common/inlineScriptInterpreter.unit.test.ts` with 34 focused unit tests covering:

- Empty input and no-compatible-interpreter results.
- Unconstrained and constrained newest-version selection.
- Multi-clause, wildcard, and compatible-release constraints.
- Errored, versionless, unparseable, and Python 2 environments.
- Empty-string constraints and leading `v` prefixes.
- Stable ties and non-mutating input behavior.
- Ranking versions with prerelease, development, and local suffixes.
- Lower-bound extraction for `>=`, `==`, wildcard `==`, and `~=`.
- Tightest-floor selection across multiple clauses.
- Upper-bound-only and unsupported operators.
- Malformed clauses, wildcard misuse, and warning behavior.

Required checks pass on this branch:

- `npm run lint`
- `npm run compile-tests`
- `npm run unittest`: 1418 passing, 0 failing, 4 pending

### User impact

**Zero.**

Nothing in the extension calls these helpers yet. This PR discovers no interpreters, installs no Python versions, creates no environments, and adds no settings, commands, UI, filesystem access, or activation behavior.

PR 5 will call `pickCompatibleInterpreter` for the normal creation path. PR 6 will call `extractLowerBoundVersion` when no compatible installed interpreter exists and the user is offered a uv installation.

### Design context

This implements Q4 step 1 from #1601: select the newest installed Python satisfying `requires-python`, and derive a lower-bound version for the uv fallback when no installed interpreter qualifies.

It reuses the existing `matchesPythonVersion` implementation rather than introducing a second PEP 440 matcher.

### Explicit installation consent

Following review feedback, the uv fallback now has an explicit, informed-consent contract for inline scripts:

- The modal explains the script's `requires-python` constraint and the exact Python version requested.
- If uv is missing, both the message and action disclose that uv and Python will be installed.
- No prompt is offered when a compatible concrete install version cannot be derived.
- Script-controlled prompt text is flattened, bounded, and stripped of control characters; install versions must be numeric release segments.
- Dismissal performs no installation and remains retryable; inline-script setup does not inherit the global "Don't ask again" suppression state.
- Approval forwards the displayed version to `uv python install`, and installed-version lookup matches release-segment boundaries.

The inline-script manager still has no creation call site in this PR, so this establishes and tests the consent API without changing current user behavior. PR 6 will invoke it from the uv fallback.

### DIFF
--- a/src/common/inlineScriptInterpreter.ts
+++ b/src/common/inlineScriptInterpreter.ts
@@ -1,0 +1,217 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import { PythonEnvironment } from '../api';
+import { matchesPythonVersion } from './inlineScriptMetadata';
+import { traceWarn } from './logging';
+
+/**
+ * Pick the newest installed Python that can serve as a base interpreter for
+ * a PEP 723 script. Returns `undefined` if no candidate is usable (the
+ * caller is then expected to prompt for a uv install or surface an error).
+ *
+ * **Caller contract**: `installed` must contain only BASE interpreters
+ * (system Pythons, pyenv-installed, uv-installed, conda `base`) — never
+ * venvs / conda named envs / poetry / pipenv project envs. This function
+ * does not filter derived envs out, and using one as a venv base produces
+ * a nested or broken environment. `api.getEnvironments('global')` is the
+ * right source (with the caveat that pipenv's `'global'` scope is known
+ * to leak derived envs).
+ */
+export function pickCompatibleInterpreter(
+    installed: ReadonlyArray<PythonEnvironment>,
+    requiresPython: string | undefined,
+): PythonEnvironment | undefined {
+    const constraint = requiresPython && requiresPython.length > 0 ? requiresPython : undefined;
+    const candidates = installed.filter((env) => isUsableBaseInterpreter(env, constraint));
+    if (candidates.length === 0) {
+        return undefined;
+    }
+    const sorted = [...candidates].sort((a, b) => compareVersionsDescending(a.version, b.version));
+    return sorted[0];
+}
+
+/**
+ * Extract a lower-bound version string from a PEP 440 `requires-python`
+ * specifier, suitable as the `version` argument to `uv python install`.
+ *
+ * Examples:
+ *   ">=3.13"          → "3.13"
+ *   ">=3.11,<3.13"    → "3.11"   (tightest lower bound across clauses)
+ *   "~=3.12.4"        → "3.12.4"
+ *   "==3.12.*"        → "3.12"
+ *   "==3.12.7"        → "3.12.7"
+ *
+ * Returns `undefined` for specifiers without a clean lower bound (`<3.13`,
+ * `!=3.10`, `>3.12`, `===…`, illegal shapes like `~=3` or `>=3.*`). The
+ * caller falls back to the uv default and re-verifies with
+ * `matchesPythonVersion` after install.
+ */
+export function extractLowerBoundVersion(requiresPython: string | undefined): string | undefined {
+    if (!requiresPython) {
+        return undefined;
+    }
+    const clauses = requiresPython
+        .split(',')
+        .map((c) => c.trim())
+        .filter((c) => c.length > 0);
+    if (clauses.length === 0) {
+        return undefined;
+    }
+
+    let best: number[] | undefined;
+    let bestStr: string | undefined;
+    for (const clause of clauses) {
+        const lb = lowerBoundForClause(clause);
+        if (lb === undefined) {
+            continue;
+        }
+        if (best === undefined || compareReleaseSegments(lb.segments, best) > 0) {
+            best = lb.segments;
+            bestStr = lb.display;
+        }
+    }
+    return bestStr;
+}
+
+function isUsableBaseInterpreter(env: PythonEnvironment, requiresPython: string | undefined): boolean {
+    if (env.error) {
+        return false;
+    }
+    if (typeof env.version !== 'string' || env.version.length === 0) {
+        return false;
+    }
+    if (parseLeadingMajor(env.version) !== 3) {
+        return false;
+    }
+    if (requiresPython !== undefined && !matchesPythonVersion(requiresPython, env.version)) {
+        return false;
+    }
+    return true;
+}
+
+function parseLeadingMajor(version: string): number | undefined {
+    const m = version.match(/^\s*v?(\d+)/i);
+    if (!m) {
+        return undefined;
+    }
+    const n = Number.parseInt(m[1], 10);
+    return Number.isNaN(n) ? undefined : n;
+}
+
+function parseReleaseSegments(version: string): number[] | undefined {
+    const m = version.match(/^v?(\d+(?:\.\d+)*)/i);
+    if (!m) {
+        return undefined;
+    }
+    return m[1].split('.').map((s) => Number.parseInt(s, 10));
+}
+
+function compareReleaseSegments(a: ReadonlyArray<number>, b: ReadonlyArray<number>): number {
+    const n = Math.max(a.length, b.length);
+    for (let i = 0; i < n; i++) {
+        const av = a[i] ?? 0;
+        const bv = b[i] ?? 0;
+        if (av < bv) {
+            return -1;
+        }
+        if (av > bv) {
+            return 1;
+        }
+    }
+    return 0;
+}
+
+function compareVersionsDescending(a: string, b: string): number {
+    const aSeg = parseReleaseSegments(a);
+    const bSeg = parseReleaseSegments(b);
+    if (aSeg === undefined && bSeg === undefined) {
+        return 0;
+    }
+    if (aSeg === undefined) {
+        return 1;
+    }
+    if (bSeg === undefined) {
+        return -1;
+    }
+    return compareReleaseSegments(bSeg, aSeg);
+}
+
+const CLAUSE_RE = /^(===|~=|==|!=|>=|<=|>|<)\s*(.+)$/;
+
+interface LowerBound {
+    readonly segments: number[];
+    readonly display: string;
+}
+
+function lowerBoundForClause(clause: string): LowerBound | undefined {
+    const m = clause.match(CLAUSE_RE);
+    if (!m) {
+        traceWarn(`inline-script interpreter: unrecognized requires-python clause: ${JSON.stringify(clause)}`);
+        return undefined;
+    }
+    const op = m[1];
+    const raw = m[2].trim();
+
+    switch (op) {
+        case '>=': {
+            // Per PEP 440 wildcards are only legal with `==` / `!=`. Stay
+            // consistent with matchesPythonVersion (which rejects `>=X.*`)
+            // so we never hand uv a value the picker will then reject.
+            if (raw.endsWith('.*')) {
+                traceWarn(
+                    `inline-script interpreter: wildcards are only valid with '==' / '!=': ${JSON.stringify(clause)}`,
+                );
+                return undefined;
+            }
+            const segments = parseReleaseSegments(raw);
+            if (segments === undefined) {
+                return undefined;
+            }
+            return { segments, display: segmentsToString(segments) };
+        }
+        case '==': {
+            const literal = raw.endsWith('.*') ? raw.slice(0, -2) : raw;
+            const segments = parseReleaseSegments(literal);
+            if (segments === undefined) {
+                return undefined;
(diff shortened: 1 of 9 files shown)


StellaHuang95:
One feedback from Brett is to ask for user's permission to install Python using UV if none is compatible.

StellaHuang95:
Addressed in `3cc949e`. The inline-script uv fallback now requires an explicit modal approval, identifies the requirement and exact Python version, discloses when uv itself will also be installed, and forwards only the approved numeric version. If no concrete compatible version can be derived, it does not offer an install. Dismissal is side-effect-free and retryable, and tests cover consent copy, cancellation, validation/sanitization, telemetry, and the actual `uv python install <version>` task arguments.

src/common/inlineScriptInterpreter.ts
```diff
@@ -0,0 +1,218 @@
...
+    const m = version.match(/^\s*v?(\d+)/i);
+    if (!m) {
+        return undefined;
+    }
+    const n = Number.parseInt(m[1], 10);
+    return Number.isNaN(n) ? undefined : n;
+}
+
+function parseReleaseSegments(version: string): number[] | undefined {
+    const m = version.match(/^v?(\d+(?:\.\d+)*)/i);
```
brettcannon: Why concern for `v`?

src/common/inlineScriptInterpreter.ts
```diff
@@ -0,0 +1,218 @@
...
+        return false;
+    }
+    if (requiresPython !== undefined && !matchesPythonVersion(requiresPython, env.version)) {
+        return false;
+    }
+    return true;
+}
+
+function parseLeadingMajor(version: string): number | undefined {
+    const m = version.match(/^\s*v?(\d+)/i);
```
brettcannon: Why the concern for `v`?

brettcannon (commented):
I don't understand where a "v" prefix would show up?

src/common/inlineScriptInterpreter.ts
StellaHuang95: @brettcannon So my understanding is that the pep723 specs says "requires-python: A string that specifies the Python version(s) with which the script is compatible. The value of this field MUST be a valid [version specifier](https://packaging.python.org/en/latest/specifications/version-specifiers/#version-specifiers)." (See [here](https://packaging.python.org/en/latest/specifications/inline-script-metadata/#:~:text=requires%2Dpython%3A%20A%20string%20that%20specifies%20the%20Python%20version(s)%20with%20which%20the%20script%20is%20compatible.%20The%20value%20of%20this%20field%20MUST%20be%20a%20valid%20version%20specifier)). 

And if I look for what is a `valid version specifier`, it says it can have preceding v character. See [here](https://packaging.python.org/en/latest/specifications/version-specifiers/#preceding-v-character)

StellaHuang95: but I guess you're right, since this function is parsing existing python version not a declaration, actual CPython version can't have a leading `v`. Let me update it, but I think we should keep the other one.

src/common/inlineScriptInterpreter.ts
StellaHuang95: Same reason I mentioned below, a valid version specifier can have a leading `v`

src/common/inlineScriptInterpreter.ts
```diff
@@ -0,0 +1,218 @@
...
+
+function parseReleaseSegments(version: string): number[] | undefined {
+    const m = version.match(/^v?(\d+(?:\.\d+)*)/i);
+    if (!m) {
+        return undefined;
+    }
+    return m[1].split('.').map((s) => Number.parseInt(s, 10));
+}
+
+function compareReleaseSegments(a: ReadonlyArray<number>, b: ReadonlyArray<number>): number {
```
eleanorjboyd: maybe a short doc string here- not totally sure what release segments are

src/common/inlineScriptInterpreter.ts
```diff
@@ -0,0 +1,218 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
```
eleanorjboyd, line 2: general note on this file- its a lot of helper methods I feel could maybe be used elsewhere as they are just about versioning. Maybe lift them to a helper file so they can be useful elsewhere too

eleanorjboyd (dismissed):

src/common/inlineScriptInterpreter.ts
StellaHuang95: Addressed in `a03d0a3`. I moved the release-segment operations to `common/utils/pep440Release.ts` and added JSDoc defining release segments as the dotted numeric components of a version (for example, `3.12.4` becomes `[3, 12, 4]`). The docs also call out zero-padding (`3.12` compares equal to `3.12.0`) and which non-release components the parser intentionally omits.

src/common/inlineScriptInterpreter.ts
StellaHuang95: Addressed in `a03d0a3`. I extracted the generic `parseReleaseSegments` and `compareReleaseSegments` operations into `common/utils/pep440Release.ts` and now reuse them from both interpreter selection and inline metadata matching. I kept interpreter selection and lower-bound extraction in `inlineScriptInterpreter.ts` because those are specific to the inline-script workflow. I also added focused tests for the shared semantics.

eleanorjboyd (dismissed):

src/common/utils/pep440Release.ts
edvilme, line 1: I think we are already using [@rebovatebot/pep440](https://www.npmjs.com/package/@renovatebot/pep440?activeTab=versions) to handle pep440 versions in other places in the repo

src/common/inlineScriptInterpreter.ts
```diff
@@ -0,0 +1,196 @@
...
+ *   ">=3.11,<3.13"    → "3.11"   (tightest lower bound across clauses)
+ *   "~=3.12.4"        → "3.12.4"
+ *   "==3.12.*"        → "3.12"
+ *   "==3.12.7"        → "3.12.7"
+ *
+ * Returns `undefined` for specifiers without a clean lower bound (`<3.13`,
+ * `!=3.10`, `>3.12`, `===…`, illegal shapes like `~=3` or `>=3.*`). The
+ * caller falls back to the uv default and re-verifies with
+ * `matchesPythonVersion` after install.
+ */
```
edvilme, line 51: I think we already have some mechanism for doing this somewhere in the repo as well, although I am not entirely sure

src/managers/builtin/uvPythonInstaller.ts
```diff
@@ -22,6 +22,28 @@ import { isUvInstalled, resetUvInstallationCache } from './helpers';
...
+
+export interface UvPythonInstallPromptOptions {
+    readonly version?: string;
+    readonly requiresPython?: string;
+}
+
+const MAX_PROMPT_DETAIL_LENGTH = 120;
+const INSTALLABLE_PYTHON_VERSION = /^\d+(?:\.\d+)*$/;
+const PROMPT_CONTROL_CHARACTERS =
+    /[\u0000-\u001f\u007f-\u009f\u061c\u200b-\u200f\u202a-\u202e\u2060-\u206f\ufeff]/g;
```
edvilme: Maybe we can move the constants to the top to have them all together?

edvilme: Can we add some comments or explanations into what these regexes represent?

src/managers/builtin/uvPythonInstaller.ts
```diff
@@ -319,48 +344,72 @@ export async function installPythonViaUv(_log?: LogOutputChannel, version?: stri
...
+            : uvInstalled
+              ? UvInstallStrings.installPythonPrompt
+              : UvInstallStrings.installPythonAndUvPrompt;
+    const installAction = !uvInstalled
+        ? version
+            ? UvInstallStrings.installUvAndPythonVersion(version)
+            : UvInstallStrings.installUvAndPython
+        : version
+          ? UvInstallStrings.installPythonVersion(version)
+          : UvInstallStrings.installPython;
```
edvilme: Can we switch it to
```js
const installAction = uvInstalled ? version [...]
``` 
so it is easier to read?

src/common/utils/pep440Release.ts
StellaHuang95: done

src/common/inlineScriptInterpreter.ts
StellaHuang95: checked the repo didn't find the exact equivalent functions so for now I'll just keep the workflow specific extraction.

edvilme (approved):
:)

src/managers/builtin/uvPythonInstaller.ts
StellaHuang95: done

src/managers/builtin/uvPythonInstaller.ts
StellaHuang95: sure, this is just because we are getting the python version of prompt message from the pep723 script, and copilot had suggestion of preventing harmful insertions.

eleanorjboyd (approved):

PR #1641, opened 2026-07-19
Fix #1599: onDidChangePythonProjects never fires on runtime project a…

### LINKED ISSUE #1599: Public onDidChangePythonProjects API event never fires (runtime project add/remove not reported to consumers)
## Summary

The public `PythonEnvironmentApi.onDidChangePythonProjects` event never fires. API consumers that subscribe to it to react to projects being added/removed at runtime never receive a notification. The internal project list is updated correctly (so a consumer that re-queries `getPythonProjects()` later — e.g. after a restart — does see the change), but no live event is emitted.

## Where

`src/features/pythonApi.ts` — `PythonEnvironmentApiImpl`:

The emitter is declared and exposed, but `_onDidChangePythonProjects.fire(...)` is **never called** anywhere in the class:

```ts
private readonly _onDidChangePythonProjects = new EventEmitter<DidChangePythonProjectsEventArgs>();
// ...
onDidChangePythonProjects: Event<DidChangePythonProjectsEventArgs> = this._onDidChangePythonProjects.event;
```

The constructor only forwards environment events — it pushes the emitter onto `disposables` (so it gets disposed) and subscribes to `envManagers.onDidChangeActiveEnvironment` → `_onDidChangeEnvironment` and `envVarManager.onDidChangeEnvironmentVariables` → `_onDidChangeEnvironmentVariables`. There is **no** subscription to `projectManager.onDidChangeProjects`.

Meanwhile the project manager **does** track and broadcast project changes internally (`src/features/projectManager.ts`, `PythonProjectManagerImpl`):

- `add()`, `remove()`, `updateProjects()`, and `loadProjects()` all call `this._onDidChangeProjects.fire(...)`.
- It exposes `public readonly onDidChangeProjects = this._onDidChangeProjects.event;`.

And the API's `addPythonProject()` / `removePythonProject()` delegate to `projectManager.add/remove`, which updates the internal map and fires the **internal** `onDidChangeProjects` — but the **public** `onDidChangePythonProjects` is never fired.

## Impact

Any API consumer relying on `onDidChangePythonProjects` to stay in sync with the active project set silently misses runtime changes (e.g. a new per-folder environment/project added via the UI). Workarounds require a full re-query or a restart of the consumer.

Concretely, this affects Pylance's per-folder interpreter support: when a user adds an environment/project for a subfolder, Pylance doesn't pick up the new (virtual) workspace folder until it is restarted, because Pylance subscribes to `onDidChangePythonProjects` and the event never fires.

## Steps to reproduce

1. Have an extension acquire the Python Environments API and subscribe to `onDidChangePythonProjects`.
2. Add a project/environment for a folder at runtime (UI or `addPythonProject`).
3. Observe: `getPythonProjects()` now includes the new project, but the `onDidChangePythonProjects` callback never fired.

## Expected

`onDidChangePythonProjects` should fire whenever the set of python projects changes (added/removed), consistent with its documented contract ("Event raised when python projects are added or removed").

## Suggested fix

Forward the project manager's change event to the public API emitter in the `PythonEnvironmentApiImpl` constructor, e.g.:

```ts
this.projectManager.onDidChangeProjects((/* projects */) => {
    this._onDidChangePythonProjects.fire(/* DidChangePythonProjectsEventArgs: added / removed */);
});
```

The project manager currently fires the full project array, so producing the `added`/`removed` shape of `DidChangePythonProjectsEventArgs` will require either tracking the previous set in the API layer, or updating the manager to emit the delta.

## Environment

- Repo: `microsoft/vscode-python-environments` (observed on `main`)
- Relevant files: `src/features/pythonApi.ts`, `src/features/projectManager.ts`

mohityadav8:
## Summary
Fixes #1599 — `PythonEnvironmentApi.onDidChangePythonProjects` never fired when projects were added or removed at runtime. The internal project list updated correctly, but the public event was never forwarded, so consumers (notably Pylance's per-folder interpreter support) never got notified and required a restart to pick up new projects.

## Root cause
`PythonEnvironmentApiImpl`'s constructor subscribed to environment-change events but never subscribed to`projectManager.onDidChangeProjects`, so `_onDidChangePythonProjects.fire(...)` was never called.

## Fix
- Added a `previousProjects` field on `PythonEnvironmentApiImpl` that tracks the last-known project set.
- Subscribed to `projectManager.onDidChangeProjects` in the constructor.  On each internal change, compute the added/removed delta by diffing  project URIs against the previous snapshot, update the snapshot, and  fire `_onDidChangePythonProjects` with `{ added, removed }` when the  set actually changed.

Diffing by URI (rather than relying on the manager's internal eventpayload) was needed because the manager currently emits the full projectarray rather than a delta, and `DidChangePythonProjectsEventArgs` expects`added`/`removed`.

## Testing
- `npm run compile` — clean build.
- Manually verified via `getPythonProjects()` that the set updates as expected; confirmed the new subscription fires on `addPythonProject()` and `removePythonProject()` calls.

### DIFF
--- a/src/features/pythonApi.ts
+++ b/src/features/pythonApi.ts
@@ -94,12 +94,16 @@ export class PythonEnvironmentApiImpl implements PythonEnvironmentApi {
             this.envVarManager.onDidChangeEnvironmentVariables((e) => this._onDidChangeEnvironmentVariables.fire(e)),
             this.projectManager.onDidChangeProjects(() => {
                 const current = this.projectManager.getProjects();
-                const added = current.filter(
-                    (c) => !this.previousProjects.some((p) => p.uri.toString() === c.uri.toString()),
-                );
-                const removed = this.previousProjects.filter(
-                    (p) => !current.some((c) => c.uri.toString() === p.uri.toString()),
-                );
+                const currentByUri = new Map(current.map((p) => [p.uri.toString(), p] as const));
+                const previousByUri = new Map(this.previousProjects.map((p) => [p.uri.toString(), p] as const));
+
+                const added = [...currentByUri.entries()]
+                    .filter(([uri]) => !previousByUri.has(uri))
+                    .map(([, project]) => project);
+                const removed = [...previousByUri.entries()]
+                    .filter(([uri]) => !currentByUri.has(uri))
+                    .map(([, project]) => project);
+
                 this.previousProjects = current;
                 if (added.length > 0 || removed.length > 0) {
                     traceInfo(
--- a/src/features/pythonApi.ts
+++ b/src/features/pythonApi.ts
@@ -33,6 +33,7 @@ import {
 } from '../api';
 import { traceError, traceInfo } from '../common/logging';
 import { pickEnvironmentManager } from '../common/pickers/managers';
+import { timeout } from '../common/utils/asyncUtils';
 import { createDeferred } from '../common/utils/deferred';
 import { checkUri } from '../common/utils/pathUtils';
 import { handlePythonPath } from '../common/utils/pythonPath';
@@ -44,7 +45,6 @@ import {
     PythonPackageImpl,
     PythonProjectManager,
 } from '../internal.api';
-import { timeout } from '../common/utils/asyncUtils';
 import { waitForAllEnvManagers, waitForEnvManager, waitForEnvManagerId } from './common/managerReady';
 import { EnvVarManager } from './execution/envVariableManager';
 import { runAsTask } from './execution/runAsTask';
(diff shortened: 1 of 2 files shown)


mohityadav8:
cc @eleanorjboyd

src/features/pythonApi.ts
```diff
@@ -58,12 +58,15 @@ import { TerminalManager } from './terminal/terminalManager';
 const GET_ENVIRONMENT_TIMEOUT_MS = 1000;
 const GET_ENVIRONMENT_TIMED_OUT = Symbol('getEnvironmentTimedOut');
 
-class PythonEnvironmentApiImpl implements PythonEnvironmentApi {
+export class PythonEnvironmentApiImpl implements PythonEnvironmentApi {
```
edvilme, line 61: Is there a reason we are exporting this?

edvilme:
Hello @mohityadav8  thanks for the contribution!! Everything looks great, but there are some linting issues around the use of `any` in TS. Once they get fixed, we can merge this! 😄

src/features/pythonApi.ts
mohityadav8: Hi @edvilme! I exported `PythonEnvironmentApiImpl` so I could import and instantiate it directly in the new unit test file (`src/test/features/pythonApi.unit.test.ts`) to verify onDidChangePythonProjects. If you prefer a different factory/internal access pattern for unit testing this class, let me know and I'd be happy to adjust!

mohityadav8:
@edvilme thanks for review i have updated my pr now blint check is passing thanks for reviewing this

edvilme (dismissed):

mohityadav8:
@edvilme everthing is solved :D

eleanorjboyd:
@mohityadav8 could you take a look at the copilot comments- I agree that they are reasonable. Otherwise thanks so much for the contribution!

mohityadav8:
> @mohityadav8 could you take a look at the copilot comments- I agree that they are reasonable. Otherwise thanks so much for the contribution!

Solved copilot review in `e14f6e4`

edvilme:
Thanks so much for your contribution and fixes. PR looks good to me! :D

edvilme (approved):

StellaHuang95 (approved):

eleanorjboyd (approved):

PR #1642, opened 2026-07-19
🐛 fire onDidChangePythonProjects event on project change

Yashraj-Jangra:
Fixes the issue where the public `onDidChangePythonProjects` API event never fired when projects were added or removed at runtime.

### Cause
The public `onDidChangePythonProjects` emitter was declared in `PythonEnvironmentApiImpl` but was never actually fired. The API layer also wasn't subscribed to the project manager's internal project change events.

### Changes
- Exposed `PythonEnvironmentApiImpl` for unit testing.
- Subscribed to `projectManager.onDidChangeProjects` in the constructor.
- Added a private cache of previous projects to compute the delta (`added` and `removed` projects) on changes, firing the public event when projects are modified.

### Verification
- Added a unit test in `src/test/features/pythonApi.unit.test.ts` to verify the delta logic.
- Added an integration test in `src/test/integration/pythonProjects.integration.test.ts`.
- Verified all unit tests pass locally.

### DIFF
--- a/src/features/pythonApi.ts
+++ b/src/features/pythonApi.ts
@@ -58,12 +58,13 @@ import { TerminalManager } from './terminal/terminalManager';
 const GET_ENVIRONMENT_TIMEOUT_MS = 1000;
 const GET_ENVIRONMENT_TIMED_OUT = Symbol('getEnvironmentTimedOut');
 
-class PythonEnvironmentApiImpl implements PythonEnvironmentApi {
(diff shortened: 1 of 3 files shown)


Yashraj-Jangra:

edvilme:
Thx for the contribution. Unfortunately this is already solved by #1641 
However we encourage you to contribute on any open issues or things you would like to see improved in the extension, and we will review!
Best!

PR #1646, opened 2026-07-22
implement create env

StellaHuang95:


StellaHuang95:

src/common/lockfile.apis.ts
```diff
@@ -0,0 +1,92 @@
...
+    );
+}
+
+function createLockError(message: string, code: string, lockPath: string): NodeJS.ErrnoException {
+    return Object.assign(new Error(message), { code, path: lockPath });
+}
+
+async function delay(milliseconds: number): Promise<void> {
+    return new Promise((resolve) => setTimeout(resolve, milliseconds));
+}
```
StellaHuang95: Add a trailing newline to this file and `src/test/common/lockfile.apis.unit.test.ts`; both new files currently lack one, which will fail the repository’s Prettier check.

src/managers/builtin/inlineScriptEnvManager.ts
```diff
@@ -1,27 +1,85 @@
...
+
+// Waiters never steal a lock: interrupted/cancelled creation remains fail-closed.
+// The planned PR13 inline-script clear-cache lifecycle flow will own recovery for abandoned locks.
+const CACHE_LOCK_TIMEOUT_MS = 5 * 60 * 1000;
+const CACHE_LOCK_RETRY_MS = 500;
+const INLINE_SCRIPT_MANAGER_ID = `${PYTHON_EXTENSION_ID}:inline-script`;
+
+interface SelectedBaseInterpreter {
+    readonly environment: PythonEnvironment;
+    readonly canonicalPath: string;
```
StellaHuang95: Copilot generated:
📍 src/managers/builtin/inlineScriptEnvManager.ts (type block before the class)
Confirmed review-rule violation (typescript-named-types-at-bottom, non-downgradeable): `SelectedBaseInterpreter`, `BuildCacheEntryResult`, `CacheEntryInspection`, `EnvironmentInspection`, and `PythonReleaseComparison` are declared ahead of the class, pushing the implementation down. Move these type/interface declarations to the bottom of the file, after the `InlineScriptEnvManager` class.

[verified]

src/managers/builtin/inlineScriptEnvManager.ts
```diff
@@ -61,6 +179,329 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
...
+            }
+            try {
+                return { environment, canonicalPath: await fs.realpath(executable) };
+            } catch (error) {
+                this.log.warn(
+                    `Skipping base interpreter that cannot be resolved at ${executable}: ${this.errorMessage(error)}`,
+                );
+            }
+        }
+
```
StellaHuang95, line 208: Copilot generated:
📍 src/managers/builtin/inlineScriptEnvManager.ts:buildCacheEntry / createOrReuseEnvironment
[unverified] Cross-cutting concern raised by Skeptic, Architect, and Advocate: on a cancelled install `retainLock` nulls the release, and because the lock never steals, every subsequent `create()` for the same deps+interpreter busy-retries for `CACHE_LOCK_TIMEOUT_MS` (5 min) then returns `undefined`, with no in-product recovery until PR13. Documented, tested, and behind the internal flag (so not a merge blocker), but the flag must not be enabled by default without a recovery escape hatch (e.g. a minimal "clear inline-script cache" command or fast-fail on a retained lock).

[verified]

src/managers/builtin/inlineScriptEnvManager.ts
```diff
@@ -61,6 +179,329 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
...
+        let sidecarResult;
+        try {
+            sidecarResult = await inspectMetaJson(envDir);
+        } catch {
+            return { kind: 'uncertain' };
+        }
+        if (sidecarResult.kind !== 'valid') {
+            return { kind: sidecarResult.kind === 'unavailable' ? 'uncertain' : 'stale' };
```
StellaHuang95, line 303: Copilot generated:
📍 src/managers/builtin/inlineScriptEnvManager.ts:inspectCacheEntry (comparePythonReleases path)
[unverified] Skeptic finding: when a cached inline-owned entry's resolved version is non-PEP440 (e.g. "Unknown"), `comparePythonReleases` returns `uncertain`, so the entry is preserved and never rebuilt — `create()` returns `undefined` on every call with only a warn log until manual cache clearing. Since the containment checks already prove the entry is physically ours, treat an unparseable version on an inline-manager-owned entry as `stale` (rebuild) rather than `uncertain` (preserve).

[verified]

src/managers/builtin/inlineScriptEnvManager.ts
```diff
@@ -61,6 +179,329 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
...
+                    return undefined;
+                }
+            }
+
+            const build = await this.buildCacheEntry(envDir, cacheRoot, packages, selectedBase);
+            if (build.retainLock) {
+                release = undefined;
+            }
+            return build.environment;
+        } catch (error) {
```
StellaHuang95, line 256: Copilot generated:
📍 src/managers/builtin/inlineScriptEnvManager.ts (private containment/path helpers)
Architect finding (High): `isPhysicallyContained`, `isExpectedPhysicalEntry`, `isDescendantPath`, and `getVenvPythonPath` are security-critical cache-layout logic living as private methods on the manager. `inlineScriptCacheLayout.ts` already owns the analogous helpers, and the planned PR13 delete path will re-derive "contained" — risking divergence between create-safety and delete-safety definitions with data-loss stakes. Extract these into the shared cache-layout module (e.g. exported `resolveVenvPythonPath` and `assertContainedCacheEntry`) before PR13.

[verified]

src/managers/builtin/inlineScriptEnvManager.ts
```diff
@@ -61,6 +179,329 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
...
+                derived: await fs.pathExists(path.join(environment.sysPrefix, 'pyvenv.cfg')),
+            })),
+        );
+        let candidates = derivedChecks.filter((candidate) => !candidate.derived).map((candidate) => candidate.environment);
+
+        while (candidates.length > 0) {
+            const environment = pickCompatibleInterpreter(candidates, metadata.requiresPython);
+            if (!environment) {
+                return undefined;
+            }
```
StellaHuang95, line 193: Copilot generated:
📍 src/managers/builtin/inlineScriptEnvManager.ts:getVenvPythonPath
Architect finding (Medium): the `win32 ? Scripts/python.exe : bin/python` rule is now duplicated in at least three places (`venvUtils.createWithProgress`, this `getVenvPythonPath`, and the test helper `venvPythonPath`). Consolidate into a single exported helper so a platform-layout change is one edit, not three.

[verified]

src/managers/builtin/inlineScriptEnvManager.ts
```diff
@@ -61,6 +179,329 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
...
 
+    private getScriptUri(scope: CreateEnvironmentScope): Uri | undefined {
+        const uri = scope instanceof Uri ? scope : Array.isArray(scope) && scope.length === 1 ? scope[0] : undefined;
+        return uri?.scheme === 'file' ? uri : undefined;
+    }
+
+    private async selectBaseInterpreter(metadata: InlineScriptMetadata): Promise<SelectedBaseInterpreter | undefined> {
+        const reported = (await this.api.getEnvironments('global')).filter(
+            (environment) =>
+                BASE_INTERPRETER_MANAGER_IDS.has(environment.envId.managerId) &&
```
StellaHuang95, line 168: Copilot generated:
📍 src/managers/builtin/inlineScriptEnvManager.ts:selectBaseInterpreter
[unverified] Skeptic edge case: `fs.pathExists(path.join(environment.sysPrefix, 'pyvenv.cfg'))` with an empty `sysPrefix` becomes a CWD-relative probe that could false-positive "derived" and silently drop a valid base interpreter. Guard against an empty/relative `sysPrefix` before the derived-venv check. Low likelihood.

[verified]

src/common/lockfile.apis.ts
StellaHuang95: Copilot generated:
Verified from diff: file ends with no trailing newline (`\ No newline at end of file`). This will fail the repo's Prettier `--check` CI gate and block merge. Add a trailing newline (same for `src/test/common/lockfile.apis.unit.test.ts`).

[verified]

src/common/lockfile.apis.ts
```diff
@@ -0,0 +1,92 @@
...
+    filePath: string,
+    options: AcquireFileLockOptions,
+): Promise<() => Promise<void>> {
+    const lockPath = `${path.resolve(filePath)}.lock`;
+    const ownerMarker = path.join(lockPath, `owner-${process.pid}-${crypto.randomBytes(16).toString('hex')}`);
+    const deadline = Date.now() + options.timeoutMs;
+
+    while (true) {
+        try {
+            await fsapi.mkdir(lockPath);
```
StellaHuang95, line 17: Copilot generated:
📍 src/common/lockfile.apis.ts:acquireFileLock (marker-write failure path)
[unverified] Skeptic finding: if `writeFile(ownerMarker, {flag:'wx'})` fails and the subsequent `rmdir(lockPath).catch(()=>undefined)` also fails/races, an owner-less lock dir persists and all future acquirers hit `EEXIST` → `ELOCKED` with no release closure to clean it. This is an accepted fail-closed edge, but log a distinct diagnostic so a future clear-cache flow can target owner-less lock dirs.

[verified]

src/common/lockfile.apis.ts
```diff
@@ -0,0 +1,92 @@
...
+            try {
+                await fsapi.writeFile(ownerMarker, '', { flag: 'wx' });
+            } catch (error) {
+                await fsapi.rmdir(lockPath).catch(() => undefined);
+                throw error;
+            }
+
+            let released = false;
+            return async () => {
```
StellaHuang95: Copilot generated:
📍 src/common/lockfile.apis.ts (release closure)
Skeptic observation: `release()` throws `ERELEASED`/`ECOMPROMISED` rather than being idempotent. Correct today (each site releases at most once, `retainLock` nulls the reference), but fragile: any future refactor that double-invokes release, or a `finally` running after an early release, will throw a confusing error out of a cleanup path. Consider a no-op-on-second-call to harden.

[verified]

src/common/inlineScriptCacheKey.ts
```diff
@@ -31,6 +31,39 @@ function normalizeExtras(inner: string): string {
     return deduped.length > 0 ? `[${deduped.join(',')}]` : '';
 }
 
+function normalizeRequirementTail(value: string): string {
```
StellaHuang95, line 34: Copilot generated:
📍 src/common/inlineScriptCacheKey.ts:normalizeRequirementTail
Architect/Skeptic: the hand-rolled quote/backslash escape tracking is non-obvious (though traced as correct, including `\\`, escaped quotes, and unterminated quotes which degrade to a harmless extra cache-miss). Add a one-line comment stating the invariant: "PEP 508 marker string literal — preserve quoted content verbatim; backslash escapes the next char."

[verified]

src/managers/builtin/helpers.ts
```diff
@@ -73,12 +73,22 @@ export async function runUV(
             spawnOptions.timeout = timeout;
         }
         const proc = spawnProcess('uv', args, spawnOptions);
+        let cancellationRequested = false;
```
StellaHuang95: Copilot generated:
📍 src/managers/builtin/helpers.ts:runUV / runPython
Architect (Low): the new `cancellationRequested` guard makes `runUV` and `runPython` near-identical copies. Repo style discourages copy/paste variants — consider a shared spawn-with-cancellation helper. The cancellation logic itself is correct and well-tested.

[verified]

src/managers/builtin/inlineScriptEnvManager.ts
```diff
@@ -61,8 +156,340 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
...
+            }
+            try {
+                return { environment, canonicalPath: await fs.realpath(executable) };
+            } catch (error) {
+                this.log.warn(
+                    `Skipping base interpreter that cannot be resolved at ${executable}: ${this.errorMessage(error)}`,
+                );
+            }
+        }
+
```
StellaHuang95, line 208: Copilot generated:
📍 src/managers/builtin/inlineScriptEnvManager.ts:createOrReuseEnvironment (retainLock branch)
The Skeptic found a new (Low) bug in the retain path: `lock` is set to `undefined` *before* `retainedLock.retain()` is awaited, so if `retain()` rejects (e.g. `fs.rename` fails with EACCES/EPERM/EBUSY) the error is only logged and `finally` sees `lock === undefined` → no release. On disk the `owner-*` marker persists but `retained` was never written, so the intended fast-fail `ELOCKRETAINED` degrades to a 5-min `ELOCKED`, and a future `retained`-keyed cleanup won't detect the orphan. Fix: keep `lock` non-null until `retain()` resolves (release in `finally` on failure), or write the `retained` sentinel before unlinking the owner marker so a failed transition still leaves a detectable marker.

[verified]

src/managers/builtin/inlineScriptEnvManager.ts
```diff
@@ -61,8 +156,340 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
...
+                try {
+                    await retainedLock.retain();
+                } catch (error) {
+                    this.log.error(
+                        `Failed to mark the inline-script cache lock as retained: ${this.errorMessage(error)}`,
+                    );
+                }
+            }
+            return build.environment;
+        } catch (error) {
```
StellaHuang95, line 256: Copilot generated:
📍 src/managers/builtin/inlineScriptEnvManager.ts:inspectExpectedCacheEnvironment
Architect (High, pre-PR13): the containment *primitive* was correctly extracted to shared `resolveCacheEntryPath`/`isDescendantPath`, but the ownership *policy* — `inspectExpectedCacheEnvironment` plus the `INLINE_SCRIPT_MANAGER_ID` check and delete-gating — still lives private on the manager. PR13's delete path will re-derive "is this entry safely mine" with data-loss stakes; extract this into a shared `assertOwnedCacheEntry(environment, cacheRoot, envDir)` before then so create-safety and delete-safety cannot diverge.

[verified]

PR #1651, opened 2026-07-23
Add generic environment-creation utilities (PEP 723 PR 5a/16)

StellaHuang95:

> **Split for review (3 PRs).** Reviewers flagged the original PR 5 as too large, so it is split into three stacked PRs grouped by dependency layer:
> - **5a — generic env-creation utilities — this PR (#1651).** Based on `main`; independent; merges first.
> - **5b — inline-script cache + interpreter utilities — #1655.** Stacked on 5a.
> - **5c — `create()` happy path (manager + wiring) — #1656.** Stacked on 5b.
>
> Applied together the three PRs are byte-for-byte identical to the original single change. **Merge order: 5a → 5b → 5c.**

### Roadmap context

This is the first slice of **PR 5 of 16** in the PEP 723 inline-script roadmap. The full plan lives in #1602.

| Phase | PR | Status |
|---|---|---|
| **Phase 1: Foundation** | PR 1: cache key hash utility | merged (#1634) |
| | PR 2: cache layout + `meta.json` sidecar | merged (#1635) |
| | PR 3: `requires-python` to interpreter selection | merged (#1636) |
| | **PR 5a: generic env-creation utilities** | **this PR (#1651)** |
| | **PR 5b: inline-script cache + interpreter utilities** | **#1655** |
| | **PR 5c: `create()` happy path (manager + wiring)** | **#1656** |
| | PR 7: persistence with `get`, `set`, and Memento | not started (needs 4) |
| **Phase 4+: UX / lifecycle** | PRs 11-16 | not started |

### Why this PR

PR 5c implements `InlineScriptEnvManager.create()`. Before touching the manager, this PR lands the **generic, reusable primitives** it relies on — a cross-process file lock, a venv Python-path helper, a cancellation-hardened process runner, and two small `createWithProgress` options. None of this code is inline-script-specific, so it is reviewed on its own.

### What this PR adds

**Cross-process file lock** (`src/common/lockfile.apis.ts`, new): `acquireFileLock` uses an atomic `mkdir` of a `<path>.lock` directory plus a per-owner marker file, returning `AcquiredFileLock { release, retain }`. `retain()` writes a `retained` marker so a later acquirer **fails fast with `ELOCKRETAINED`** instead of waiting out the 5-minute timeout — used when a build is cancelled mid-flight. Distinct error codes (`ELOCKED`, `ELOCKRETAINED`, `ELOCKORPHANED`, `ECOMPROMISED`, `ERETAINFAILED`) separate contention from corruption.

**Shared `getVenvPythonPath`** (`src/common/utils/virtualEnvironment.ts`, new): returns `Scripts\python.exe` on Windows, else `bin/python`. Replaces an inline copy in `venvUtils` and is reused by 5b/5c.

**Hardened process helper** (`src/managers/builtin/helpers.ts`): `runUV` and `runPython` now share one `runProcess` implementation whose cancellation guards `kill()` in `try/catch` and still emits a clean `CancellationError` if the process errors after a cancel. Per-caller options preserve existing behavior (`collectStderr`, `logPrefix`).

**`venvUtils.ts`:** `createWithProgress` gains `CreateWithProgressOptions { trackUvEnvironment }`, and `CreateEnvironmentResult` gains `pkgInstallationCancelled` so a caller can tell cancellation apart from a real install failure. Existing callers are unaffected (both are optional / additive).

### Tests

- **`lockfile.apis.unit.test.ts`** — 9 tests: contention, retain/fail-fast, orphaned and compromised locks, and timeout.
- **`virtualEnvironment.unit.test.ts`** — 2 tests for `getVenvPythonPath` on Windows and POSIX.
- **`helpers.cancellation.unit.test.ts`** — 4 tests for `runProcess` cancellation safety.
- **`venvUtils.createWithProgress.unit.test.ts`** — 3 tests for `trackUvEnvironment` and `pkgInstallationCancelled`.

On this branch alone `npm run compile-tests` is clean and `npm run unittest` reports **1447 passing, 0 failing, 4 pending**.

### User impact

**None.** These are internal primitives with no new user-visible behavior. The refactors to `helpers.ts` and `venvUtils.ts` are behavior-preserving for existing callers.

### DIFF
--- a/src/common/lockfile.apis.ts
+++ b/src/common/lockfile.apis.ts
@@ -1,0 +1,135 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import * as crypto from 'crypto';
+import * as fsapi from 'fs-extra';
+import * as path from 'path';
+
+/** Acquire an atomic lock released only explicitly; interrupted operations remain locked. */
+export async function acquireFileLock(filePath: string, options: AcquireFileLockOptions): Promise<AcquiredFileLock> {
+    const lockPath = `${path.resolve(filePath)}.lock`;
+    const ownerMarker = path.join(lockPath, `owner-${process.pid}-${crypto.randomBytes(16).toString('hex')}`);
+    const retainedMarker = path.join(lockPath, 'retained');
+    const deadline = Date.now() + options.timeoutMs;
+
+    while (true) {
+        try {
+            await fsapi.mkdir(lockPath);
+            try {
+                await fsapi.writeFile(ownerMarker, '', { flag: 'wx' });
+            } catch (error) {
+                try {
+                    await fsapi.rmdir(lockPath);
+                } catch {
+                    throw createLockError(
+                        'Lock initialization failed and left an owner-less lock directory',
+                        'ELOCKORPHANED',
+                        lockPath,
+                    );
+                }
+                throw error;
+            }
+
+            let state: LockState = 'held';
+            return {
+                retain: async () => {
+                    if (state !== 'held') {
+                        return;
+                    }
+                    state = 'retained';
+                    try {
+                        await fsapi.writeFile(retainedMarker, '', { flag: 'wx' });
+                    } catch (error) {
+                        if (isAlreadyExistsError(error)) {
+                            return;
+                        }
+                        try {
+                            await fsapi.rename(ownerMarker, retainedMarker);
+                        } catch (renameError) {
+                            if (!isAlreadyExistsError(renameError)) {
+                                throw createLockError(
+                                    'Failed to mark the lock as retained',
+                                    'ERETAINFAILED',
+                                    lockPath,
+                                );
+                            }
+                        }
+                    }
+                },
+                release: async () => {
+                    if (state !== 'held') {
+                        return;
+                    }
+                    state = 'released';
+                    try {
+                        await fsapi.unlink(ownerMarker);
+                    } catch (error) {
+                        if (isFileNotFoundError(error)) {
+                            throw createLockError('Lock ownership was compromised', 'ECOMPROMISED', lockPath);
+                        }
+                        throw error;
+                    }
+                    await fsapi.rmdir(lockPath);
+                },
+            };
+        } catch (error) {
+            if (!isAlreadyExistsError(error)) {
+                throw error;
+            }
+            if (await isRetainedLock(lockPath)) {
+                throw createLockError('Lock was retained after an interrupted operation', 'ELOCKRETAINED', lockPath);
+            }
+            if (Date.now() >= deadline) {
+                throw createLockError('Lock is already being held', 'ELOCKED', lockPath);
+            }
+            await delay(Math.min(options.retryIntervalMs, Math.max(0, deadline - Date.now())));
+        }
+    }
+}
+
+function isAlreadyExistsError(error: unknown): boolean {
+    return hasErrorCode(error, 'EEXIST');
+}
+
+function isFileNotFoundError(error: unknown): boolean {
+    return hasErrorCode(error, 'ENOENT');
+}
+
+async function isRetainedLock(lockPath: string): Promise<boolean> {
+    try {
+        await fsapi.lstat(path.join(lockPath, 'retained'));
+        return true;
+    } catch (error) {
+        if (isFileNotFoundError(error)) {
+            return false;
+        }
+        throw error;
+    }
+}
+
+function hasErrorCode(error: unknown, code: string): boolean {
+    return (
+        typeof error === 'object' && error !== null && 'code' in error && (error as NodeJS.ErrnoException).code === code
+    );
+}
+
+function createLockError(message: string, code: string, lockPath: string): NodeJS.ErrnoException {
+    return Object.assign(new Error(message), { code, path: lockPath });
+}
+
+async function delay(milliseconds: number): Promise<void> {
+    return new Promise((resolve) => setTimeout(resolve, milliseconds));
+}
+
+export interface AcquireFileLockOptions {
+    readonly timeoutMs: number;
+    readonly retryIntervalMs: number;
+}
+
+export interface AcquiredFileLock {
+    readonly release: () => Promise<void>;
+    /** Keep the lock and make later acquisition attempts fail immediately. */
+    readonly retain: () => Promise<void>;
+}
+
+type LockState = 'held' | 'released' | 'retained';
--- a/src/common/lockfile.apis.ts
+++ b/src/common/lockfile.apis.ts
@@ -5,6 +5,19 @@ import * as crypto from 'crypto';
 import * as fsapi from 'fs-extra';
 import * as path from 'path';
 
+export interface AcquireFileLockOptions {
+    readonly timeoutMs: number;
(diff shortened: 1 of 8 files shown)


src/common/utils/virtualEnvironment.ts
edvilme, line 1: I feel we must have something like this somewhere in the codebase, though I could be wrong

src/managers/builtin/helpers.ts
```diff
@@ -66,40 +66,14 @@ export async function runUV(
...
-            }
-        });
+    return runProcess('uv', args, {
+        cwd,
+        displayName: 'uv',
+        log,
+        token,
+        timeout,
+        collectStderr: false,
+        logPrefix: '',
```
edvilme: This is also used by other places of the codebase. Is there a reason for this change and the removal of the listeners and exceptions?

src/common/lockfile.apis.ts
```diff
@@ -0,0 +1,135 @@
...
+}
+
+function isAlreadyExistsError(error: unknown): boolean {
+    return hasErrorCode(error, 'EEXIST');
+}
+
+function isFileNotFoundError(error: unknown): boolean {
+    return hasErrorCode(error, 'ENOENT');
+}
+
```
edvilme: If these are not used anywhere else, I would just add the `hasErrorCode(error, 'XYZ')` inline

src/common/lockfile.apis.ts
```diff
@@ -0,0 +1,135 @@
...
+    readonly retryIntervalMs: number;
+}
+
+export interface AcquiredFileLock {
+    readonly release: () => Promise<void>;
+    /** Keep the lock and make later acquisition attempts fail immediately. */
+    readonly retain: () => Promise<void>;
+}
+
+type LockState = 'held' | 'released' | 'retained';
```
edvilme: Maybe move interfaces and types to the top of the file

src/common/lockfile.apis.ts
```diff
@@ -0,0 +1,135 @@
...
+import * as path from 'path';
+
+/** Acquire an atomic lock released only explicitly; interrupted operations remain locked. */
+export async function acquireFileLock(filePath: string, options: AcquireFileLockOptions): Promise<AcquiredFileLock> {
+    const lockPath = `${path.resolve(filePath)}.lock`;
+    const ownerMarker = path.join(lockPath, `owner-${process.pid}-${crypto.randomBytes(16).toString('hex')}`);
+    const retainedMarker = path.join(lockPath, 'retained');
+    const deadline = Date.now() + options.timeoutMs;
+
+    while (true) {
```
edvilme, line 28: Just a question here: why do we need the while loop? Not saying it is wrong, just curious in case I am not understanding correctly

src/managers/builtin/venvUtils.ts
```diff
@@ -340,9 +358,9 @@ export async function createWithProgress(
     venvRoot: Uri,
     envPath: string,
     packages?: PipPackages,
+    options?: CreateWithProgressOptions,
```
edvilme: Why do we need this to be another object instead of just passing a `trackUvEnvironment` property?

src/common/utils/virtualEnvironment.ts
StellaHuang95: There's one line code in src\managers\conda\condaUtils.ts that does something similar but it's not a helper, and the windows layouts are different, so I have this separate helper.

src/common/lockfile.apis.ts
StellaHuang95: yeah sure it's only used within the file, I can make it inline.

src/common/lockfile.apis.ts
StellaHuang95: sure, will do that.

src/managers/builtin/helpers.ts
StellaHuang95: It's just refactoring to get rid of the duplicate code. The error handling stuff is moved to the share runProcess. Maybe I shouldn't include refactor of a hot path in this pr, let me think about it.

src/managers/builtin/venvUtils.ts
StellaHuang95: It's mainly because so that the caller is more explicit about what's passed to the function and also it's open to extend to other options in the future. I could also make it a boolean value and add comment clarifying what that is in the caller.

src/common/lockfile.apis.ts
StellaHuang95: The loop prevents two VS Code processes from creating the same environment simultaneously.

For example:

1. Process A creates the lock and starts building the environment.
2. Process B tries to create the same lock but cannot because A owns it.
3. B waits briefly, then the  while (true)  loop tries again.
4. When A finishes, it removes the lock.
5. B’s next attempt succeeds. B can then reuse the environment A created.

The while(true) does not run forever. It either gets the lock, detects a retained lock, or reaches the timeout. Without the loop, Process B would fail immediately merely because Process A was still working.

src/managers/builtin/helpers.ts
StellaHuang95: ok made small fix instead of refactoring. This is to fix if os refused to kill process and throw, the fix makes sure the operation will still be reported as cancelled.

src/managers/builtin/venvUtils.ts
StellaHuang95: done

src/common/lockfile.apis.ts
StellaHuang95: done

src/common/lockfile.apis.ts
StellaHuang95: done

src/common/lockfile.apis.ts
```diff
@@ -0,0 +1,127 @@
...
+    );
+}
+
+function createLockError(message: string, code: string, lockPath: string): NodeJS.ErrnoException {
+    return Object.assign(new Error(message), { code, path: lockPath });
+}
+
+async function delay(milliseconds: number): Promise<void> {
+    return new Promise((resolve) => setTimeout(resolve, milliseconds));
+}
```
edvilme, line 127: Maybe this could be moved to a more generic location? But it doesn't bother me too much.

edvilme (approved):
LGTM:)

src/managers/builtin/helpers.ts
```diff
@@ -73,12 +73,22 @@ export async function runUV(
...
+            try {
+                proc.kill();
+            } catch {
+                // Preserve cancellation when signaling fails.
+            }
             reject(new CancellationError());
         });
 
         proc.on('error', (err) => {
+            if (cancellationRequested) {
```
eleanorjboyd, line 88: is cleanup required by this error stage? you went through and confirmed we aren't leaving stuff behind in people's file systems right?

eleanorjboyd (approved):

src/managers/builtin/helpers.ts
StellaHuang95: This change I made only catches `proc.kill()` throwing so the operation still reports `CancellationError` because it didn't catch the exception before, these 2 functions don't do cleanups, cleanup belongs to the caller. If you're asking about inline script manager in particular, that will be handled at the inline manager level later.

PR #1652, opened 2026-07-23
Add uv fallback for inline script environments (PEP 723 PR 6/16)

StellaHuang95:


src/managers/builtin/inlineScriptEnvManager.ts
```diff
@@ -1,27 +1,65 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
```
StellaHuang95, line 1: Copilot generated:
📍 src/managers/builtin/inlineScriptEnvManager.ts (buildCacheEntry / createOrReuseEnvironment)

A single user **Cancel** during package install permanently poisons this cache key with no recovery. The Skeptic and Architect both traced this (High): `pkgInstallationCancelled` → `buildCacheEntry` returns `{ retainLock: true }` *without* `removeCacheEntry`, `createOrReuseEnvironment` calls `lock.retain()`, and every future `create()` for the same deps+interpreter then hits `mkdir`→`EEXIST`→`isRetainedLock()`→throws `ELOCKRETAINED`→caught→returns `undefined`. The half-built env has no sidecar, so it never gets a `lastUsedAt` and TTL cleanup can't reclaim it either. While the Advocate notes retain-on-*interruption* is intentional fail-closed design, conflating a deliberate user Cancel with a crash makes this unrecoverable without manual globalStorage surgery. Treat cancellation distinctly: `removeCacheEntry(envDir)` + normal `release()` to allow retry, or make a retained-lock acquisition reclaim-and-rebuild. Add a cross-call regression test (the existing rollback suite stubs the lock, so it misses this).

[verified]

StellaHuang95, line 1: Copilot generated:
📍 src/managers/builtin/inlineScriptEnvManager.ts (createOrReuseEnvironment / buildCacheEntry)

Review-rule violation (`readonly-by-default.md`): `createOrReuseEnvironment(..., packages: string[], ...)` and `buildCacheEntry(..., packages: string[], ...)` only read `packages` (passed through to `createWithProgress({ install: packages })`) and never mutate it. Change both parameter types to `ReadonlyArray<string>` (or `readonly string[]`). Confirmed rule violations are non-downgradeable per the shared policy.

[verified]

src/test/managers/builtin/venvUtils.createWithProgress.unit.test.ts
```diff
@@ -0,0 +1,137 @@
...
+
+suite('createWithProgress uv tracking options', () => {
+    let addUvEnvironmentStub: sinon.SinonStub;
+    let api: PythonEnvironmentApi;
+    let baseEnvironment: PythonEnvironment;
+    let envPath: string;
+    let log: LogOutputChannel;
+    let manager: EnvironmentManager;
+    let nativeFinder: NativePythonFinder;
+    let tempRoot: string;
```
StellaHuang95, line 27: Copilot generated:
Review-rule violation (`platform-helper.md`): this test hard-codes `process.platform === 'win32' ? ... : ...` to build the python path, but the repo requires the `isWindows()` helper from `common/utils/platformUtils` — which this same PR already uses in product code (`virtualEnvironment.ts`, `inlineScriptCacheLayout.ts`). Use `isWindows()` or the new `getVenvPythonPath` here. Confirmed rule violations are non-downgradeable per the shared policy.

[verified]

StellaHuang95:
GitHub cannot anchor PR review comments to unchanged lines in the diff. Falling back to a general PR comment for [src/test/managers/builtin/inlineScriptEnvManager.unit.test.ts:L1](https://github.com/microsoft/vscode-python-environments/blob/e7ead2dbe1bc8204f67d6b4d492f6c6d338e7b01/src/test/managers/builtin/inlineScriptEnvManager.unit.test.ts#L1).

Copilot generated:
📍 src/test/managers/builtin/inlineScriptEnvManager.unit.test.ts (baseExecutable / uvExecutable construction, 3+ sites)

Review-rule violation (`platform-helper.md`): multiple raw `process.platform === 'win32'` checks are used to construct executable paths. Replace with the repo's `isWindows()` helper (or `getVenvPythonPath`), consistent with the product code this PR added. Confirmed rule violations are non-downgradeable per the shared policy.

[verified]

src/managers/builtin/venvUtils.ts
```diff
@@ -340,9 +358,9 @@ export async function createWithProgress(
     venvRoot: Uri,
```
StellaHuang95, line 358: Copilot generated:
📍 src/managers/builtin/venvUtils.ts (createWithProgress)

The Architect flags (Medium) that `createWithProgress` now takes 9 positional parameters, and the new tests assert against `firstCall.args[4]`/`args[6]`/`args[7]`/`args[8]` throughout. Any future reordering breaks many tests with opaque index failures, and the long positional list is itself a smell. Consider grouping the trailing arguments (`packages`, `options`, and the env/root URIs) into an options object to decouple callers and tests from arg order.

[verified]

src/common/inlineScriptCacheKey.ts
```diff
@@ -31,6 +31,40 @@ function normalizeExtras(inner: string): string {
     return deduped.length > 0 ? `[${deduped.join(',')}]` : '';
 }
 
+function normalizeRequirementTail(value: string): string {
```
StellaHuang95, line 34: Copilot generated:
📍 src/common/inlineScriptCacheKey.ts (normalizeRequirementTail)

The backslash-escape branch of this hand-rolled quote state machine is untested — every test uses unescaped literals (`"X  Y"`, `'0'`). Since this feeds cache-key canonicalization, a future regression here silently causes cache collisions (wrong env reused) or misses (rebuilds). The Advocate hand-traced `'a\'b'` as correct but flags it as unpinned (Medium). Add a unit test for an escaped quote inside a marker literal asserting the inner escaped quote is preserved and does not prematurely terminate the quoted region.

[verified]

src/managers/builtin/inlineScriptEnvManager.ts
StellaHuang95: The recoverability concern makes sense, but I kept the fail-closed behavior here. Cancellation currently rejects immediately after proc.kill(), before the child reports close, and a package-manager child may still be writing into the cache path. Removing the entry and releasing the lock at that point could race a surviving pip/uv process and allow a retry to rebuild underneath it. PR 13's guarded clear-cache flow is the planned recovery path and needs to land before the feature becomes user-facing.

StellaHuang95: Addressed in the corrected parent stack: both helper parameters are now ReadonlyArray<string>, with a clone only at the mutable PipPackages boundary.

src/test/managers/builtin/venvUtils.createWithProgress.unit.test.ts
StellaHuang95: Addressed in the corrected parent stack by using getVenvPythonPath(envPath).

src/managers/builtin/venvUtils.ts
StellaHuang95: Agreed that the positional API is worth improving, but I am deferring this refactor. The helper already had eight positional parameters, and converting it now would touch all four production callers plus unrelated tests without changing behavior. A request/options-object refactor will be safer as a dedicated change when another option is needed.

src/common/inlineScriptCacheKey.ts
StellaHuang95: Addressed in the corrected cache-utils parent commit with an escaped-quote marker test using String.raw, including preserved whitespace after the escaped quote.

StellaHuang95:
Addressed the platform-helper feedback: inherited PR 5 tests now use isWindows()/getVenvPythonPath, and the three PR 6 uv-executable cases use isWindows().

PR #1653, opened 2026-07-23
Add inline script environment persistence (PEP 723 PR 7/16)

StellaHuang95:


src/common/inlineScriptCacheLayout.ts
```diff
@@ -25,6 +23,7 @@ export const META_JSON_FILENAME = '.meta.json';
 export const META_SCHEMA_VERSION = 1 as const;
 
 const MAX_META_JSON_BYTES = 1024 * 1024;
+const INLINE_SCRIPT_MANAGER_ID = `${PYTHON_EXTENSION_ID}:inline-script`;
```
StellaHuang95: Copilot generated:
`INLINE_SCRIPT_MANAGER_ID` is defined identically here and in `inlineScriptEnvManager.ts`, and the `${PYTHON_EXTENSION_ID}:conda` id appears both in `BASE_INTERPRETER_MANAGER_IDS` and inline in `selectBaseInterpreter`'s filter (Architect, verified in diff). Single-source-of-truth violation with a 2–3 site blast radius on any id change — hoist these to a shared `constants.ts` export.

[verified]

src/common/inlineScriptCacheLayout.ts
```diff
@@ -5,16 +5,14 @@ import * as crypto from 'crypto';
 import * as fsapi from 'fs-extra';
 import * as path from 'path';
 import { Uri } from 'vscode';
+import type { PythonEnvironment } from '../api';
+import { PYTHON_EXTENSION_ID } from './constants';
 import { traceWarn } from './logging';
```
StellaHuang95, line 10: Copilot generated:
-16
The `.meta.json` shape changed (dropped `scriptFsPath`/`requiresPython`, added `baseInterpreterPath`/`baseInterpreterVersion`) but `META_SCHEMA_VERSION` stays `1`, the dir stays `script-envs-v1`, and the shortened doc comment removed the explicit "bump `META_SCHEMA_VERSION` together" guardrail (Advocate + Architect). This is safe only because the feature is behind an undeclared internal flag and v1 was never shipped, and `validateMeta` now rejects the old shape (test `rejects the provisional pre-writer v1 sidecar shape` confirms). Restore the "bump version + dir name together" comment as a future guardrail before GA.

[verified]

src/common/inlineScriptCacheLayout.ts
StellaHuang95: Addressed across the corrected stack: CONDA_MANAGER_ID and INLINE_SCRIPT_MANAGER_ID now live in common/constants.ts, and PR 7 imports the shared inline-script ID instead of declaring another copy.

src/common/inlineScriptCacheLayout.ts
StellaHuang95: Addressed in the corrected cache-utils parent commit. The guardrail now explicitly says to bump the script-envs-vN bucket and META_SCHEMA_VERSION together for incompatible formats; neither value was bumped because this provisional schema was never shipped.

PR #1655, opened 2026-07-23
Add inline-script cache and interpreter utilities (PEP 723 PR 5b/16)

StellaHuang95:

> **Split for review (3 PRs).** Reviewers flagged the original PR 5 as too large, so it is split into three PRs grouped by dependency layer:
> - **5a — generic env-creation utilities — #1651.** Merged.
> - **5b — inline-script cache + interpreter utilities — this PR (#1655).** Rebased on `main`.
> - **5c — `create()` happy path (manager + wiring) — #1656.** Stacked on 5b.
>
> #1651 has merged and this branch has been rebased. The diff now contains only this PR's seven files. **Remaining merge order: 5b → 5c.**

### Roadmap context

This is the second slice of **PR 5 of 16**. See #1651 for the full roadmap table.

|---|---|---|
| | **PR 5b: inline-script cache + interpreter utilities** | **this PR (#1655)** |
| | PR 5c: `create()` happy path (manager + wiring) | #1656 |

### Why this PR

With the generic primitives from 5a in place, this PR lands the **inline-script-specific utilities** that `create()` (5c) composes: a normalized dependency cache key, cache-layout ownership/status checks, and interpreter-constraint handling. These are pure functions with no manager wiring yet, so they are reviewed on their own.

### What this PR adds

**Cache-key tail normalization** (`src/common/inlineScriptCacheKey.ts`): adds `normalizeRequirementTail`, a quote-aware scanner that collapses whitespace and tightens comparator spacing (`>= 1.0` → `>=1.0`) in a requirement's version/marker tail while **preserving quoted PEP 508 marker literals verbatim** (e.g. `python_version >= "3.11"`). Direct-reference requirements (`pkg @ https://…`) are kept verbatim after the name and extras. The effect is that semantically identical dependency strings normalize to the same cache key, so they reuse the same cached environment.

**Cache-layout additions** (`src/common/inlineScriptCacheLayout.ts`): `resolveCacheEntryPath` (containment under the cache root), `inspectOwnedCacheEntry` (realpath ownership), `getBaseInterpreterStatus` (`available | missing | unavailable`), `inspectMetaJson` (typed sidecar read), and a stricter `validateMeta`. The `.meta.json` sidecar schema is `{ schemaVersion, baseInterpreterPath, baseInterpreterVersion, lastUsedAt }`. Uses `getVenvPythonPath` from merged PR #1651.

**Interpreter-constraint trimming** (`src/common/inlineScriptInterpreter.ts`): `pickCompatibleInterpreter` now trims `requires-python`, so a whitespace-only constraint is treated as no constraint.

**Manager ID constants** (`src/common/constants.ts`): centralizes the conda and inline-script manager IDs used by interpreter filtering.

### Tests

- **`inlineScriptCacheKey.unit.test.ts`** — canonicalization cases including marker literals and direct references.
- **`inlineScriptCacheLayout.unit.test.ts`** — the new containment, ownership, base-interpreter-status, and typed sidecar-read helpers.
- **`inlineScriptInterpreter.unit.test.ts`** — constraint trimming / selection.

On this rebased branch `npm run compile-tests` is clean and `npm run unittest` reports **1467 passing, 0 failing, 5 pending**.

### User impact

**None.** These are pure utilities. Nothing calls the new code paths until the manager lands in 5c (#1656).

### Merge order

#1651 has merged. Merge this PR next, then #1656.

### DIFF
--- a/src/common/inlineScriptCacheKey.ts
+++ b/src/common/inlineScriptCacheKey.ts
@@ -31,6 +31,40 @@ function normalizeExtras(inner: string): string {
     return deduped.length > 0 ? `[${deduped.join(',')}]` : '';
 }
 
+function normalizeRequirementTail(value: string): string {
+    let result = '';
+    let unquoted = '';
+    let quote: "'" | '"' | undefined;
+    let escaped = false;
+
+    const flushUnquoted = () => {
+        result += unquoted.replace(/\s+/g, ' ').replace(/\s*([<>=!~]=?)\s*/g, '$1');
+        unquoted = '';
+    };
+
+    // Preserve PEP 508 marker literals verbatim; a backslash escapes the next character.
+    for (const character of value) {
+        if (quote) {
+            result += character;
+            if (character === quote && !escaped) {
+                quote = undefined;
+            }
+            escaped = character === '\\' && !escaped;
+            if (character !== '\\') {
+                escaped = false;
+            }
+        } else if (character === "'" || character === '"') {
+            flushUnquoted();
+            quote = character;
+            result += character;
+        } else {
+            unquoted += character;
+        }
+    }
+    flushUnquoted();
+    return result.trim();
+}
+
 /**
  * Canonicalize a PEP 723 dependency entry so common variants of the same
  * requirement produce identical strings. Not a full PEP 508 parser — only
@@ -70,10 +104,12 @@ export function normalizeDependency(dep: string): string {
         rest = rest.slice(extrasMatch[0].length);
     }
 
-    const compactedRest = rest
-        .replace(/\s+/g, ' ')
-        .replace(/\s*([<>=!~]=?)\s*/g, '$1')
-        .trim();
+    const directReference = rest.trim();
+    if (directReference.startsWith('@')) {
+        return `${name}${extras} ${directReference}`;
+    }
+
+    const compactedRest = normalizeRequirementTail(rest);
 
     return `${name}${extras}${compactedRest}`;
 }
--- a/src/common/inlineScriptCacheKey.ts
+++ b/src/common/inlineScriptCacheKey.ts
@@ -31,6 +31,7 @@ function normalizeExtras(inner: string): string {
     return deduped.length > 0 ? `[${deduped.join(',')}]` : '';
 }
 
+/** ` >=  2 ; python_version  <  "3.13"` becomes `>=2 ; python_version<"3.13"`. */
 function normalizeRequirementTail(value: string): string {
     let result = '';
     let unquoted = '';
--- a/src/common/inlineScriptCacheLayout.ts
+++ b/src/common/inlineScriptCacheLayout.ts
@@ -5,16 +5,14 @@ import * as crypto from 'crypto';
 import * as fsapi from 'fs-extra';
 import * as path from 'path';
 import { Uri } from 'vscode';
+import type { PythonEnvironment } from '../api';
+import { INLINE_SCRIPT_MANAGER_ID } from './constants';
 import { traceWarn } from './logging';
+import { normalizePath } from './utils/pathUtils';
 import { isWindows } from './utils/platformUtils';
+import { getVenvPythonPath } from './utils/virtualEnvironment';
 
-/**
- * Versioned name of the cache root under the extension's `globalStorageUri`.
- *
- * Bump the `-v1` suffix together with {@link META_SCHEMA_VERSION} on any
- * incompatible on-disk change, so old envs sit unread and TTL out naturally
- * instead of being migrated in place.
- */
+/** Bump this and {@link META_SCHEMA_VERSION} together for incompatible cache formats. */
 export const INLINE_SCRIPT_CACHE_DIR_NAME = 'script-envs-v1';
 
 export const META_JSON_FILENAME = '.meta.json';
@@ -33,13 +31,20 @@ const MAX_META_JSON_BYTES = 1024 * 1024;
 export interface InlineScriptEnvMeta {
     /** Version of the serialized metadata schema. */
     readonly schemaVersion: typeof META_SCHEMA_VERSION;
-    /** Filesystem path of the script recorded for lifecycle bookkeeping. */
-    readonly scriptFsPath: string;
+    /** Canonical base-interpreter path. */
+    readonly baseInterpreterPath: string;
+    /** Base-interpreter version. */
+    readonly baseInterpreterVersion: string;
     /** Last successful use as a canonical UTC string produced by `Date.toISOString()`. */
     readonly lastUsedAt: string;
-    /** The script's `requires-python` declaration, when present. */
-    readonly requiresPython?: string;
 }
+
+export type InlineScriptMetaReadResult =
+    | { readonly kind: 'valid'; readonly metadata: InlineScriptEnvMeta }
+    | { readonly kind: 'missing' | 'invalid' | 'unavailable' };
+
+export type BaseInterpreterStatus = 'available' | 'missing' | 'unavailable';
+export type CacheEnvironmentInspection = 'expected' | 'stale' | 'uncertain';
 
 /**
  * In-memory summary of one cached entry, populated by the separate disk walk.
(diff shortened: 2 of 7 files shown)


src/common/inlineScriptCacheKey.ts
```diff
@@ -31,6 +31,40 @@ function normalizeExtras(inner: string): string {
...
 }
 
+function normalizeRequirementTail(value: string): string {
+    let result = '';
+    let unquoted = '';
+    let quote: "'" | '"' | undefined;
+    let escaped = false;
+
+    const flushUnquoted = () => {
+        result += unquoted.replace(/\s+/g, ' ').replace(/\s*([<>=!~]=?)\s*/g, '$1');
```
edvilme, line 42: Maybe add an example/explanation for the regex?

src/common/inlineScriptCacheKey.ts
```diff
@@ -31,6 +31,40 @@ function normalizeExtras(inner: string): string {
     return deduped.length > 0 ? `[${deduped.join(',')}]` : '';
 }
 
+function normalizeRequirementTail(value: string): string {
```
StellaHuang95, line 35: This is added to handle requirement tail, for example:
`requests[security]   >=  2.31 ; python_version  <  "3.13"`

edvilme: Is this for normalizing and then hashing?

src/common/inlineScriptInterpreter.ts
```diff
@@ -24,7 +24,8 @@ export function pickCompatibleInterpreter(
     installed: ReadonlyArray<PythonEnvironment>,
     requiresPython: string | undefined,
 ): PythonEnvironment | undefined {
-    const constraint = requiresPython && requiresPython.length > 0 ? requiresPython : undefined;
+    const trimmedConstraint = requiresPython?.trim();
+    const constraint = trimmedConstraint ? trimmedConstraint : undefined;
```
edvilme, line 28:
```suggestion
    const constraint = requiresPython?.trim();
```
Would return undefined anyways

src/common/inlineScriptCacheLayout.ts
```diff
@@ -232,21 +292,30 @@ function validateMeta(value: unknown): InlineScriptEnvMeta | undefined {
...
+        typeof obj.baseInterpreterVersion !== 'string' ||
+        obj.baseInterpreterVersion.trim().length === 0 ||
+        obj.baseInterpreterVersion.trim() !== obj.baseInterpreterVersion
+    ) {
         return undefined;
     }
-    if (obj.requiresPython !== undefined && typeof obj.requiresPython !== 'string') {
+    if (!isCanonicalIsoTimestamp(obj.lastUsedAt)) {
         return undefined;
     }
```
edvilme, line 307: Maybe this could be extracted into a single method, and have just one comparison return undefined?

src/common/inlineScriptCacheLayout.ts
```diff
@@ -160,53 +206,60 @@ export function selectStaleEntries(entries: ReadonlyArray<CacheEntrySummary>, no
...
  */
 export async function verifyBaseInterpreterExists(envDir: Uri): Promise<boolean> {
-    return isWindows() ? verifyWindowsBaseInterpreter(envDir) : verifyPosixBaseInterpreter(envDir);
+    return (await getBaseInterpreterStatus(envDir)) === 'available';
+}
+
+/** Classify the base interpreter; `unavailable` denotes transient I/O. */
+export async function getBaseInterpreterStatus(envDir: Uri): Promise<BaseInterpreterStatus> {
+    return isWindows() ? getWindowsBaseInterpreterStatus(envDir) : getPosixBaseInterpreterStatus(envDir);
 }
```
edvilme, line 215: Is this used anywhere else? Maybe we can call it directly to avoid the thin wrapper

src/common/inlineScriptCacheLayout.ts
```diff
@@ -33,14 +31,21 @@ const MAX_META_JSON_BYTES = 1024 * 1024;
...
-    /** The script's `requires-python` declaration, when present. */
-    readonly requiresPython?: string;
 }
 
+export type InlineScriptMetaReadResult =
+    | { readonly kind: 'valid'; readonly metadata: InlineScriptEnvMeta }
+    | { readonly kind: 'missing' | 'invalid' | 'unavailable' };
+
+export type BaseInterpreterStatus = 'available' | 'missing' | 'unavailable';
+export type CacheEnvironmentInspection = 'expected' | 'stale' | 'uncertain';
```
edvilme, line 47: Can this be enums?

src/common/inlineScriptCacheKey.ts
StellaHuang95: yes, it removes formatting differences before computing the cache key.

src/common/inlineScriptCacheKey.ts
StellaHuang95: Sure will add that, it's mainly collapsing whitespaces and removes whitespaces around comparison operators

src/common/inlineScriptInterpreter.ts
StellaHuang95: No it doesn't, if it's 
```
# /// script
# requires-python = "   "
# dependencies = ["requests"]
# ///
```
it actually trims the spaces and return `''`, in which case, the downstream rejects any installed Python version. For undefined, which is when the script doesn't have `requires-python` specified, it returns true for installed Python. It's more like a design decision question, maybe I should be less strict here and allow the first case to specify base interpreter too.

src/common/inlineScriptCacheLayout.ts
StellaHuang95: Yes that could be a smaller helper to reduce duplication

src/common/inlineScriptCacheLayout.ts
StellaHuang95: Yes, it's used in test files currently and will be consumed in follow up PR

src/common/inlineScriptCacheLayout.ts
StellaHuang95: What's the reason to have this in enums? They behavior the same so I'd rather use types to be shorter and straightforward, and just following the same pattern used in other places in the repo for small finite status.

eleanorjboyd:
## Test verification report

### Must fix

1. **Ownership checks are not independently protected.** In [`inspectOwnedCacheEntry` tests](https://github.com/microsoft/vscode-python-environments/blob/a98295de45e77125e03010abf2cfdf423b2b0f4b/src/test/common/inlineScriptCacheLayout.unit.test.ts#L510), the stale case changes both `sysPrefix` and `environmentPath`. Mutations removing either production comparison still left all 4 ownership tests passing. Please add separate stale tests where only `sysPrefix` differs and where only `environmentPath` differs.

2. **Direct-child enforcement is untested.** The [`resolveCacheEntryPath` tests](https://github.com/microsoft/vscode-python-environments/blob/a98295de45e77125e03010abf2cfdf423b2b0f4b/src/test/common/inlineScriptCacheLayout.unit.test.ts#L459) cover a direct entry, the root, an outside path, and a missing path—but not a nested descendant. Removing the exact direct-child comparison while retaining descendant containment left all 3 tests passing. Please add a nested path such as `cacheRoot/nested/key` and assert it is rejected.

### Summary
- **Suites reviewed:** 3
- **Added tests reviewed:** 27
- **Critical gaps:** 2
- **Always-passing assertions / swallowed failures:** 0
- **Flakiness or over-mocking findings:** 0
- **Grade:** B

### Verification evidence
- Full baseline: **1,466 passing, 6 pending**.
- Affected suites: **136 passing**.
- Trace logging confirmed the ownership and containment tests execute their real production functions.
- Mutation checks exposed the two gaps above.
- All temporary traces and mutations were reverted; the worktree is clean.

src/common/inlineScriptCacheKey.ts
```diff
@@ -31,6 +31,41 @@ function normalizeExtras(inner: string): string {
     return deduped.length > 0 ? `[${deduped.join(',')}]` : '';
 }
 
+/** ` >=  2 ; python_version  <  "3.13"` becomes `>=2 ; python_version<"3.13"`. */
+function normalizeRequirementTail(value: string): string {
```
edvilme, line 35: I am not entirely sure, but I think this could be further simplified with some regexes rather than iterating over the characters. If not, this looks good to me :)

src/common/inlineScriptCacheLayout.ts
```diff
@@ -224,6 +277,17 @@ function isFileNotFoundError(err: unknown): boolean {
     return typeof err === 'object' && err !== null && 'code' in err && (err as NodeJS.ErrnoException).code === 'ENOENT';
 }
 
+function isDescendantPath(rootPath: string, candidatePath: string): boolean {
+    const relative = path.relative(rootPath, candidatePath);
+    return (
+        relative.length > 0 && relative !== '..' && !relative.startsWith(`..${path.sep}`) && !path.isAbsolute(relative)
+    );
+}
```
edvilme, line 285: Perhaps we could see if there are other places we could use this in, and put it on an utils file or something similar?

edvilme (commented):
2 comments/questions but otherwise, looks good to me ;)

edvilme (approved):

eleanorjboyd (approved):

src/common/inlineScriptCacheKey.ts
StellaHuang95: Looking to address the feedback while getting 5c pr ready. I considered this, but the loop is intended to be quotes aware. It normalizes whitespace outside marker literals while preserving it inside quoted values, and include quotes. It's difficult to get regex to do that. I'd keep it this way for now.

src/common/inlineScriptCacheLayout.ts
StellaHuang95: venv and pip utils have something similar, but for different purposes and different behavior. This one rejects the root itself on purpose because we don't want to accidentally touch the root of the cached envs and modify every cached envs.

PR #1656, opened 2026-07-23
Add PEP 723 inline script environment creation (PEP 723 PR 5c/16)

StellaHuang95:

> **Split for review (3 PRs).** Reviewers flagged the original PR 5 as too large, so it is split into three PRs grouped by dependency layer:
> - **5a — generic env-creation utilities — #1651.** Merged.
> - **5b — inline-script cache + interpreter utilities — #1655.** Merged.
> - **5c — `create()` happy path (manager + wiring) — this PR (#1656).** Rebased on `main`.
>
> #1651 and #1655 have merged and this branch has been rebased. The diff now contains only this PR's five files.

### Roadmap context

This is the final slice of **PR 5 of 16** — the actual `create()` happy path. See #1651 for the full roadmap table.

|---|---|---|
| | **PR 5c: `create()` happy path (manager + wiring)** | **this PR (#1656)** |

### Why this PR

`InlineScriptEnvManager.create()` was a deliberately empty no-op after PR 4. This PR implements its **happy path**: the case where the machine already has a base interpreter that satisfies the script's `requires-python`, so no uv Python install is required. Given a PEP 723 script, it builds — or reuses — a dependency-keyed virtual environment under the extension's global storage, following the pipx-style cache design from Q4 of #1601. The uv-install fallback (no compatible interpreter present) is deferred to PR 6.

It composes the primitives from 5a (#1651) and the inline-script utilities from 5b (#1655); this PR adds only the manager and its wiring.

### What this PR does

**Wires the manager's collaborators** (`extension.ts`, `inlineScriptMain.ts`): `registerInlineScriptFeatures` and the `InlineScriptEnvManager` constructor now receive the `NativePythonFinder`, the `PythonEnvironmentApi`, the base (system) environment manager, and `globalStorageUri`.

**Implements `create(scope)`** (`inlineScriptEnvManager.ts`):

- Accepts exactly one local `file:` URI (a bare `Uri` or single-element array). Anything else — `'global'`, a folder, or multiple URIs — logs a warning and returns `undefined`.
- Reads PEP 723 metadata from the script; missing or invalid metadata returns `undefined`.
- Merges `metadata.dependencies` with `options.additionalPackages`, trims each, and rejects empty entries.
- Selects a base interpreter, computes the dependency + interpreter cache key, and de-duplicates concurrent in-process `create()` calls for the same key via a `pendingCreations` map.

**Base-interpreter selection** (`selectBaseInterpreter`): starts from `getEnvironments('global')`, keeps only true base managers (system, pyenv, conda `base`), and excludes derived environments by rejecting a non-absolute `sysPrefix` or the presence of `pyvenv.cfg`. It then picks the newest compatible interpreter with `pickCompatibleInterpreter` and resolves the executable through `fs.realpath` so the cache key is canonical. If a candidate cannot be resolved it falls through to the next.

**Create-or-reuse under a cross-process lock** (`createOrReuseEnvironment`): acquires a directory lock (5a), inspects the existing cache entry, and reuses / rebuilds / preserves accordingly, always releasing the lock in `finally`.

**Fail-closed cache inspection** (`inspectCacheEntry`) returns `absent | stale | uncertain | reusable`:

- Rejects symlinks and non-directories; verifies the entry is contained under the cache root (`resolveCacheEntryPath`).
- Reads and validates the `.meta.json` sidecar and confirms the recorded base-interpreter path and version still match the selected base.
- Confirms the base interpreter is still present on disk (`getBaseInterpreterStatus`).
- Resolves the cached venv to a real `PythonEnvironment`, confirms it is genuinely ours via realpath containment (`inspectOwnedCacheEntry`), compares Python release segments, and re-checks `requires-python` with the existing `matchesPythonVersion`.
- Only conclusive evidence marks an entry `stale` (rebuild); any doubt yields `uncertain`, and an uncertain entry is **preserved, never deleted**. A reused entry has its `lastUsedAt` refreshed.

**Environment build** (`buildCacheEntry`): delegates to the existing `createWithProgress` venv flow with `trackUvEnvironment` set to `false` so cached script environments are not registered as workspace venvs. On success it writes the sidecar and re-validates that the built environment matches the requested release and is owned by this entry. On failure it removes the directory and returns empty. On **cancellation** it retains the lock so a half-built environment is not silently reused later.

### Tests

- **`inlineScriptEnvManager.unit.test.ts`** — 40 tests across scope/metadata validation, base-interpreter selection, cache creation, cache reuse, transaction rollback, and events/disposal.
- **`inlineScriptMain.unit.test.ts`** — updated for the new `registerInlineScriptFeatures` signature.

On this rebased branch `npm run compile-tests` is clean and `npm run unittest` reports **1491 passing, 0 failing, 5 pending**.

### User impact

**None on the default path.** The manager is still registered only when the undeclared `python-envs.inlineScripts.enabled` flag is on, so default users see nothing.

`create()` is now a declared method (PR 4 omitted it), so with the flag on the inline manager can appear as a create target. But it acts only on a single local script URI and no-ops on every other scope, and **nothing in the extension routes a script URI to it yet**. Wiring the trigger is later work: routing in PR 9, and the "Set up env for this script" picker item and bulk command in PR 11/12.

### Merge order

#1651 and #1655 have merged. This PR is the remaining final slice.

### DIFF
--- a/src/managers/builtin/inlineScriptEnvManager.ts
+++ b/src/managers/builtin/inlineScriptEnvManager.ts
@@ -1,8 +1,12 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-import { Disposable, Event, EventEmitter, l10n, LogOutputChannel, MarkdownString, ThemeIcon } from 'vscode';
+import * as fs from 'fs-extra';
+import * as path from 'path';
+import { Disposable, Event, EventEmitter, l10n, LogOutputChannel, MarkdownString, ThemeIcon, Uri } from 'vscode';
 import {
+    CreateEnvironmentOptions,
+    CreateEnvironmentScope,
     DidChangeEnvironmentEventArgs,
     DidChangeEnvironmentsEventArgs,
     EnvironmentManager,
@@ -10,18 +14,50 @@ import {
     GetEnvironmentsScope,
     IconPath,
     PythonEnvironment,
+    PythonEnvironmentApi,
     RefreshEnvironmentsScope,
     ResolveEnvironmentContext,
     SetEnvironmentScope,
 } from '../../api';
+import { computeCacheKey } from '../../common/inlineScriptCacheKey';
+import {
+    InlineScriptEnvMeta,
+    META_SCHEMA_VERSION,
+    getBaseInterpreterStatus,
+    getScriptEnvCacheRoot,
+    getScriptEnvDir,
+    inspectOwnedCacheEntry,
+    inspectMetaJson,
+    resolveCacheEntryPath,
+    writeMetaJson,
+} from '../../common/inlineScriptCacheLayout';
+import { pickCompatibleInterpreter } from '../../common/inlineScriptInterpreter';
+import {
+    InlineScriptMetadata,
+    matchesPythonVersion,
+    readInlineScriptMetadataFromFile,
+} from '../../common/inlineScriptMetadata';
+import { CONDA_MANAGER_ID, PYTHON_EXTENSION_ID, SYSTEM_MANAGER_ID } from '../../common/constants';
+import { acquireFileLock, AcquiredFileLock } from '../../common/lockfile.apis';
+import { normalizePath } from '../../common/utils/pathUtils';
+import { compareReleaseSegments, parseReleaseSegments } from '../../common/utils/pep440Release';
+import { getVenvPythonPath } from '../../common/utils/virtualEnvironment';
+import { NativePythonFinder } from '../common/nativePythonFinder';
+import { createWithProgress, resolveVenvPythonEnvironmentPath } from './venvUtils';
 
-/**
- * Skeleton EnvironmentManager for PEP 723 inline-script envs. Every
- * method returns the empty / undefined / no-op equivalent; `create`,
- * `remove`, and `quickCreateConfig` are intentionally omitted so the
- * picker UI hides their entry points until later PRs land them.
- */
+const BASE_INTERPRETER_MANAGER_IDS = new Set([
+    SYSTEM_MANAGER_ID,
+    CONDA_MANAGER_ID,
+    `${PYTHON_EXTENSION_ID}:pyenv`,
+]);
+
+const CACHE_LOCK_TIMEOUT_MS = 5 * 60 * 1000;
+const CACHE_LOCK_RETRY_MS = 500;
+
+/** Manages extension-owned PEP 723 script environments. */
 export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
+    private readonly pendingCreations = new Map<string, Promise<PythonEnvironment | undefined>>();
+
     private readonly _onDidChangeEnvironments = new EventEmitter<DidChangeEnvironmentsEventArgs>();
     public readonly onDidChangeEnvironments: Event<DidChangeEnvironmentsEventArgs> =
         this._onDidChangeEnvironments.event;
@@ -39,7 +75,67 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
     );
     public readonly iconPath: IconPath = new ThemeIcon('file-code');
 
-    constructor(public readonly log: LogOutputChannel) {}
+    constructor(
+        private readonly nativeFinder: NativePythonFinder,
+        private readonly api: PythonEnvironmentApi,
+        private readonly baseManager: EnvironmentManager,
+        private readonly globalStorageUri: Uri,
+        public readonly log: LogOutputChannel,
+    ) {}
+
+    async create(
+        scope: CreateEnvironmentScope,
+        options?: CreateEnvironmentOptions,
+    ): Promise<PythonEnvironment | undefined> {
+        try {
+            const scriptUri = this.getScriptUri(scope);
+            if (!scriptUri) {
+                this.log.warn('Inline-script environment creation requires exactly one local file URI.');
+                return undefined;
+            }
+
+            const metadata = await readInlineScriptMetadataFromFile(scriptUri);
+            if (!metadata) {
+                this.log.warn(`No valid PEP 723 metadata found in ${scriptUri.fsPath}.`);
+                return undefined;
+            }
+
+            const packages = [...(metadata.dependencies ?? []), ...(options?.additionalPackages ?? [])].map((value) =>
+                value.trim(),
+            );
+            if (packages.some((value) => value.length === 0)) {
+                this.log.warn(`Inline-script dependencies must not contain empty entries: ${scriptUri.fsPath}.`);
+                return undefined;
+            }
+
+            const selectedBase = await this.selectBaseInterpreter(metadata);
+            if (!selectedBase) {
+                this.log.warn(`No installed Python satisfies the inline-script requirements for ${scriptUri.fsPath}.`);
+                return undefined;
+            }
+            const cacheKey = computeCacheKey({
+                dependencies: packages,
+                interpreterPath: selectedBase.canonicalPath,
+            });
+            const pending = this.pendingCreations.get(cacheKey);
+            if (pending) {
+                return await pending;
+            }
+
+            const creation = this.createOrReuseEnvironment(cacheKey, packages, metadata, selectedBase);
+            this.pendingCreations.set(cacheKey, creation);
+            try {
+                return await creation;
+            } finally {
+                if (this.pendingCreations.get(cacheKey) === creation) {
+                    this.pendingCreations.delete(cacheKey);
+                }
+            }
+        } catch (error) {
+            this.log.error(`Failed to set up inline-script environment: ${this.errorMessage(error)}`);
+            return undefined;
+        }
+    }
 
     async refresh(_scope: RefreshEnvironmentsScope): Promise<void> {
         return;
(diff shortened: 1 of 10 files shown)


eleanorjboyd:
## Test verification report

**CI clarification:** all checks are green for this commit. The TypeScript unit-test job runs only on Ubuntu and Windows; macOS CI runs smoke/E2E/integration suites, not these unit tests.

### Must fix

1. **No test protects an unresolvable cached environment.** Production deliberately preserves an entry when `resolveVenvPythonEnvironmentPath()` returns `undefined` in [`inspectCacheEntry()`](https://github.com/microsoft/vscode-python-environments/blob/95cb552e628eb250d7f179b396192ea201de9b47/src/managers/builtin/inlineScriptEnvManager.ts#L325). Mutating this result from `uncertain` to `stale` left all 41 manager tests passing. Please add a valid-cache test with `resolveVenvStub.resolves(undefined)` and assert the entry is preserved and not rebuilt.

### Warnings

2. **Registration does not verify that the manager itself is disposable.** Replacing the manager in `disposables` with a duplicate registration disposable still left both [`registerInlineScriptFeatures` tests](https://github.com/microsoft/vscode-python-environments/blob/95cb552e628eb250d7f179b396192ea201de9b47/src/test/managers/builtin/inlineScriptMain.unit.test.ts#L50) passing. Assert that the captured manager is actually present in `disposables`, not only that its length is 2.

### Summary
- **New tests reviewed:** 43
- **CI:** all checks passing
- **Local macOS unit run:** 38 passing, 5 failing due to `/var` vs `/private/var` path identity
- **Critical gaps:** 1
- **Warnings:** 2
- **Grade:** B

Trace checks confirmed the coalescing, cache-reuse, and rollback tests execute real manager paths. Mutations showed coalescing and sidecar rollback tests correctly catch regressions. All temporary repairs, traces, and mutations were reverted; the worktree is clean.

StellaHuang95:
Will do a separate PR to put all the inline related files into a "inline" directory. Don't want to make this PR too messy and difficult to review.

StellaHuang95:
> ## Test verification report
> **CI clarification:** all checks are green for this commit. The TypeScript unit-test job runs only on Ubuntu and Windows; macOS CI runs smoke/E2E/integration suites, not these unit tests.
> 
> ### Must fix
> 1. **No test protects an unresolvable cached environment.** Production deliberately preserves an entry when `resolveVenvPythonEnvironmentPath()` returns `undefined` in [`inspectCacheEntry()`](https://github.com/microsoft/vscode-python-environments/blob/95cb552e628eb250d7f179b396192ea201de9b47/src/managers/builtin/inlineScriptEnvManager.ts#L325). Mutating this result from `uncertain` to `stale` left all 41 manager tests passing. Please add a valid-cache test with `resolveVenvStub.resolves(undefined)` and assert the entry is preserved and not rebuilt.
> 
> ### Warnings
> 2. **Registration does not verify that the manager itself is disposable.** Replacing the manager in `disposables` with a duplicate registration disposable still left both [`registerInlineScriptFeatures` tests](https://github.com/microsoft/vscode-python-environments/blob/95cb552e628eb250d7f179b396192ea201de9b47/src/test/managers/builtin/inlineScriptMain.unit.test.ts#L50) passing. Assert that the captured manager is actually present in `disposables`, not only that its length is 2.
> 
> ### Summary
> * **New tests reviewed:** 43
> * **CI:** all checks passing
> * **Local macOS unit run:** 38 passing, 5 failing due to `/var` vs `/private/var` path identity
> * **Critical gaps:** 1
> * **Warnings:** 2
> * **Grade:** B
> 
> Trace checks confirmed the coalescing, cache-reuse, and rollback tests execute real manager paths. Mutations showed coalescing and sidecar rollback tests correctly catch regressions. All temporary repairs, traces, and mutations were reverted; the worktree is clean.

feedback addressed.

eleanorjboyd:
### Low — Consolidate the duplicated file-not-found helper

[`InlineScriptEnvManager.isFileNotFoundError()`](https://github.com/microsoft/vscode-python-environments/blob/63838ca34f1f5bde05b662f9baa2fdec35cb878a/src/managers/builtin/inlineScriptEnvManager.ts#L423-L430) duplicates the helper already used throughout [`inlineScriptCacheLayout.ts`](https://github.com/microsoft/vscode-python-environments/blob/63838ca34f1f5bde05b662f9baa2fdec35cb878a/src/common/inlineScriptCacheLayout.ts#L276-L278). Keeping identical errno classification in two places risks future semantic drift. Move it to an accessible common filesystem utility and reuse it from both modules.

No Critical, High, or Medium findings. Lock retention, rollback, cache ownership, symlink containment, interpreter selection, cross-platform paths, constants placement, and comments all look sound.

src/managers/builtin/inlineScriptEnvManager.ts
```diff
@@ -1,27 +1,63 @@
...
+import { normalizePath } from '../../common/utils/pathUtils';
+import { compareReleaseSegments, parseReleaseSegments } from '../../common/utils/pep440Release';
+import { getVenvPythonPath } from '../../common/utils/virtualEnvironment';
+import { NativePythonFinder } from '../common/nativePythonFinder';
+import { createWithProgress, resolveVenvPythonEnvironmentPath } from './venvUtils';
+
+const BASE_INTERPRETER_MANAGER_IDS = new Set([
+    SYSTEM_MANAGER_ID,
+    CONDA_MANAGER_ID,
+    `${PYTHON_EXTENSION_ID}:pyenv`,
```
edvilme: Why is this not a `PYENV_MANAGER_ID`?

src/managers/builtin/inlineScriptEnvManager.ts
```diff
@@ -39,7 +75,67 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
...
+            }
+
+            const metadata = await readInlineScriptMetadataFromFile(scriptUri);
+            if (!metadata) {
+                this.log.warn(`No valid PEP 723 metadata found in ${scriptUri.fsPath}.`);
+                return undefined;
+            }
+
+            const packages = [...(metadata.dependencies ?? []), ...(options?.additionalPackages ?? [])].map((value) =>
+                value.trim(),
```
edvilme:
```suggestion
                value.trim()
```

src/managers/builtin/inlineScriptEnvManager.ts
```diff
@@ -39,7 +75,67 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
...
+                this.log.warn(`Inline-script dependencies must not contain empty entries: ${scriptUri.fsPath}.`);
+                return undefined;
+            }
+
+            const selectedBase = await this.selectBaseInterpreter(metadata);
+            if (!selectedBase) {
+                this.log.warn(`No installed Python satisfies the inline-script requirements for ${scriptUri.fsPath}.`);
+                return undefined;
+            }
+            const cacheKey = computeCacheKey({
```
edvilme, line 139:
```suggestion
            }
            
            const cacheKey = computeCacheKey({
```

src/managers/builtin/inlineScriptEnvManager.ts
```diff
@@ -39,7 +75,67 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
...
+                dependencies: packages,
+                interpreterPath: selectedBase.canonicalPath,
+            });
+            const pending = this.pendingCreations.get(cacheKey);
+            if (pending) {
+                return await pending;
+            }
+
+            const creation = this.createOrReuseEnvironment(cacheKey, packages, metadata, selectedBase);
```
edvilme: Being picky here, but can we have it receive an object so we have named arguments in this method, and then just call 
```js
            const creation = this.createOrReuseEnvironment({cacheKey, packages, metadata, selectedBase});
```

src/managers/builtin/inlineScriptEnvManager.ts
```diff
@@ -61,8 +157,309 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
         return undefined;
     }
 
+    private getScriptUri(scope: CreateEnvironmentScope): Uri | undefined {
+        const uri = scope instanceof Uri ? scope : Array.isArray(scope) && scope.length === 1 ? scope[0] : undefined;
+        return uri?.scheme === 'file' ? uri : undefined;
+    }
+
+    private async selectBaseInterpreter(metadata: InlineScriptMetadata): Promise<SelectedBaseInterpreter | undefined> {
+        const reported = (await this.api.getEnvironments('global')).filter(
```
edvilme: I would probably extract the inline await into a variable `globalEnvironments` to better catch it if something goes wrong.

src/managers/builtin/inlineScriptEnvManager.ts
```diff
@@ -61,8 +157,309 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
...
+                continue;
+            }
+            try {
+                return { environment, canonicalPath: await fs.realpath(executable) };
+            } catch (error) {
+                this.log.warn(
+                    `Skipping base interpreter that cannot be resolved at ${executable}: ${this.errorMessage(error)}`,
+                );
+            }
+        }
```
edvilme, line 236: Maybe this could be refactored into a for loop that returns when it has found an executable?

src/managers/builtin/inlineScriptEnvManager.ts
```diff
@@ -61,8 +157,309 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
...
+interface BuildCacheEntryResult {
+    readonly environment?: PythonEnvironment;
+    readonly retainLock?: boolean;
+}
+
+type CacheEntryInspection =
+    | { readonly kind: 'absent' | 'stale' | 'uncertain' }
+    | { readonly kind: 'reusable'; readonly environment: PythonEnvironment };
+
+type PythonReleaseComparison = 'same' | 'different' | 'uncertain';
```
edvilme: Can we have the interfaces and types at the top?

src/managers/builtin/inlineScriptEnvManager.ts
```diff
@@ -61,8 +157,309 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
...
+        }
+
+        const sidecar: InlineScriptEnvMeta = {
+            schemaVersion: META_SCHEMA_VERSION,
+            baseInterpreterPath: selectedBase.canonicalPath,
+            baseInterpreterVersion: selectedBase.environment.version,
+            lastUsedAt: new Date().toISOString(),
+        };
+        try {
+            await writeMetaJson(envDir, sidecar);
```
edvilme:
```suggestion
        try {
            await writeMetaJson(envDir, {
                schemaVersion: META_SCHEMA_VERSION,
                baseInterpreterPath: selectedBase.canonicalPath,
                baseInterpreterVersion: selectedBase.environment.version,
                lastUsedAt: new Date().toISOString(),
            });
```
Maybe we can add it inline?

src/managers/builtin/inlineScriptEnvManager.ts
```diff
@@ -61,8 +157,309 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
...
+    private isFileNotFoundError(error: unknown): boolean {
+        return (
+            typeof error === 'object' &&
+            error !== null &&
+            'code' in error &&
+            (error as NodeJS.ErrnoException).code === 'ENOENT'
+        );
+    }
+
+    private comparePythonReleases(actual: string, expected: string): PythonReleaseComparison {
```
edvilme: I think this is mostly used to check if the releases are the same, so this could probably be changed into something like `areEqualPythonReleaseVersions` that returns a bool value?

edvilme (changes requested):
Will take a look at the rest throughout the day, but left some comments in the meantime :)

src/managers/builtin/inlineScriptEnvManager.ts
StellaHuang95: yeah make sense to centralize it, done

src/managers/builtin/inlineScriptEnvManager.ts
StellaHuang95: sure

src/managers/builtin/inlineScriptEnvManager.ts
StellaHuang95: sure

src/managers/builtin/inlineScriptEnvManager.ts
StellaHuang95: I'm personally not a fan of inline style just for easier debugging purposes, but I don't mind the stylistic change

StellaHuang95:
> ### Low — Consolidate the duplicated file-not-found helper
> [`InlineScriptEnvManager.isFileNotFoundError()`](https://github.com/microsoft/vscode-python-environments/blob/63838ca34f1f5bde05b662f9baa2fdec35cb878a/src/managers/builtin/inlineScriptEnvManager.ts#L423-L430) duplicates the helper already used throughout [`inlineScriptCacheLayout.ts`](https://github.com/microsoft/vscode-python-environments/blob/63838ca34f1f5bde05b662f9baa2fdec35cb878a/src/common/inlineScriptCacheLayout.ts#L276-L278). Keeping identical errno classification in two places risks future semantic drift. Move it to an accessible common filesystem utility and reuse it from both modules.
> 
> No Critical, High, or Medium findings. Lock retention, rollback, cache ownership, symlink containment, interpreter selection, cross-platform paths, constants placement, and comments all look sound.

Addressed both of your test verification report and consolidate the helper.

src/managers/builtin/inlineScriptEnvManager.ts
StellaHuang95: The call is self-explanatory and distinct, I prefer using named args for unclear values like false or numbers, I don't mind addressing it, but we may want to come up with some stylistic principles so that it's consistent across the entire repo.

src/managers/builtin/inlineScriptEnvManager.ts
StellaHuang95: sure

src/managers/builtin/inlineScriptEnvManager.ts
StellaHuang95: I’d prefer to keep the while  loop here. `pickCompatibleInterpreter` returns only the single best candidate, not a sorted list. If that candidate’s executable cannot be resolved, we remove it and call the picker again to get the next-best candidate. A `for` loop would require changing the picker API or duplicating its compatibility and sorting logic here. The loop is bounded because it removes one candidate on every iteration.

src/managers/builtin/inlineScriptEnvManager.ts
edvilme: Ahh ok, fair enough ;)

src/managers/builtin/inlineScriptEnvManager.ts
```diff
@@ -61,6 +185,281 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
         return undefined;
     }
 
+    private getScriptUri(scope: CreateEnvironmentScope): Uri | undefined {
+        const uri = scope instanceof Uri ? scope : Array.isArray(scope) && scope.length === 1 ? scope[0] : undefined;
+        return uri?.scheme === 'file' ? uri : undefined;
+    }
```
edvilme, line 191: Maybe this can be a util function as it is used once and not bound to the instance of the class?

eleanorjboyd (approved):

src/managers/builtin/inlineScriptEnvManager.ts
```diff
@@ -61,6 +185,281 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
...
+        const expectedRelease = parseReleaseSegments(expected);
+        if (actualRelease === undefined || expectedRelease === undefined) {
+            return false;
+        }
+        return compareReleaseSegments(actualRelease, expectedRelease) === 0;
+    }
+
+    private errorMessage(error: unknown): string {
+        return error instanceof Error ? error.message : String(error);
+    }
```
edvilme, line 461: Maybe this can also be a helper function outside the class since it is not bound to an instance? And could potentially be used in other places as well?

edvilme (approved):
Couple of comments about maybe extracting some of the helper functions off of the class since they are not bound to a specific instance, but otherwise LGTM :)

src/managers/builtin/inlineScriptEnvManager.ts
StellaHuang95: I considered moving it, but this helper represents the inline manager’s specific scope contract, which is exactly one local file URI. It is not currently shared by any other manager, so I think keeping it private keeps that ownership more clearly than introducing a generic utility.

src/managers/builtin/inlineScriptEnvManager.ts
StellaHuang95: Sure I can move to common/errors/utils.ts

PR #1663, opened 2026-07-25
fix: hide Python activity bar icon in non-Python workspaces

mohityadav8:
Fixes microsoft/vscode-python#26015

Added context key python-envs.workspaceHasPython via findFiles + FileSystemWatcher. ANDed into when clauses of the activitybar container and both views.


src/features/views/workspacePythonContext.ts
```diff
@@ -0,0 +1,24 @@
+import { Disposable } from 'vscode';
+import { executeCommand } from '../../common/command.api';
+import { createFileSystemWatcher, findFiles, onDidChangeWorkspaceFolders } from '../../common/workspace.apis';
+
+export const PYTHON_WORKSPACE_KEY = 'python-envs.workspaceHasPython';
+
+const MARKER_GLOB = '**/{*.py,pyproject.toml,setup.py,requirements.txt,Pipfile,manage.py,app.py,.venv,.conda,mspythonconfig.json}';
```
edvilme, line 7: I think app.py, manage.py, setup.py and *.py are redundant here

mohityadav8: @edvilme done 👍

edvilme (dismissed):
Thanks for your contributions! 
LGTML :)

mohityadav8:
hi @edvilme  Both CI failures are unrelated to this PR's changes:

**1. `Check for changed files` (package-lock.json)**
The `package.json` edits in this PR only modify the `contributes` section (VS Code metadata-- `viewsContainers` and `views` `when` clauses).No npm dependencies were added or changed, so `package-lock.json` does not need updating  . Could a maintainer add the `skip package*.json` label to pass this check?

**2. Integration test failure (`pythonProjects.integration.test.js:150`)**

The failure is:
AssertionError: Retrieved environment should match set environment

'Python 3.14.6.final.0-j5ae0pzs6ek'
'Python 3.14.6.final.0-m8ni2jdqggb'
This is a pre-existing flaky test comparing non-deterministic environment hash suffixes. It is not triggered by any code in this PR -- the new file (`workspacePythonContext.ts`) and the changes to `extension.ts` and `package.json` have no interaction with environment resolution or the project manager.

package.json
```diff
@@ -680,14 +680,14 @@
                     "name": "Python Projects",
                     "icon": "files/logo.svg",
                     "contextualTitle": "Python Projects",
-                    "when": "config.python.useEnvironmentsExtension != false"
+                    "when": "config.python.useEnvironmentsExtension != false && python-envs.workspaceHasPython"
```
StellaHuang95, line 682: The PR hides the activity-bar icon unless this context key is true:

`python-envs.workspaceHasPython`

That key is not defined initially. It only gets set after the extension’s  `activate()`  function runs:

`registerWorkspacePythonContext(context.subscriptions);`

However,  `package.json` currently activates the extension only when VS Code opens a Python-language document:

`"activationEvents": [
    "onLanguage:python"
]`

This creates a circular dependency:

1. The extension must activate to detect Python files and set the context key.
2. The activity-bar icon is hidden until that key is set.
3. A hidden view cannot be opened to activate the extension.
4. If no  .py  file is opened,  `onLanguage:python` never activates the extension.

Example

A user opens a repository containing:

my-project/
├── pyproject.toml
├── requirements.txt
└── README.md

The repository is clearly a Python project, but the user has not opened a  .py  file yet.

Expected: The Python activity-bar icon appears because pyproject.toml  identifies the workspace as Python.

Actual: The extension does not activate, so it never searches for  pyproject.toml . The context key remains unset, and the icon stays hidden.

src/features/views/workspacePythonContext.ts
```diff
@@ -0,0 +1,24 @@
...
+import { executeCommand } from '../../common/command.api';
+import { createFileSystemWatcher, findFiles, onDidChangeWorkspaceFolders } from '../../common/workspace.apis';
+
+export const PYTHON_WORKSPACE_KEY = 'python-envs.workspaceHasPython';
+
+const MARKER_GLOB = '**/{*.py,pyproject.toml,requirements.txt,Pipfile,.venv,.conda,mspythonconfig.json}';
+const EXCLUDE = '**/{node_modules,.git,site-packages}/**';
+
+async function refresh(): Promise<void> {
+    const hits = await findFiles(MARKER_GLOB, EXCLUDE, 1);
```
StellaHuang95, line 11: Is this only searching inside open workspaces folders. What if user opens a standalone Python file without opening the folder?

src/features/views/workspacePythonContext.ts
```diff
@@ -0,0 +1,24 @@
...
+    await executeCommand('setContext', PYTHON_WORKSPACE_KEY, hits.length > 0);
+}
+
+export function registerWorkspacePythonContext(disposables: Disposable[]): void {
+    const watcher = createFileSystemWatcher(MARKER_GLOB, false, true, false);
+    disposables.push(
```
StellaHuang95, line 17: Is `EXCLUDE` only passed to findFiles() but not the `createFileSystemWatchter`? Does that mean the watcher still listens for every `.py`  file created or deleted under `site-packages`? I am a bit worried about the perf here.

mohityadav8: Correct-- `createFileSystemWatcher` has no exclude parameter in the VS Code API, so it fires for site-packages too. Fix is to filter in the event handlers:

```ts
const EXCLUDE_RE = /[\\/](node_modules|\.git|site-packages)[\\/]/;

watcher.onDidCreate((uri) => { if (!EXCLUDE_RE.test(uri.fsPath)) void refresh(); }),
watcher.onDidDelete((uri) => { if (!EXCLUDE_RE.test(uri.fsPath)) void refresh(); }),
```

This avoids unnecessary `refresh()` calls from excluded directories.

src/features/views/workspacePythonContext.ts
mohityadav8: `findFiles` only searches workspace folders, so standalone file case is missed. Will add a fallback check on open text documents:

```ts
async function refresh(): Promise<void> {
    const hits = await findFiles(MARKER_GLOB, EXCLUDE, 1);
    if (hits.length > 0) {
        await executeCommand('setContext', PYTHON_WORKSPACE_KEY, true);
        return;
    }
    const hasPythonDoc = workspace.textDocuments.some(
        (doc) => doc.languageId === 'python',
    );
    await executeCommand('setContext', PYTHON_WORKSPACE_KEY, hasPythonDoc);
}
```

And subscribe to `onDidOpenTextDocument` in `registerWorkspacePythonContext`.

package.json
mohityadav8: Good catch. Fix is to add `workspaceContains` activation events to `package.json` so the extension activates even before a `.py` file is opened:

"activationEvents": [
    "onLanguage:python",
    "workspaceContains:**/*.py",
    "workspaceContains:pyproject.toml",
    "workspaceContains:requirements.txt",
    "workspaceContains:Pipfile",
    "workspaceContains:setup.py",
    "workspaceContains:mspythonconfig.json",
    "workspaceContains:.venv",
    "workspaceContains:.conda"
]

This breaks the circular dependency - extension activates when any marker file is present in the workspace, sets the context key, and the icon appears without needing a .py file open first.

edvilme:
Hello @mohityadav8 thanks for following on with the reviews. I think there is still one unresolved issue and a merge conflict before we can take another look :)

mohityadav8:
> Hello @mohityadav8 thanks for following on with the reviews. I think there is still one unresolved issue and a merge conflict before we can take another look :)

Solve conflict can u please share what's the issue I forgot to see in workflow

mohityadav8:
@edvilme these two passes are failing in every pr so can you merge this?

edvilme:
Hello @mohityadav8 thanks for taking the time to work on this. I believe there are some linting issues that are causing CI to fail. Please address those and then we can proceed with review. Thanks!

heejaechang:

src/features/views/workspacePythonContext.ts
```diff
@@ -0,0 +1,24 @@
+import { Disposable } from 'vscode';
+import { executeCommand } from '../../common/command.api';
+import { createFileSystemWatcher, findFiles, onDidChangeWorkspaceFolders } from '../../common/workspace.apis';
+
+export const PYTHON_WORKSPACE_KEY = 'python-envs.workspaceHasPython';
+
+const MARKER_GLOB = '**/{*.py,pyproject.toml,setup.py,requirements.txt,Pipfile,manage.py,app.py,.venv,.conda,mspythonconfig.json}';
+const EXCLUDE = '**/{node_modules,.git,site-packages}/**';
```
heejaechang, line 8: > **Warning** · Non-blocking recommendation

The classifier omits legitimate notebook-only and stub-only workspaces (`*.ipynb`, `*.pyi`), while `.venv` and `.conda` may be directories that `findFiles` does not return. Reuse a canonical project classifier if available, or expand and test the taxonomy with reliable environment markers such as `pyvenv.cfg`.

src/features/views/workspacePythonContext.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export function registerWorkspacePythonContext(disposables: Disposable[]): void {
+    const watcher = createFileSystemWatcher(MARKER_GLOB, false, true, false);
+    disposables.push(
+        watcher,
+        watcher.onDidCreate(() => void refresh()),
+        watcher.onDidDelete(() => void refresh()),
+        onDidChangeWorkspaceFolders(() => void refresh()),
+    );
+    void refresh();
```
heejaechang, line 23: > **Warning** · Non-blocking recommendation

Independent refreshes can overlap, allowing an older `findFiles` result to overwrite newer workspace state. Serialize or invalidate superseded scans, and add coverage for searches resolving in reverse order.

heejaechang, line 23: > **Warning** · Non-blocking recommendation

Each `refresh()` promise is discarded, so failures from `findFiles` or `setContext` can become unhandled rejections. Route these calls through the extension's established async error-reporting or logging pattern.

src/extension.ts
```diff
@@ -113,6 +114,7 @@ import { registerPoetryFeatures } from './managers/poetry/main';
 import { registerPyenvFeatures } from './managers/pyenv/main';
 
 export async function activate(context: ExtensionContext): Promise<PythonEnvironmentApi | undefined> {
+    registerWorkspacePythonContext(context.subscriptions);
     // Only skip activation if user explicitly set useEnvironmentsExtension to false.
```
heejaechang, line 118: > **Warning** · Non-blocking recommendation

Registration occurs before the `useEnvironmentsExtension === false` activation exit, so the disabled feature still creates a watcher and initiates discovery. Move registration after the configuration guard; the contribution-level condition already keeps the UI hidden.

src/features/views/workspacePythonContext.ts
heejaechang, line 23: > **Warning** · Non-blocking recommendation

This persistent UI policy has no regression coverage. Add focused tests for initial discovery, marker creation and last-marker deletion, exclusions, workspace-folder changes, directory markers, and overlapping refreshes.

heejaechang (approved):
Approved via Review Center.

src/features/views/workspacePythonContext.ts
```diff
@@ -0,0 +1,24 @@
...
+}
```
heejaechang, line 24: > **Issue** · Please address or respond

This adds asynchronous initial discovery and event-driven context transitions without regression tests. Add coverage for initial state, marker creation, deletion of the final marker, workspace-folder changes, directory markers, and overlapping refreshes.

[verified]

src/features/views/workspacePythonContext.ts
heejaechang, line 23: > **Warning** · Non-blocking recommendation

Could concurrent fire-and-forget refreshes complete out of order and let an older scan overwrite newer workspace state? Serialize refreshes or use a generation token, and cover reverse-order `findFiles` resolution in a test.

src/extension.ts
```diff
@@ -93,7 +93,8 @@ import { ProjectView } from './features/views/projectView';
 import { PythonStatusBarImpl } from './features/views/pythonStatusBar';
 import { updateViewsAndStatus } from './features/views/revealHandler';
 import { TemporaryStateManager } from './features/views/temporaryStateManager';
-import { PythonEnvTreeItem } from './features/views/treeViewItems';
+import { ProjectItem, PythonEnvTreeItem } from './features/views/treeViewItems';
```
heejaechang, line 96: > **Warning** · Non-blocking recommendation

`ProjectItem` is newly imported but is not used by this change. Remove the import to avoid an unused-local lint or compile failure.

heejaechang:

Issue #1622, opened 2026-07-06
Add unit tests for pip and conda package managers


eleanorjboyd:
The `Package` pillar of the public API (`managePackages`, `getPackages`, `refreshPackages`) is backed by `src/managers/builtin/pipPackageManager.ts` and `src/managers/conda/condaPackageManager.ts`, and **neither has a direct unit test**. Only their helper utilities (`pipListUtils`, `pipUtils`, `installArgs`, `normalizePackageName`) and the host-based `packageManagement.integration.test.ts` provide coverage, so the manager classes' own logic (install/uninstall dispatch, change-event firing, error propagation) is untested at the unit level.

@edvilme you pick to decide exactly how this should look. I think unit test plus some sort of integration test would be ideal.

edvilme:
Hello, yes! There are a couple of things I want to get polished on the packaging side of things before doing this to make sure all of the package managers behave similarly and that there is a good way of testing this

edvilme:
Fixed in #1704

PR #1677, opened 2026-07-28
Add package manager command smoke tests

edvilme:
## Summary

Adds focused unit coverage for the concrete package manager command classes introduced in #1621, across Pip/UV, Conda, and Poetry. Also corrects the `UvUninstallCommand` doc comment.

## What the tests cover

- **Read commands** (list, list-direct-names, version, available-versions, Poetry `show`): stub the runner with representative output and assert the **parsed result** — package lists (with rows missing a version dropped), direct names with normalization, `uv pip tree` indentation filtering, version-string parsing, and available-version prerelease filtering.
- **Mutation commands** (install / uninstall / add / remove): grouped as smoke tests that assert `execute()` resolves. They have no output to parse; their concrete command strings are exercised by integration tests and by the adoption PR #1686.

These deliberately avoid asserting the exact argument vector, which would merely restate `buildCommand`.

## Doc fix

- `UvUninstallCommand`: `uv pip uninstall` has no `-y`/`--yes` flag and is non-interactive by default — corrected the misleading comment.

## Relationship

- Depends only on the already-merged command classes (#1621); targets `main` directly.
- Independent of the adoption PR #1686. These two were previously a GitHub stack; the stack has been undone and both PRs now target `main`. Both currently add the command test files, so whichever merges first the other needs a trivial rebase on those files.

## Testing

- `npm run compile-tests`
- `npm run unittest` (58 command tests passing)

### DIFF
--- a/src/test/managers/builtin/commands.unit.test.ts
+++ b/src/test/managers/builtin/commands.unit.test.ts
@@ -1,0 +1,111 @@
+import assert from 'assert';
+import * as sinon from 'sinon';
+import { LogOutputChannel } from 'vscode';
+import * as workspaceApis from '../../../common/workspace.apis';
+import { PipAvailableVersionsCommand, UvAvailableVersionsCommand } from '../../../managers/builtin/commands/availableVersions';
+import { PipInstallCommand, UvInstallCommand } from '../../../managers/builtin/commands/install';
+import { PipListCommand, UvListCommand } from '../../../managers/builtin/commands/list';
+import { PipListDirectNamesCommand, UvListDirectNamesCommand } from '../../../managers/builtin/commands/listDirectNames';
+import { PipUninstallCommand, UvUninstallCommand } from '../../../managers/builtin/commands/uninstall';
+import { PipVersionCommand, UvVersionCommand } from '../../../managers/builtin/commands/version';
+import * as helpers from '../../../managers/builtin/helpers';
+import { createMockLogOutputChannel } from '../../mocks/helper';
+
+suite('Pip and UV commands', () => {
+    let mockLog: LogOutputChannel;
+    let runPythonStub: sinon.SinonStub;
+    let runUvStub: sinon.SinonStub;
+
+    setup(() => {
+        mockLog = createMockLogOutputChannel();
+        sinon.stub(workspaceApis, 'getConfiguration').returns({
+            get: () => undefined,
+        } as unknown as ReturnType<typeof workspaceApis.getConfiguration>);
+        runPythonStub = sinon.stub(helpers, 'runPython').resolves('');
+        runUvStub = sinon.stub(helpers, 'runUV').resolves('');
+    });
+
+    teardown(() => {
+        sinon.restore();
+    });
+
+    test('PipAvailableVersionsCommand executes without error', async () => {
+        runPythonStub.resolves(JSON.stringify({ versions: ['1.0.0'] }));
+        const command = new PipAvailableVersionsCommand({ pythonExecutable: 'python', log: mockLog });
+
+        await assert.doesNotReject(() => command.execute({ packageName: 'package', pythonVersion: '3.13.1' }));
+    });
+
+    test('UvAvailableVersionsCommand executes without error', async () => {
+        runUvStub.resolves(JSON.stringify({ versions: ['1.0.0'] }));
+        const command = new UvAvailableVersionsCommand({ pythonExecutable: 'python', log: mockLog });
+
+        await assert.doesNotReject(() => command.execute({ packageName: 'package', pythonVersion: '3.13.1' }));
+    });
+
+    test('PipInstallCommand executes without error', async () => {
+        const command = new PipInstallCommand({ pythonExecutable: 'python', log: mockLog });
+
+        await assert.doesNotReject(() => command.execute({ packages: [{ packageName: 'package' }] }));
+    });
+
+    test('UvInstallCommand executes without error', async () => {
+        const command = new UvInstallCommand({ pythonExecutable: 'python', log: mockLog });
+
+        await assert.doesNotReject(() => command.execute({ packages: [{ packageName: 'package' }] }));
+    });
+
+    test('PipUninstallCommand executes without error', async () => {
+        const command = new PipUninstallCommand({ pythonExecutable: 'python', log: mockLog });
+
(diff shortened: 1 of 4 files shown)


edvilme:
Contributes to #1622

edvilme:
Note: These tests only verify that the commands are valid, a further PR will add integration tests with the package managers and round-trips such as: fetching available versions for a package, installing it, listing (direct) packages, uninstalling, etc...

StellaHuang95:
## Finding: The tests pass even when the commands do nothing or produce incorrect results

**Severity:** Medium

**Locations:**

- [`src/test/managers/builtin/commands.unit.test.ts#L32-L109`](https://github.com/microsoft/vscode-python-environments/blob/c229085966fa46eedc6557b86a243521ec04ce1c/src/test/managers/builtin/commands.unit.test.ts#L32-L109)
- [`src/test/managers/conda/commands.unit.test.ts#L29-L71`](https://github.com/microsoft/vscode-python-environments/blob/c229085966fa46eedc6557b86a243521ec04ce1c/src/test/managers/conda/commands.unit.test.ts#L29-L71)
- [`src/test/managers/poetry/commands.unit.test.ts#L31-L60`](https://github.com/microsoft/vscode-python-environments/blob/c229085966fa46eedc6557b86a243521ec04ce1c/src/test/managers/poetry/commands.unit.test.ts#L31-L60)

Every test uses this general assertion:

```ts
await assert.doesNotReject(() => command.execute(...));
```

In simple terms, this asks only:

> “Did this function finish without throwing an error?”

It does **not** ask:

- Was the package-manager executable called?
- Was the correct subcommand used?
- Were the correct arguments passed?
- Was the correct environment selected?
- Was cancellation forwarded?
- Was the returned output parsed correctly?
- Did the command return the expected packages or versions?

### Why that matters

For example, this test is named:

```ts
test('PipInstallCommand executes without error', async () => {
    const command = new PipInstallCommand(...);

    await assert.doesNotReject(() =>
        command.execute({ packages: [{ packageName: 'package' }] }),
    );
});
```

The real implementation should invoke something equivalent to:

```text
python -m pip install package
```

But all of the following broken implementations would still pass:

```ts
async execute(): Promise<void> {
    // Do nothing.
}
```

```ts
async execute(): Promise<void> {
    await runPython('python', ['-m', 'pip', 'uninstall', 'package']);
}
```

```ts
async execute(): Promise<void> {
    await runPython('python', ['incorrect-command']);
}
```

Because `runPython` is stubbed to resolve successfully and the test never inspects its arguments, a completely wrong command looks identical to a correct one.

### Concrete regressions these tests would miss

#### Pip and UV

The tests would not detect:

- Pip using `uninstall` instead of `install`.
- Pip omitting `-m pip`.
- UV omitting `--python <interpreter>`.
- UV using the wrong Python environment.
- Upgrade requests omitting `--upgrade`.
- Package versions being dropped.
- Editable-install arguments being incorrectly transformed.
- Uninstall omitting Pip’s noninteractive `-y` option.

For example, `UvInstallCommand` is expected to construct:

```text
uv pip install --python python package
```

The test does not assert any part of that command.

#### Conda

The tests would not detect:

- Installing into the wrong environment.
- Omitting `--prefix environment`.
- Omitting `--yes` and causing an interactive hang.
- Using `install` instead of `remove`.
- Incorrect formatting of versioned packages.
- Upgrade operations omitting `--update-all`.

The environment path is particularly important:

```ts
condaEnvironmentPath: 'environment'
```

The test supplies it but never confirms it reaches the command runner. The implementation could silently install packages into Conda’s currently active environment and the test would still pass.

#### Poetry

The tests would not detect:

- Running Poetry in the wrong working directory.
- Omitting `add`, `remove`, `show`, or `--top-level`.
- Formatting `package@version` incorrectly.
- Dropping the project’s `cwd`.
- Returning the wrong package names.

Poetry operates against a project’s `pyproject.toml`, so the working directory is behaviorally important. This test:

```ts
new PoetryAddCommand({
    pythonExecutable: 'poetry',
    cwd: 'project',
    log: mockLog,
});
```

never verifies that `'project'` is passed to `runPoetry`. If it were omitted, Poetry could modify a different project or fail to find `pyproject.toml`.

### Recommended improvement

At minimum, each mutation command should verify the runner invocation.

For example:

```ts
await command.execute({
    packages: [{ packageName: 'package' }],
});

assert.ok(
    runPythonStub.calledOnceWithExactly(
        'python',
        ['-m', 'pip', 'install', 'package'],
        undefined,
        mockLog,
        undefined,
        sinon.match.number,
    ),
);
```

The exact assertion should match the helper’s contract, but the important point is to verify the executable, arguments, environment/cwd, cancellation token, and timeout.

src/test/managers/builtin/commands.unit.test.ts
```diff
@@ -0,0 +1,137 @@
...
+        runUvStub = sinon.stub(helpers, 'runUV').resolves('');
+    });
+
+    teardown(() => {
+        sinon.restore();
+    });
+
+    // Mutation commands have no output to parse; unit coverage is a smoke test that
+    // execute() resolves. Their concrete command strings are exercised by integration tests.
+    suite('mutation commands execute without error', () => {
```
eleanorjboyd, line 34: This integration-coverage claim is not currently true. `packageManagement.integration.test.ts` only selects `venv` environments for install/uninstall, so it never exercises the Conda or Poetry command classes; that path also dispatches to Pip or UV dynamically, so it does not deterministically cover either concrete command string. Because these unit tests do not inspect the runner invocation either, subcommand, flag, interpreter, environment-prefix, and `cwd` regressions remain uncovered. Please assert the runner arguments in these suites (including the analogous Conda/Poetry tests), or remove this claim until deterministic integration coverage exists.

eleanorjboyd (commented):
The added parsing assertions are a useful improvement. I left one inline comment on the remaining command-construction coverage gap and the inaccurate integration-test claim; this is the main issue I would address before treating these as meaningful command tests.

edvilme:
> The added parsing assertions are a useful improvement. I left one inline comment on the remaining command-construction coverage gap and the inaccurate integration-test claim; this is the main issue I would address before treating these as meaningful command tests.

I think you are right. I will try changing the order of these PRs. I think it would be better to have the proper integration first and then perform actual roundtrip integration tests; maybe even in the same PR.

edvilme:
Superseeded

PR #1682, opened 2026-07-30
feat: add getPackageAvailableVersions to public API and bump API to 1.1.0

edvilme:
## Summary

- add `PythonPackageGetterApi.getPackageAvailableVersions` to the public API so consumers can query a package's available versions
- bump `@vscode/python-environments` from `1.0.0` to `1.1.0`
- synchronize both API lockfile version fields
- document the public API changes since the latest npm release

## API changes

- **add `PythonPackageGetterApi.getPackageAvailableVersions(environment, packageName): Promise<Pep440Version[] | undefined>`** — exposes a package's available versions (newest-first) to API consumers, delegating to the environment's package manager and resolving to `undefined` when version listing is unsupported
- re-export `Pep440Version` for package manager version APIs
- add `PackageInfo.isTransitive` and `GetPackagesOptions.skipCache`
- add optional package manager hooks for watch targets, direct dependencies, tool versions, available package versions, and install-spec formatting
- add the optional `GetPackagesOptions` parameter to `PackageManager.getPackages` and `PythonPackageGetterApi.getPackages`
- update `PackageManager.refresh` and `PythonPackageGetterApi.refreshPackages` to return the refreshed package list when available

## Validation

- `npm run lint`
- `npm run compile-tests`
- `npm run unittest` (`1494` passing, `5` pending)
- `git diff --check`

### DIFF
--- a/api/CHANGELOG.md
+++ b/api/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added optional `PackageManager.getVersion?(environment: PythonEnvironment): Promise<Pep440Version | undefined>` to return the version of the underlying package management tool, such as pip, uv, or conda.
 - Added optional `PackageManager.getPackageAvailableVersions?(environment: PythonEnvironment, packageName: string): Promise<Pep440Version[] | undefined>` to return the available versions of a package in newest-first order when supported.
 - Added optional `PackageManager.formatInstallSpec?(packageName: string, version: string): string` to format a versioned install specification using manager-specific syntax, such as `name==version` for pip or `name=version` for conda.
+- Added `PythonPackageGetterApi.getPackageAvailableVersions(environment: PythonEnvironment, packageName: string): Promise<Pep440Version[] | undefined>` so API consumers can query a package's available versions in newest-first order. Resolves to `undefined` when the environment's package manager does not support version listing.
 
 ### Changed
 
--- a/api/CHANGELOG.md
+++ b/api/CHANGELOG.md
@@ -22,7 +22,3 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Added the optional `options?: GetPackagesOptions` parameter to `PackageManager.getPackages(environment, options?)` and `PythonPackageGetterApi.getPackages(environment, options?)`. Consumers can set `options.skipCache` to request fresh package data.
-
-## [1.37.0]
-
-- Aligned the API package version with the Python Environments extension version.
--- a/api/CHANGELOG.md
+++ b/api/CHANGELOG.md
@@ -22,8 +22,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Added the optional `options?: GetPackagesOptions` parameter to `PackageManager.getPackages(environment, options?)` and `PythonPackageGetterApi.getPackages(environment, options?)`. Consumers can set `options.skipCache` to request fresh package data.
-- Changed `PackageManager.refresh(environment)` from `Promise<void>` to `Promise<Package[] | undefined>`, allowing implementations to return the refreshed package list.
-- Changed `PythonPackageGetterApi.refreshPackages(environment)` from `Promise<void>` to `Promise<Package[] | undefined>`, exposing the refreshed package list to API consumers.
 
 ## [1.37.0]
 
--- a/api/CHANGELOG.md
+++ b/api/CHANGELOG.md
@@ -5,6 +5,25 @@ All notable changes to the `@vscode/python-environments` API package are documen
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.1.0]
+
+### Added
+
+- Re-exported the `Pep440Version` type from `@renovatebot/pep440` for use with the new package version APIs.
+- Added the optional `PackageInfo.isTransitive?: boolean` property to indicate whether a package is a transitive dependency.
+- Added `GetPackagesOptions` with an optional `skipCache?: boolean` property. When `true`, package managers bypass cached data and query the underlying package management tool.
+- Added optional `PackageManager.getPackageWatchTargets?(environment: PythonEnvironment): RelativePattern[]` to return manager-specific filesystem patterns to monitor for package installation and removal changes, in addition to the default site-packages metadata locations.
+- Added optional `PackageManager.getDirectPackageNames?(environment: PythonEnvironment): Promise<Set<string> | undefined>` to return a best-effort set of direct, non-transitive package names when supported by the package manager.
+- Added optional `PackageManager.getVersion?(environment: PythonEnvironment): Promise<Pep440Version | undefined>` to return the version of the underlying package management tool, such as pip, uv, or conda.
+- Added optional `PackageManager.getPackageAvailableVersions?(environment: PythonEnvironment, packageName: string): Promise<Pep440Version[] | undefined>` to return the available versions of a package in newest-first order when supported.
+- Added optional `PackageManager.formatInstallSpec?(packageName: string, version: string): string` to format a versioned install specification using manager-specific syntax, such as `name==version` for pip or `name=version` for conda.
+
+### Changed
+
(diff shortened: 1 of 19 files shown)


edvilme:
Adding "skip tests" label since this functionality already exists but is now being exposed to the public API

StellaHuang95 (approved):

api/CHANGELOG.md
```diff
@@ -5,6 +5,26 @@ All notable changes to the `@vscode/python-environments` API package are documen
...
+- Added optional `PackageManager.formatInstallSpec?(packageName: string, version: string): string` to format a versioned install specification using manager-specific syntax, such as `name==version` for pip or `name=version` for conda.
+- Added `PythonPackageGetterApi.getPackageAvailableVersions(environment: PythonEnvironment, packageName: string): Promise<Pep440Version[] | undefined>` so API consumers can query a package's available versions in newest-first order. Resolves to `undefined` when the environment's package manager does not support version listing.
+
+### Changed
+
+- Added the optional `options?: GetPackagesOptions` parameter to `PackageManager.getPackages(environment, options?)` and `PythonPackageGetterApi.getPackages(environment, options?)`. Consumers can set `options.skipCache` to request fresh package data.
+- Changed `PackageManager.refresh(environment)` from `Promise<void>` to `Promise<Package[] | undefined>`, allowing implementations to return the refreshed package list.
```
eleanorjboyd: this is a big change here- we would need to go to 2.0 if we want to change the return value of this method. Any other ways we could maybe do that?

api/CHANGELOG.md
```diff
@@ -5,6 +5,26 @@ All notable changes to the `@vscode/python-environments` API package are documen
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.1.0]
```
eleanorjboyd, line 8: The published 1.0 package has top-level main and types; api/package.json now has only conditional exports. This accumulated unpublished change is not mentioned in the changelog.

eleanorjboyd: preserve top-level main and types alongside exports, and test packed ESM and legacy CommonJS consumers.

api/CHANGELOG.md
```diff
@@ -5,6 +5,26 @@ All notable changes to the `@vscode/python-environments` API package are documen
...
+- Added optional `PackageManager.formatInstallSpec?(packageName: string, version: string): string` to format a versioned install specification using manager-specific syntax, such as `name==version` for pip or `name=version` for conda.
+- Added `PythonPackageGetterApi.getPackageAvailableVersions(environment: PythonEnvironment, packageName: string): Promise<Pep440Version[] | undefined>` so API consumers can query a package's available versions in newest-first order. Resolves to `undefined` when the environment's package manager does not support version listing.
+
+### Changed
+
+- Added the optional `options?: GetPackagesOptions` parameter to `PackageManager.getPackages(environment, options?)` and `PythonPackageGetterApi.getPackages(environment, options?)`. Consumers can set `options.skipCache` to request fresh package data.
+- Changed `PackageManager.refresh(environment)` from `Promise<void>` to `Promise<Package[] | undefined>`, allowing implementations to return the refreshed package list.
+- Changed `PythonPackageGetterApi.refreshPackages(environment)` from `Promise<void>` to `Promise<Package[] | undefined>`, exposing the refreshed package list to API consumers.
+
```
eleanorjboyd: would also want to packed-package contract tests before publishing

eleanorjboyd (changes requested):
see two items about refresh and the exported top level items

api/CHANGELOG.md
edvilme: I can look into having it return nothing for now, and then maybe discuss later on when reviewing the API

api/CHANGELOG.md
edvilme: The api file and types are generated inline by the pipelines (#1619). I can update the changelog to mention that as well

StellaHuang95:

api/package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "@vscode/python-environments",
     "description": "An API facade for the Python Environments extension in VS Code",
-    "version": "1.0.0",
+    "version": "1.1.0",
```
StellaHuang95, line 4: > **Warning** · Non-blocking recommendation

Please confirm the published npm version history before releasing 1.1.0. The changelog contains a 1.37.0 entry, so publishing 1.1.0 would be a downgrade if 1.37.0 was ever published; reconcile the changelog or select a version above the published latest.

api/CHANGELOG.md
```diff
@@ -5,6 +5,26 @@ All notable changes to the `@vscode/python-environments` API package are documen
...
+- Added optional `PackageManager.formatInstallSpec?(packageName: string, version: string): string` to format a versioned install specification using manager-specific syntax, such as `name==version` for pip or `name=version` for conda.
+- Added `PythonPackageGetterApi.getPackageAvailableVersions(environment: PythonEnvironment, packageName: string): Promise<Pep440Version[] | undefined>` so API consumers can query a package's available versions in newest-first order. Resolves to `undefined` when the environment's package manager does not support version listing.
+
+### Changed
+
+- Added the optional `options?: GetPackagesOptions` parameter to `PackageManager.getPackages(environment, options?)` and `PythonPackageGetterApi.getPackages(environment, options?)`. Consumers can set `options.skipCache` to request fresh package data.
+- Changed `PackageManager.refresh(environment)` from `Promise<void>` to `Promise<Package[] | undefined>`, allowing implementations to return the refreshed package list.
+- Changed `PythonPackageGetterApi.refreshPackages(environment)` from `Promise<void>` to `Promise<Package[] | undefined>`, exposing the refreshed package list to API consumers.
```
StellaHuang95: > **Warning** · Non-blocking recommendation

Changing `PackageManager.refresh` from `Promise<void>` to `Promise<Package[] | undefined>` breaks external `PackageManager` implementations that return `Promise<void>`. Please confirm that external implementations are unsupported or document why this implementer-facing breaking change is appropriate for a minor release.

src/features/pythonApi.ts
```diff
@@ -319,6 +320,17 @@ export class PythonEnvironmentApiImpl implements PythonEnvironmentApi {
         }
```
StellaHuang95, line 320: > **Warning** · Non-blocking recommendation

Please add coverage for both manager behaviors: returning available versions when supported and resolving `undefined` when the underlying manager omits the optional hook. The latter relies on the internal wrapper's guard and is the documented fallback contract for this new public method.

StellaHuang95 (dismissed):
Approved via Review Center.

api/CHANGELOG.md
edvilme: Fixed in [b62da99](https://github.com/microsoft/vscode-python-environments/pull/1682/commits/b62da9904863721bf6fe9604571889126d4c4c53)

api/CHANGELOG.md
edvilme: Yes, fixed in bc432d0

api/package.json
edvilme: Yes, fixed in ea372a6
Last release on npm was 1.0.0

api/CHANGELOG.md
edvilme: Fixed in [b62da99](https://github.com/microsoft/vscode-python-environments/pull/1682/commits/b62da9904863721bf6fe9604571889126d4c4c53)

src/features/pythonApi.ts
edvilme: This follows the same design pattern from the other methods

eleanorjboyd (changes requested):
Thanks for addressing the previous feedback. The packed-package validation still has three blocking holes:

1. The publish pipeline runs `cp ../src/api.ts src/main.ts`, but `api/src` is ignored and absent from a clean checkout. The step fails before compilation unless it creates the directory first.

2. `api/test/consumer.ts` does not enforce the exact public return type. Assigning an explicitly narrow implementation to `PythonPackageGetterApi["getPackageAvailableVersions"]` still compiles if the published contract is widened to `Promise<unknown>` or `Promise<any>`. Please assert exact type equality or call the typed API and assign its result to `Promise<Pep440Version[] | undefined>`.

3. The package test exercises conditional `exports`; `require.resolve()` does not prove that top-level `main` is present, and `import.meta.resolve()` can resolve a missing target. Please inspect the installed package metadata and assert top-level `main` and `types` plus all referenced target files exist.

Also, normalize or realpath the expected and resolved temporary paths: on macOS Node may return `/private/var/...` for a test root created as `/var/...`, causing a false failure.

eleanorjboyd (approved):

PR #1685, opened 2026-07-31
Some refactoring around inline script manager

StellaHuang95:
This PR addresses the following feedback:

1. Make error message handling a helper function. 
2. Move all inline script related files into its own subfolders.

No code logics or implementations are changed, it's pure refactoring to address previous feedback.


src/features/inlineScript/lazyDetector.ts
edvilme, line 1: What is the difference of including this in src/common? I wonder if this is because it interacts with vscode, while the other are utils?

src/managers/builtin/inlineScript/envManager.ts
edvilme, line 1: Seems good to me, do we have a preference over src/managers/inlineScript?

edvilme (approved):
Thanks!

src/features/inlineScript/lazyDetector.ts
StellaHuang95: The common folder contains reusable parsing, cache, and interpreter primitives. lazyDetector subscribes to VS Code document events, owns activation/disposal lifecycle, and emits feature telemetry, so I kept it under features.

src/managers/builtin/inlineScript/envManager.ts
StellaHuang95: I prefer keeping it under managers/builtin. The top-level manager folders generally represent external tools such as conda, pyenv, poetry, and pipenv. The inline-script manager is extension-owned and reuses the built-in venv creation flow, so  builtin/inlineScript seems like the more consistent architectural fit.

eleanorjboyd (approved):

PR #1678, opened 2026-07-28
Adopt package manager command classes

edvilme:
## Summary

Adopts the package manager command classes introduced in #1621 and covered by the smoke tests in #1677 across Pip/UV, Conda, and Poetry.

## Changes

- Replaces existing command construction and execution paths with the new command classes.
- Preserves existing command flags, environment targeting, version ordering, prerelease defaults, timeouts, and Poetry cwd behavior.
- Removes superseded utility wrappers and obsolete tests tied to those utilities.

## Dependency

This PR is stacked on #1677 and intentionally contains only the migration and cleanup commit. Review and merge #1621 and #1677 first; this PR can then be retargeted to `main`.

This replaces #1675, whose base branch could not be changed after GitHub registered it as part of the original stack.

## Testing

- `npm run lint`
- `npm run compile-tests`
- `npm run unittest` (1,451 passing, 4 pending)

PR #1686, opened 2026-07-31
Adopt package manager command classes

### LINKED ISSUE #1650: Package list shows "No packages found" when pip emits an upgrade notice (JSON parse failure)
The Python Environments sidebar reports **"No packages found"** for a venv that actually has packages installed. The root cause is that the extension parses `pip list --format=json` output as JSON, but older pip versions append a `[notice] A new release of pip is available...` message to stdout after the JSON array, which breaks parsing.

**Environment:**
- Python Envs extension version: `1.37.2026070201`
- OS: macOS
- Env type: venv (`ms-python.python:venv`)
- Python: 3.10.20
- pip: 23.0.1 (bundled; emits upgrade notice to stdout)

**Steps to reproduce:**
1. Create/select a venv whose pip is outdated enough to print an upgrade notice (e.g. pip 23.0.1).
2. Install one or more packages (e.g. `pytest`).
3. Open the Python Environments sidebar and click the environment.

**Expected behavior:**
The installed packages (pytest, pluggy, iniconfig, etc.) are listed.

**Actual behavior:**
The sidebar shows "No packages found." The output log shows the package refresh failing to parse JSON:

```
Running: .../.venv/bin/python -m pip list --format=json
python: [{"name": "pip", "version": "23.0.1"}, {"name": "setuptools", "version": "79.0.1"}]
python:
[notice] A new release of pip is available: 23.0.1 -> 26.1.2
[notice] To update, run: .../.venv/bin/python -m pip install --upgrade pip
Failed to parse pip list JSON output Unexpected non-whitespace character after JSON at position 85 (line 3 column 1)
```

The packages are genuinely installed — a manual `pip list` confirms pytest 9.1.1 and its dependencies are present. Only the extension's parsing fails.

**Additional related symptoms (same root cause):**
- A "problem installing packages" notification can appear even when the install itself succeeds, because the post-install package refresh fails to parse.
- Transient `No module named pip` errors immediately after in-place venv recreation (`python3.10 -m venv .venv`), before pip is bootstrapped, trigger the same refresh failure path.

**Suggested fix:**
Make the pip-list parser robust to non-JSON trailing output. Options:
- Extract only the JSON array/object from stdout before parsing (e.g. slice from the first `[`/`{` to its matching close), rather than `JSON.parse(entireStdout)`.
- Run pip with `--disable-pip-version-check` (and/or `PIP_DISABLE_PIP_VERSION_CHECK=1`) when listing packages to suppress the notice.
- Pass `--no-python-version-warning` / route notices to stderr where feasible.

edvilme:
## Summary

Adopts the package manager command classes introduced in #1621 across Pip/UV, Conda, and Poetry while preserving existing behavior and error contracts.

## Changes

- Replaces existing command construction and execution paths with the new command classes.
- Uses JSON unconditionally for machine-readable commands when the installed tool supports it, including all required flags.
- Uses a dedicated text parser for Pip 21.2-25.0; Pip 25.1+ always uses `pip index versions --json`.
- Preserves operation-specific timeout behavior: install/uninstall mutations are unbounded, while list/direct-list operations retain 30-second timeouts.
- Propagates malformed version output as an operational failure rather than treating it as a successful empty result.
- Keeps destructive environment removal confirmation bypass explicit through `runHeadless`.

## Relationship

- Targets `main` directly and includes the changes merged in #1717.
- Replaces #1678, whose base branch could not be changed after GitHub registered it as part of a stack.
- Deterministic command argument and parser tests are intentionally included here so the adopted implementations are covered in the same PR. The superseded network-dependent roundtrip test was removed in favor of the controlled package-manager integration fixtures.

## Testing

- `npm run lint`
- `npm run compile-tests`
- `npm run unittest` (1,855 passing, 6 pending)
- Focused package-manager command, selection, timeout, parsing, and removal-confirmation tests


src/managers/builtin/pipPackageManager.ts
```diff
@@ -65,43 +81,66 @@ export class PipPackageManager implements PackageManager, Disposable {
...
+            }
+
+            await updatePackagesAndNotify(this, environment, this.packages.get(environment.envId.id), (changes) => {
+                this._onDidChangePackages.fire({ environment, manager: this, changes });
+            });
+        } catch (e) {
+            if (e instanceof CancellationError) {
+                // Cancellation is a normal control-flow exit; do not surface an error.
+                return;
+            }
```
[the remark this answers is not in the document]
edvilme: Fixed: manage() now rethrows CancellationError instead of returning early, so callers like venv creation (pkgInstallationCancelled) can distinguish cancel from failure. Restores parity with main and the conda/poetry managers.

src/managers/builtin/pipPackageManager.ts
```diff
@@ -205,38 +255,22 @@ export class PipPackageManager implements PackageManager, Disposable {
     }
 
     /**
-     * Returns direct (non-transitive) package names using `pip list --not-required` or `uv pip tree --depth=0`.
+     * Returns direct (non-transitive) package names using `pip list --not-required` or `uv pip list --not-required`.
      *
```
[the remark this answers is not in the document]
edvilme: Fixed: the docstring now reflects that uv uses `uv pip tree --depth=0` while pip uses `pip list --not-required`.

src/managers/builtin/utils.ts
```diff
@@ -22,13 +13,22 @@ import {
...
 
 const PIXI_EXTENSION_ID = 'renan-r-santos.pixi-code';
 const PIXI_RECOMMEND_DONT_ASK_KEY = 'pixi-extension-recommend-dont-ask';
 let pixiRecommendationShown = false;
 
+/**
+ * Parse package specifications (strings) into package objects.
+ * Each string becomes a package object with packageName and empty version.
+ */
+export function parsePackageSpecs(packageStrings: string[]): { packageName: string; version?: string }[] {
```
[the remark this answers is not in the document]
edvilme: Fixed: parsePackageSpecs now omits the version property instead of setting it to undefined, and the wording is clarified.

src/test/managers/builtin/commands.unit.test.ts
```diff
@@ -0,0 +1,111 @@
...
+    teardown(() => {
+        sinon.restore();
+    });
+
+    test('PipAvailableVersionsCommand executes without error', async () => {
+        runPythonStub.resolves(JSON.stringify({ versions: ['1.0.0'] }));
+        const command = new PipAvailableVersionsCommand({ pythonExecutable: 'python', log: mockLog });
+
+        await assert.doesNotReject(() => command.execute({ packageName: 'package', pythonVersion: '3.13.1' }));
+    });
```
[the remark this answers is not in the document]
edvilme: Addressed: the command smoke tests were removed from this PR to avoid overlap. The reworked versions asserting parsed values live in #1677 (e.g. PipAvailableVersionsCommand parses versions and filters prereleases).

src/test/managers/conda/commands.unit.test.ts
```diff
@@ -0,0 +1,73 @@
...
+    teardown(() => {
+        sinon.restore();
+    });
+
+    test('CondaAvailableVersionsCommand executes without error', async () => {
+        runCondaStub.resolves(JSON.stringify({ package: [{ version: '1.0.0' }] }));
+        const command = new CondaAvailableVersionsCommand({ pythonExecutable: 'conda', log: mockLog });
+
+        await assert.doesNotReject(() => command.execute({ packageName: 'package', pythonVersion: '' }));
+    });
```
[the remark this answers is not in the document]
edvilme: Addressed: the command smoke tests were removed from this PR to avoid overlap. The reworked version asserting parsed values (CondaAvailableVersionsCommand) lives in #1677.

src/test/managers/poetry/commands.unit.test.ts
```diff
@@ -0,0 +1,62 @@
...
+
+        await assert.doesNotReject(() => command.execute({ packages: [{ packageName: 'package' }] }));
+    });
+
+    test('PoetryShowCommand executes without error', async () => {
+        runPoetryStub.resolves('package 1.0.0 description');
+        const command = new PoetryShowCommand({ pythonExecutable: 'poetry', cwd: 'project', log: mockLog });
+
+        await assert.doesNotReject(() => command.execute());
+    });
```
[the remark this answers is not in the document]
edvilme: Addressed: the command smoke tests were removed from this PR to avoid overlap. The reworked version asserting parsed values (PoetryShowCommand) lives in #1677.

src/managers/builtin/helpers.ts
```diff
@@ -153,7 +153,7 @@ export async function runPython(
         proc.stderr?.on('data', (data) => {
             const s = data.toString('utf-8');
             builder += s;
-            log?.append(`python: ${s}`);
+            log?.append(s);
         });
```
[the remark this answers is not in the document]
edvilme: Fixed: restored the `python:` prefix on stderr log output for consistency with stdout.

StellaHuang95:

src/test/integration/packageManagerRoundtrip.integration.test.ts
```diff
@@ -0,0 +1,246 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
StellaHuang95: > **Issue** · Please address or respond

The file-scoped helpers and capability type precede the `suite(...)` block, violating the test-helper placement rule. Move them below the suite so the test's setup, execution, and assertions are the primary flow. [verified]

src/managers/poetry/poetryPackageManager.ts
```diff
@@ -192,31 +186,34 @@ export class PoetryPackageManager implements PackageManager, Disposable {
...
-                throw err;
-            }
+            const removeCmd = new PoetryRemoveCommand({
+                pythonExecutable: poetry,
+                cwd,
+                log: this.log,
+            });
+            const packages = parsePackageSpecs(options.uninstall);
+            await withProgress(
+                { location: ProgressLocation.Notification, title: 'Managing packages with Poetry', cancellable: true },
```
StellaHuang95: > **Warning** · Non-blocking recommendation

`runPoetryManage` and `getDirectPackageNames` previously inherited `process.cwd()`, but now use the environment-derived Poetry cwd. Because Poetry resolves its project and lockfile from cwd, this can change which project package operations affect. Confirm this is intentional, preserve the prior behavior if not, and add focused coverage. [verified]

src/managers/builtin/pipPackageManager.ts
```diff
@@ -65,43 +81,68 @@ export class PipPackageManager implements PackageManager, Disposable {
...
+            // Execute uninstall if needed
+            if (toUninstall.length > 0) {
+                const uninstallCmd: PipUninstallCommand | UvUninstallCommand = await createPipOrUvCommand(
+                    manageCommandOptions,
+                    environment.environmentPath.fsPath,
+                    PipUninstallCommand,
+                    UvUninstallCommand,
+                );
+                const packages = parsePackageSpecs(toUninstall);
+                await withProgress(
```
StellaHuang95: > **Warning** · Non-blocking recommendation

The uninstall operation now has its own progress scope but still uses the title `Installing packages`. Use an uninstall-specific title so the notification reflects the operation in progress. [verified]

src/managers/builtin/pipPackageManager.ts
```diff
@@ -65,43 +81,68 @@ export class PipPackageManager implements PackageManager, Disposable {
...
+                throw new Error('Unable to determine Python executable path');
+            }
+
+            // Centralize command options for install/uninstall operations
+            const manageCommandOptions: CommandConstructorOptions = {
+                pythonExecutable,
+                log: this.log,
+            };
+
+            // Execute uninstall if needed
```
StellaHuang95: > **Warning** · Non-blocking recommendation

The previous pip management path rejected Python 2 environments, but that guard is no longer present here. Confirm the command classes enforce the same unsupported-version behavior or restore the guard to avoid a regression. [verified]

StellaHuang95:
GitHub cannot anchor PR review comments to unchanged lines in the diff. Falling back to a general PR comment for [src/test/managers/builtin/pipListUtils.unit.test.ts:L1](https://github.com/microsoft/vscode-python-environments/blob/01575b10ff6bf783c1a3c936d577e26d174a2162/src/test/managers/builtin/pipListUtils.unit.test.ts#L1).

Deleting these parser tests removes deterministic malformed-output coverage for pip list, uv tree, and available-version parsing. The new live integration test does not exercise those branches; retain equivalent command-layer unit coverage in this change or ensure it is already present before removing these tests. [verified]

src/test/integration/packageManagerRoundtrip.integration.test.ts
```diff
@@ -0,0 +1,246 @@
...
+    test('install/list/direct/uninstall roundtrip for each package manager', async function () {
+        const byManager = await getEnvironmentsByManager();
+
+        if (byManager.size === 0) {
+            console.log('No environments discovered; skipping package manager roundtrip');
+            this.skip();
+            return;
+        }
+
+        const failures: string[] = [];
```
StellaHuang95: > **Warning** · Non-blocking recommendation

This only prefers venv-like environments; when none is found it still installs and uninstalls Flask in the first discovered environment for each manager, which can be a global or system interpreter. Restrict execution to environments proven safe to modify, and skip managers without one. [verified]

edvilme: Fixed in 7e9c7a3. The test now only selects explicitly isolated environments (venv, Poetry, Pipenv, and non-base Conda); system, global, base Conda, Pyenv, and unknown managers are skipped.

src/managers/poetry/poetryPackageManager.ts
edvilme: Fixed in 0f2ac6f. Poetry add/remove and direct-package listing now preserve the previous behavior by inheriting the process working directory. Project-derived cwd remains limited to package fetching, and focused unit tests cover both restored paths.

src/managers/builtin/pipPackageManager.ts
edvilme: Fixed in e4c7f10. Pip package management once again rejects Python 2 environments before selecting or executing a pip/uv command, with regression coverage confirming command selection is not reached.

edvilme:
@StellaHuang95 Addressed the parser-coverage recommendation in 39f8559. Added deterministic command-layer tests for pip/uv list parsing, malformed and non-array output, incomplete entries, uv tree top-level filtering, direct-name normalization, and available-version parsing/filtering. This retains coverage without restoring the superseded legacy parser helpers.

src/test/integration/packageManagerRoundtrip.integration.test.ts
edvilme: Fixed in 29c2d41. The suite now appears before the file-scoped constants, capability type, and helper functions, keeping the test flow primary.

src/managers/builtin/pipPackageManager.ts
edvilme: Fixed in 29c2d41. Pip uninstall now uses the localized `Uninstalling packages` progress title, with unit coverage for the notification title.

src/test/integration/packageManagerRoundtrip.integration.test.ts
edvilme, line 1: Fixes #1622

eleanorjboyd (changes requested):
The command-class refactor looks directionally good, and I did not find a confirmed production-code regression. I am requesting changes because the new coverage currently gives false confidence and required CI is still red:

1. The available-version roundtrip branch casts `PythonEnvironmentApi` to an optional `getPackageAvailableVersions` capability, but that method is not on the exported API, so the guard is always false and the coverage never runs.

2. The transitivity assertions allow `isTransitive: undefined`; if enrichment disappears for every package, dependency classification is skipped and the test still passes.

3. Poetry and uv are core parts of this refactor but lack behavioral command/lifecycle coverage. The Poetry tests only verify the mocked runner cwd and do not validate verbs, package arguments, parsing, or add/remove behavior; uv install/uninstall is likewise unexercised.

Please make these assertions fail on the corresponding regressions and resolve the current smoke, E2E, and integration failures before re-review.

src/managers/builtin/pipPackageManager.ts
```diff
@@ -179,49 +243,42 @@ export class PipPackageManager implements PackageManager, Disposable {
...
+                    return undefined;
+                }
             }
 
-            // pip < 21.2 - version picking is undefined; `pip index versions` is unavailable.
+            const versions = await availableVersionsCmd.execute({
+                packageName,
+                pythonVersion: baseVersion,
+            });
+            return versions.sort((a, b) => compare(b.public, a.public));
```
StellaHuang95: > **Issue** · Please address or respond

📍 src/managers/builtin/pipPackageManager.ts:271
Pip 21.2–25.0 passes this compatibility gate, but `PipAvailableVersionsCommand` always adds `--json`; runtime verification with Pip 24.x fails with `no such option: --json`, causing version lookup to return `undefined`. Preserve the text-output path below Pip 25.1 or make the command version-aware.

[verified]

src/managers/builtin/pipUtils.ts
```diff
@@ -275,7 +277,17 @@ export async function getWorkspacePackagesToInstall(
...
-        installed = (await refreshPipPackages(environment, log, { showProgress: true }))?.map((pkg) => pkg.name);
+        const pythonExecutable = environment.execInfo?.run?.executable;
+        if (pythonExecutable) {
+            const listCmd: PipListCommand | UvListCommand = await createPipOrUvCommand(
+                { pythonExecutable, log },
+                environment.environmentPath.fsPath,
+                PipListCommand,
+                UvListCommand,
+            );
+            const data = await listCmd.execute();
```
StellaHuang95: > **Issue** · Please address or respond

📍 src/managers/builtin/pipUtils.ts:288
A rejected `PipListCommand.execute()` now propagates and prevents the interactive package picker from opening; this was runtime-verified. Catch and log listing failures here, then continue with an empty installed-package list as the removed wrapper did.

[verified]

src/test/managers/builtin/pipPackageManager.unit.test.ts
```diff
@@ -28,11 +31,13 @@ suite('PipPackageManager', () => {
...
-        const refreshPackages = sinon.stub(builtinUtils, 'refreshPipPackages');
-        refreshPackages
+        sinon.stub(helpers, 'shouldUseUv').resolves(false);
+        sinon.stub(windowApis, 'showErrorMessage').resolves(undefined);
+        const execute = sinon.stub(PipListCommand.prototype, 'execute');
+        execute
             .onFirstCall()
             .resolves([{ name: 'pip', version: '25.0', displayName: 'pip', description: '25.0' }]);
-        refreshPackages.onSecondCall().resolves(undefined);
+        execute.onSecondCall().rejects(new Error('pip list failed'));
```
StellaHuang95: > **Issue** · Please address or respond

📍 src/test/managers/builtin/pipPackageManager.unit.test.ts:33
These tests stub `execute()`, so they do not replace the deleted deterministic coverage for command arguments and parsers; the integration suite is network-gated and does not cover every manager. Restore command-level tests in this PR or make the referenced test PR a merge prerequisite.

[verified]

src/managers/conda/condaPackageManager.ts
```diff
@@ -23,10 +23,18 @@ import {
 } from '../../api';
 import { showErrorMessageWithLogs } from '../../common/errors/utils';
 import { CondaStrings } from '../../common/localize';
-import { traceError } from '../../common/logging';
 import { withProgress } from '../../common/window.apis';
+
+import { parsePackageSpecs } from '../builtin/utils';
```
StellaHuang95: > **Info** · Optional note

📍 src/managers/conda/condaPackageManager.ts:28
Conda and Poetry now depend on helpers under the Pip-oriented `builtin` subtree. Move cross-manager package-spec and normalization helpers to `managers/common` or `managers/base` to keep dependency direction neutral.

[verified]

StellaHuang95:
Non-blocking: Poetry command helpers re-discover the executable instead of using the command object's configured executable, so configured executable selection is ignored.

src/managers/builtin/pipPackageManager.ts
edvilme: Fixed in 4b4b292. PipAvailableVersionsCommand now uses text output for Pip 21.2–25.0 and JSON only for Pip 25.1+, with deterministic tests covering both command arguments and parsers.

src/managers/builtin/pipUtils.ts
edvilme: Fixed in 4b4b292. getWorkspacePackagesToInstall now catches and logs list-command failures, continues with an empty installed-package list, and has regression coverage confirming the picker still opens.

src/test/managers/builtin/pipPackageManager.unit.test.ts
edvilme: Fixed in 4b4b292. Restored deterministic command suites for Pip/UV, Conda, and Poetry with 21 tests asserting concrete command arguments, parser results, version handling, normalization, and Poetry cwd propagation.

src/managers/conda/condaPackageManager.ts
edvilme: Addressed in 4b4b292. parsePackageSpecs and normalizePackageName now live in managers/common/packageUtils, and Pip, Conda, Poetry, package-change, and cache-key consumers use the neutral module.

src/managers/poetry/poetryPackageManager.ts
```diff
@@ -236,126 +242,64 @@ export class PoetryPackageManager implements PackageManager, Disposable {
...
-        }
-
-        const poetryPackages: { name: string; version: string; displayName: string; description: string }[] = [];
+        const cwd = await this.getPoetryCwd(environment);
+        const showCmd = new PoetryShowCommand({
+            pythonExecutable: poetry,
+            cwd,
+            log: this.log,
+        });
+        const data = await showCmd.execute();
```
StellaHuang95: > **Issue** · Please address or respond

`PoetryShowCommand.execute()` is no longer inside the prior best-effort error boundary, so a failed `poetry show` now rejects `getPackages()`. Catch and log the failure here and return `[]` to preserve package-tree loading behavior.

[verified]

src/managers/poetry/poetryPackageManager.ts
```diff
@@ -198,31 +207,28 @@ export class PoetryPackageManager implements PackageManager, Disposable {
                 ),
             );
         }
+        const cwd = await this.getPoetryCwd(environment);
```
StellaHuang95: > **Warning** · Non-blocking recommendation

This now supplies an environment-derived project cwd to add/remove and top-level-listing commands, whereas those paths previously inherited `process.cwd()`. Preserve the prior cwd behavior, or add focused coverage demonstrating that the changed project selection is intentional and safe.

[verified]

StellaHuang95:
The new Poetry command path also has an executable-source mismatch: its runner rediscovers Poetry instead of spawning the command instance's injected executable. Thread the injected executable into the runner so selection remains consistent.

src/managers/poetry/poetryPackageManager.ts
edvilme: Fixed in d2c0993. PoetryShowCommand failures are now caught and logged by fetchPackagesFromTool, which returns an empty package list to preserve best-effort package-tree loading. Added focused regression coverage.

src/managers/poetry/poetryPackageManager.ts
edvilme: Fixed in d2c0993. Poetry add/remove and top-level listing no longer receive a project-derived cwd, so they inherit process.cwd() as before. Project cwd remains limited to poetry show package fetching, with focused manager-level tests for both paths.

StellaHuang95:
Non-blocking: Poetry command objects currently rediscover the executable instead of using their injected `pythonExecutable`, leaving two sources of truth. Consider threading the selected executable through `runPoetry`.

StellaHuang95 (dismissed):
Approved via Review Center.

rchiodo (dismissed):
Approved via Review Center.

src/managers/conda/condaPackageManager.ts
```diff
@@ -128,34 +152,13 @@ export class CondaPackageManager implements PackageManager, Disposable {
...
-                        ),
-                    );
-                }
-            }
+            const listCmd = new CondaListCommand({
+                pythonExecutable: 'conda',
+                condaEnvironmentPath: environment.environmentPath.fsPath,
+                log: this.log,
+            });
+            const data = await listCmd.execute();
```
StellaHuang95, line 179: > **Issue** · Please address or respond

`CondaListCommand` converts malformed JSON output to `[]`, so this path caches a failed list operation as a successful empty environment and will not retry. Preserve a distinct failure result or reject parse failures, and cache only successfully parsed output.

StellaHuang95:
The Poetry command runner also re-resolves the executable instead of using the injected one, leaving two executable sources of truth.

src/managers/conda/condaPackageManager.ts
edvilme: Fixed in b2eff5a. `CondaListCommand` now rejects malformed JSON and valid non-array JSON instead of returning `[]`; only a successfully parsed empty array represents an empty environment. Because `CondaPackageManager` caches only after the command resolves, a failed list is retried and only the later successful result is cached. Added command-level parsing coverage and a manager retry/cache regression test.

edvilme: Fixed in [b2eff5a](https://github.com/microsoft/vscode-python-environments/pull/1686/commits/b2eff5ab2377dfaa5468b8893afc5752e3baa025)

eleanorjboyd (dismissed):
Thanks for consolidating these package-manager operations behind command classes. There is one behavior-preservation issue that needs addressing before approval:

The migrated Pip/UV commands all inherit the five-minute timeout from the base command. Pip and UV install/uninstall previously called their runners without a timeout, so slow source builds can now be terminated after five minutes. In the other direction, list/direct-list operations previously used a 30-second timeout and can now hang for up to five minutes before failing. Please make timeout policy operation-specific (preserving unbounded mutations and 30-second list operations), or explicitly document and test an intentional policy change.

I confirmed that the network lifecycle tests are enabled in both PR and push CI via `VSC_PYTHON_PACKAGE_NETWORK_TEST=1`.

src/managers/builtin/pipPackageManager.ts
edvilme: Fixed in d660a2f. Pip 25.1+ selects the JSON-only command, while Pip 21.2-25.0 now selects a dedicated text-output command, preserving compatibility without an optional useJson parameter.

src/managers/conda/condaPackageManager.ts
```diff
@@ -190,25 +188,12 @@ export class CondaPackageManager implements PackageManager, Disposable {
...
-
-        return Array.from(uniqueVersions.values()).sort((a: Pep440Version, b: Pep440Version) =>
-            rcompare(a.public, b.public),
-        );
+        const availableVersionsCmd = new CondaAvailableVersionsCommand({
+            pythonExecutable: 'conda',
+            log: this.log,
+        });
+        const versions = await availableVersionsCmd.execute({ packageName, pythonVersion: '' });
+        return versions.sort((a, b) => compare(b.public, a.public));
```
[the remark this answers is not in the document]
edvilme: Fixed in d660a2f. Conda version lookup now rejects malformed JSON, unexpected schemas, and invalid version entries; only a valid empty package array returns []. Deterministic parser tests cover each case.

src/managers/builtin/pipPackageManager.ts
```diff
@@ -105,23 +155,27 @@ export class PipPackageManager implements PackageManager, Disposable {
...
                 cancellable: true,
             },
-            async (_progress, token) => execute(token),
+            (_progress, token) => execute(token),
         );
     }
 
     async refresh(environment: PythonEnvironment): Promise<void> {
-        await withProgress(
+        await window.withProgress(
```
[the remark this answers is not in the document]
edvilme: Fixed in d660a2f. The Pip manager now uses the established showErrorMessage and withProgress wrappers from window.apis so tests can stub UI consistently.

src/managers/builtin/commands/availableVersions.ts
```diff
@@ -30,6 +30,11 @@ export class PipAvailableVersionsCommand extends AvailableVersionsCommand {
...
             this.timeout,
         );
+        if (executeArgs.useJson === false) {
+            const match = output.match(/^Available versions:\s*(.+)$/im);
+            return this.parseVersions(match?.[1].split(',') ?? [], executeArgs.includePrerelease);
+        }
+
         const match = output.match(/{[\s\S]*}/);
         if (!match) {
             return [];
```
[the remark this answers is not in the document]
edvilme: Fixed in d660a2f. Pip and UV version commands now propagate malformed JSON and schema-invalid output. A valid versions: [] payload remains the only successful empty result.

src/managers/builtin/commands/availableVersions.ts
```diff
@@ -2,26 +2,26 @@ import type { Pep440Version } from '@renovatebot/pep440';
...
-            executeArgs.packageName,
-            '--json',
-            '--python-version',
-            executeArgs.pythonVersion,
-        ];
+    protected buildCommand(executeArgs: PipAvailableVersionsExecuteArgs): string[] {
+        const args = ['-m', 'pip', 'index', 'versions', executeArgs.packageName];
+        if (executeArgs.useJson !== false) {
+            args.push('--json');
+        }
```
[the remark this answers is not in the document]
edvilme: Fixed in d660a2f. The JSON-capable Pip command always uses --json and --disable-pip-version-check. The dedicated older-Pip text command also suppresses the version check.

src/managers/builtin/venvUtils.ts
```diff
@@ -580,6 +581,7 @@ export async function removeVenv(
 
     const confirmed =
         options?.runHeadless === true ||
+        isTestExecution() ||
```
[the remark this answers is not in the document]
edvilme: Fixed in d660a2f. Test execution no longer bypasses confirmation process-wide; only an explicit runHeadless option skips the prompt, including controlled fixture cleanup.

src/test/managers/builtin/commands.unit.test.ts
```diff
@@ -0,0 +1,341 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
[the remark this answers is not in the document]
edvilme: Addressed in d660a2f and the PR description. Deterministic command argument and parser tests are intentionally retained in this PR so the adopted implementations are covered alongside the change; the stale claim that they only live in #1677 has been removed.

src/managers/builtin/pipPackageManager.ts
```diff
@@ -276,79 +312,25 @@ export class PipPackageManager implements PackageManager, Disposable {
...
-/** Parses the legacy text output from `pip index versions <package>`. */
-export function parsePipIndexVersionsText(output: string): Pep440Version[] | undefined {
-    const match = output.match(/^Available versions:\s*(.+)$/im);
-    if (!match) {
-        return undefined;
+        const listDirectNamesCmd: PipListDirectNamesCommand | UvListDirectNamesCommand = await createPipOrUvCommand(
+            { pythonExecutable, log: this.log },
+            environment.environmentPath.fsPath,
+            PipListDirectNamesCommand,
+            UvListDirectNamesCommand,
```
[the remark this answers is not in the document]
edvilme: Fixed in d660a2f. PipListDirectNamesCommand now passes --disable-pip-version-check, with command-argument coverage.

edvilme:
The timeout behavior requested in the outstanding review is restored in d660a2f: PackageManagerCommand is unbounded by default so install/uninstall mutations are not terminated during slow builds, while ListCommand and ListDirectNamesCommand retain explicit 30-second timeouts. Unit tests cover both policies.

src/managers/builtin/pipUtils.ts
```diff
@@ -275,7 +277,22 @@ export async function getWorkspacePackagesToInstall(
...
+        const pythonExecutable = environment.execInfo?.run?.executable;
+        if (pythonExecutable) {
+            try {
+                const listCmd: PipListCommand | UvListCommand = await createPipOrUvCommand(
+                    { pythonExecutable, log },
+                    environment.environmentPath.fsPath,
+                    PipListCommand,
+                    UvListCommand,
+                );
+                const data = await listCmd.execute();
```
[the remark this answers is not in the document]
edvilme: We should listen to the `options.showProgress` and conditionally wrap it with `withProgress`

src/managers/builtin/pipPackageManager.ts
```diff
@@ -150,124 +204,111 @@ export class PipPackageManager implements PackageManager, Disposable {
...
+const pipVersionCmd = new PipVersionCommand({ pythonExecutable, log: this.log });
+            const pipVersion = await pipVersionCmd.execute();
+            if (!pipVersion) {
+            }
+            if (compare(pipVersion.public, '21.2.0') < 0) {
+                throw new PackageVersionLookupNotSupportedError(
+                    `Package version lookup requires pip 21.2 or newer; the environment has pip ${pipVersion.public}.`,
+                );
+            }
+            if (compare(pipVersion.public, '25.1') >= 0) {
```
StellaHuang95, line 292: > **Issue** · Please address or respond

📍 src/managers/builtin/pipPackageManager.ts:286
`PipVersionCommand.execute()` can return `undefined`, and the empty guard leaves subsequent `pipVersion.public` accesses uncompilable and capable of producing a raw `TypeError`. Throw a descriptive operational error in this branch and add a regression test for unparseable `pip --version` output.

[verified]

src/managers/builtin/pipPackageManager.ts
```diff
@@ -150,124 +204,111 @@ export class PipPackageManager implements PackageManager, Disposable {
...
+            return packages;
+        } catch (error) {
+            this.log.error('Error refreshing packages', error);
+            if (showErrors) {
+                setImmediate(async () => {
+const viewOutput = l10n.t('View Output');
+                    const result = await showErrorMessage(l10n.t('Error refreshing packages'), viewOutput);
+                    if (result === viewOutput) {
+                    }
+                });
```
StellaHuang95, line 232: > **Issue** · Please address or respond

📍 src/managers/builtin/pipPackageManager.ts:225
Selecting **View Output** is verified to leave the output channel hidden because this branch is empty. Restore `showErrorMessageWithLogs` or call `this.log.show()` so the advertised action works and the existing refresh-error contract remains centralized.

[verified]

src/managers/builtin/pipPackageManager.ts
```diff
@@ -150,124 +204,111 @@ export class PipPackageManager implements PackageManager, Disposable {
...
-                python,
-                ['-m', 'pip', 'index', 'versions', packageName, '--python-version', baseVersion],
-                undefined,
-                this.log,
-            );
-            return requireParsedVersions(parsePipIndexVersionsText(output), 'pip');
+            const textCommand = new PipAvailableVersionsTextCommand({ pythonExecutable, log: this.log });
+            const textVersions = await textCommand.execute({ packageName, pythonVersion: baseVersion });
+            return textVersions.sort((a, b) => compare(b.public, a.public));
         }
```
StellaHuang95, line 303: > **Warning** · Non-blocking recommendation

📍 src/managers/builtin/pipPackageManager.ts:275
The factory’s Pip/UV decision is recovered through `instanceof PipAvailableVersionsCommand`, leaving orchestration coupled to concrete command identity. Return a discriminated factory result or encapsulate the version-dependent strategy so callers do not need to rediscover which backend was selected.

[verified]

StellaHuang95:
GitHub cannot anchor PR review comments to unchanged lines in the diff. Falling back to a general PR comment for [src/managers/base/commands/packageManagerCommand.ts:L36](https://github.com/microsoft/vscode-python-environments/blob/b54fcd91759ad8aeacdf9741a8cd14849ace0ed3/src/managers/base/commands/packageManagerCommand.ts#L36).

📍 src/managers/base/commands/packageManagerCommand.ts:28
`pythonExecutable` represents an interpreter for UV, a Poetry executable, and a dummy value for Conda, so the shared base contract is misleading. Consider moving manager-specific execution context into specialized constructor options while retaining only genuinely common fields in the base.

[verified]

StellaHuang95 (dismissed):

heejaechang (dismissed):
Approved via Review Center.

rchiodo (dismissed):

src/managers/builtin/pipPackageManager.ts
edvilme: Fixed in 2157f65. The localized View Output action now calls this.log.show(), restoring the refresh-error behavior after the merged suggestion left the branch empty.

edvilme: Follow-up in f238e7c: removed the conditional branch entirely and restored the centralized showErrorMessageWithLogs path for Pip refresh failures. Added a regression test verifying the helper receives the refresh error and manager output channel.

src/managers/builtin/pipPackageManager.ts
edvilme: Fixed in 2157f65. The undefined Pip version branch now throws a descriptive operational error before any .public access, and a regression test covers successful but unparseable pip --version output.

src/managers/builtin/pipPackageManager.ts
edvilme: Addressed in c5d3d23. The factory now offers a discriminated Pip/UV result, and available-version orchestration switches on result.kind instead of inferring the backend via instanceof. Added UV-path coverage proving Pip version lookup is not invoked.

edvilme:
Regarding the non-blocking base-constructor recommendation: agreed that `pythonExecutable` is manager-specific despite living in the shared options today. I am leaving that broader constructor hierarchy redesign out of this adoption PR because it affects every Pip/UV, Conda, and Poetry command and is not required for behavioral correctness; this PR keeps the existing command contract while removing the concrete-backend inference called out in the inline review.

src/managers/conda/condaPackageManager.ts
```diff
@@ -136,34 +154,13 @@ export class CondaPackageManager implements PackageManager, Disposable {
...
-                        ),
-                    );
-                }
-            }
+            const listCmd = new CondaListCommand({
+                pythonExecutable: 'conda',
+                condaEnvironmentPath: environment.environmentPath.fsPath,
+                log: this.log,
+            });
+            const data = await listCmd.execute();
```
StellaHuang95, line 179: > **Warning** · Non-blocking recommendation

`CondaListCommand` now throws for malformed or non-array JSON, but this boundary propagates that rejection whereas the previous manager returned an empty package list. Catch and log parsing/shape failures here, return `[]`, and update the manager test to preserve the established caller contract.

[verified]

StellaHuang95 (dismissed):

heejaechang (dismissed):

rchiodo (dismissed):

src/managers/conda/condaPackageManager.ts
StellaHuang95, line 179: > **Issue** · Please address or respond

CondaListCommand parse failures now escape `getPackages()`, so a follow-up refresh can report an otherwise successful mutation as failed. Catch and log malformed-output errors at this manager boundary, return `[]` without caching it, and update the retry test to preserve the prior best-effort contract.

[verified]

edvilme: Fixed in 8558073. CondaListCommand now identifies malformed JSON/schema with CondaListOutputError; the manager catches and logs only that output error, returns [] without caching it, and retries on the next request. Genuine Conda process failures still propagate. The retry test now covers the restored best-effort contract.

edvilme: Follow-up in cb2fab4: internal mutation/refresh paths now consume an undefined fetch-failure signal, so malformed output cannot produce false removal events or replace a valid cache with []. The public getPackages fallback still returns [] without caching and retries when uncached. Command-level duplicate logging was also removed; manager-boundary logging remains. Added mutation and cached-refresh regression tests.

StellaHuang95 (approved):

heejaechang (dismissed):

rchiodo (dismissed):

src/managers/builtin/commands/availableVersions.ts
```diff
@@ -2,9 +2,29 @@ import type { Pep440Version } from '@renovatebot/pep440';
 import { AvailableVersionsCommand, type AvailableVersionsExecuteArgs } from '../../base/commands/index';
 import { runPython, runUV } from '../helpers';
 
+function parseVersionsJson(output: string, tool: 'pip' | 'uv'): string[] {
```
StellaHuang95, line 5: > **Warning** · Non-blocking recommendation

The greedy `/{[\s\S]*}/` match combines multiple JSON objects, so output such as `{"warning":"x"}\n{"versions":["1.0"]}` is rejected despite containing valid version data. Scan candidate objects until one matches the `versions` schema, or use balanced JSON extraction, and add a focused test.

[verified]

StellaHuang95:
Please add focused coverage for the direct-list and uninstall timeout behavior introduced by this change.

StellaHuang95 (dismissed):
Approved via Review Center.

src/managers/builtin/commands/availableVersions.ts
```diff
@@ -2,9 +2,24 @@ import type { Pep440Version } from '@renovatebot/pep440';
...
+        typeof parsed !== 'object' ||
+        parsed === null ||
+        !('versions' in parsed) ||
+        !Array.isArray(parsed.versions) ||
+        !parsed.versions.every((version) => typeof version === 'string')
+    ) {
+        throw new Error(`Unexpected package version JSON from ${tool}.`);
+    }
+
+    return parsed.versions;
```
StellaHuang95, line 17: > **Warning** · Non-blocking recommendation

`JSON.parse(output)` still receives the combined stdout/stderr returned by `runPython`, so a non-fatal warning can reject otherwise valid Pip JSON. Separate stdout from stderr for machine-readable commands and add a process-level warning-on-stderr regression test.

[verified]

StellaHuang95 (approved):

rchiodo (approved):

heejaechang (approved):

src/managers/conda/commands/list.ts
```diff
@@ -22,16 +35,24 @@ export class CondaListCommand extends ListCommand {
 
     async execute(executeArgs?: BaseExecuteArgs): Promise<PackageInfo[]> {
         const output = await runCondaExecutable(this.buildCommand(), this.log, executeArgs?.cancellationToken);
```
eleanorjboyd, line 37: **Medium:** `CondaListCommand` inherits the 30-second list timeout, but this call never forwards `this.timeout`; `runCondaExecutable` does not accept one. The same gap exists for `PoetryShowCommand` and `PoetryShowTopLevelCommand`, so stalled list/direct-list subprocesses can run indefinitely despite the stated timeout contract. Please extend the runners to accept the timeout, pass `this.timeout` from these commands, and assert `30_000` in their command tests.

src/managers/poetry/poetryPackageManager.ts
```diff
@@ -214,31 +213,24 @@ export class PoetryPackageManager implements PackageManager, Disposable {
...
-            try {
-                const args = ['remove', ...options.uninstall];
-                this.log.info(`Running: poetry ${args.join(' ')}`);
-                const result = await runPoetry(args, undefined, this.log, token);
-                this.log.info(result);
-            } catch (err) {
-                this.log.error(`Error removing packages with Poetry: ${err}`);
-                throw err;
-            }
+            const removeCmd = new PoetryRemoveCommand({
```
eleanorjboyd, line 218: **Medium:** the command is constructed with the resolved Poetry executable, but add/remove/show execution ignores that injected value because `runPoetry` calls `getPoetry()` again. This leaves two sources of truth and can spawn a different executable if configuration changes between construction and execution. Please either pass the injected executable into `runPoetry`, or make runner-side resolution the sole contract, and cover the actual executable passed to `spawnProcess`.

PR #1696, opened 2026-08-06
Add uv fallback for inline script environments (PEP 723 PR 6/16)

StellaHuang95:
>
> This replaces the earlier closed draft #1652 with the finalized implementation rebased on `main`.

### Roadmap context

This is **PR 6 of 16** in the PEP 723 inline-script roadmap. It extends the PR 5 `create()` happy path with the missing-compatible-interpreter fallback.

|---|---|---|
| | **PR 6: `create()` uv-install fallback** | **this PR** |
| | PR 7: persistence (`get` / `set` + Memento) | separate follow-up |

### Why this PR

PR 5 can create or reuse an inline-script environment when an installed base interpreter already satisfies the script's `requires-python`. It deliberately stops when no compatible interpreter exists.

This PR adds the consent-gated fallback for that case:

1. Re-check installed interpreters after entering a narrow fallback queue.
2. Select a safe Python version satisfying the complete PEP 440 requirement.
3. Ask the user before installing uv or Python.
4. Refresh environment discovery after installation.
5. If discovery is stale or unavailable, resolve and validate uv's returned executable directly.
6. Continue through the existing PR 5 cache create/reuse path.

### What this PR does

**Adds the inline-script fallback to `InlineScriptEnvManager.create()`**

- Keeps the installed-interpreter path unchanged and only enters the fallback when no compatible base exists.
- Skips installation prompts for quick-create/noninteractive calls.
- Serializes only fallback installations, then re-checks discovery before prompting so concurrent compatible requests reuse one installation.
- Retains successfully direct-resolved uv interpreters so queued requests can reuse them even while discovery remains stale.

**Selects a safe uv target from `requires-python`**

- Uses simple safe selectors directly, such as `>=3.13` → `3.13` and `==3.13.1` → `3.13.1`.
- Queries uv's advertised versions for bounded or exclusion-heavy constraints.
- Restricts catalog candidates to default CPython 3 builds satisfying the complete specifier.
- Handles exclusions such as `>=3.13.2,!=3.13.2` without installing the excluded floor.
- Uses an advertised release for bounded ranges rather than fabricating a potentially unavailable patch-zero release.
- Applies full PEP 440 prerelease semantics consistently across discovery, catalog selection, direct resolution, and cache validation.
- Normalizes accepted prerelease aliases (for example, `c1` → `rc1`) before passing a version to uv.

**Extends the uv installer's consent flow**

- Adds an inline-script-specific prompt that shows both the script requirement and selected Python version.
- Sanitizes and caps script-controlled prompt details.
- Validates install selectors before forwarding them to uv.
- If catalog lookup is required and uv is missing, asks for consent to install uv first.
- Re-checks whether a newly installed uv is usable by the current extension host and surfaces the existing restart-required message when needed.

**Handles stale discovery after installation**

- Refreshes environment discovery after uv installs Python.
- Treats refresh/discovery failures as recoverable.
- Resolves the executable returned by uv directly, verifies that it satisfies `requires-python`, and canonicalizes its path before creating the cached environment.

### Examples

| `requires-python` | Fallback behavior |
|---|---|
| `>=3.13` | Request uv's `3.13` selector |
| `==3.13.1` | Request exactly `3.13.1` without requiring a catalog lookup |
| `>=3.11,<3.12` | Choose an advertised compatible `3.11.x` release |
| `>=3.13.2,!=3.13.2` | Skip `3.13.2` and choose a compatible advertised release |
| `>=3.15.0a1,<3.16` | Permit an explicitly requested prerelease |
| `>=3.14,<3.16` | Do not select a prerelease implicitly |

### Safety and concurrency

- No uv or Python installation occurs without explicit user consent.
- Script-derived values cannot inject arbitrary uv arguments.
- Declined, cancelled, or failed installations do not mutate the script-environment cache.
- The fallback queue does not globally serialize environment selection or normal cache creation.
- Existing cross-process cache locking and rollback behavior from PR 5 remains unchanged.

### Tests

Coverage includes:

- simple, exact, bounded, exclusion, and prerelease requirements;
- uv catalog filtering and consented uv bootstrap;
- declined and failed installations;
- refresh and discovery failures with direct resolution;
- simultaneous same- and different-constraint requests;
- quick-create prompt suppression;
- strict post-install and cache compatibility checks;
- prompt sanitization and install-selector validation.

`npm run compile-tests`, `npm run lint`, the full unit suite, and the focused inline-script/uv suites are clean.

### User impact

**No default-path user impact yet.** This completes an internal Phase 2 manager capability. Automatic routing and user-facing entry points arrive in later roadmap PRs.

When those entry points are wired, users whose scripts require an unavailable Python will be able to approve installing a compatible interpreter rather than having environment creation stop.

### DIFF
--- a/src/managers/builtin/inlineScript/envManager.ts
+++ b/src/managers/builtin/inlineScript/envManager.ts
@@ -229,13 +229,13 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
             }),
         );
         let candidates = derivedChecks
-            .filter((candidate) => !candidate.derived)
-            .map((candidate) => candidate.environment)
             .filter(
                 (candidate) =>
-                    !metadata.requiresPython ||
-                    this.matchesInstallConstraint(metadata.requiresPython, candidate.version),
-            );
+                    !candidate.derived &&
+                    (!metadata.requiresPython ||
+                        this.matchesInstallConstraint(metadata.requiresPython, candidate.environment.version)),
+            )
+            .map((candidate) => candidate.environment);
 
         while (candidates.length > 0) {
             const environment = pickCompatibleInterpreter(candidates, undefined);
@@ -266,6 +266,8 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
         const run = this.baseInterpreterInstallationQueue.then(() =>
             this.installAndSelectBaseInterpreterSerially(metadata),
         );
+        // Keep the stored queue tail fulfilled so one failed request does not block later attempts;
+        // the caller still observes the original result through `run`.
         this.baseInterpreterInstallationQueue = run.then(
             () => undefined,
             () => undefined,
--- a/src/managers/builtin/inlineScript/envManager.ts
+++ b/src/managers/builtin/inlineScript/envManager.ts
@@ -21,7 +21,7 @@ import {
     SetEnvironmentScope,
 } from '../../../api';
 import { getErrorMessage } from '../../../common/errors/utils';
-import { computeCacheKey } from '../../../common/inlineScript/cacheKey';
+import { computeCacheKey, normalizeDependency } from '../../../common/inlineScript/cacheKey';
 import {
     META_SCHEMA_VERSION,
     getBaseInterpreterStatus,
@@ -77,6 +77,7 @@ type CacheEntryInspection =
 
 /** Manages extension-owned PEP 723 script environments. */
 export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
+    private readonly pendingSetups = new Map<string, Promise<PythonEnvironment | undefined>>();
     private readonly pendingCreations = new Map<string, Promise<PythonEnvironment | undefined>>();
     private readonly directlyResolvedBaseInterpreters = new Map<string, PythonEnvironment>();
     private baseInterpreterInstallationQueue: Promise<void> = Promise.resolve();
@@ -132,42 +133,80 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
                 return undefined;
             }
 
-            let selectedBase = await this.selectBaseInterpreter(metadata);
-            if (!selectedBase && options?.quickCreate !== true) {
-                selectedBase = await this.installAndSelectBaseInterpreter(metadata);
-            }
-            if (!selectedBase) {
-                this.log.warn(`No compatible Python is available for inline-script environment creation: ${scriptUri.fsPath}.`);
-                return undefined;
-            }
-
-            const cacheKey = computeCacheKey({
-                dependencies: packages,
-                interpreterPath: selectedBase.canonicalPath,
-            });
-            const pending = this.pendingCreations.get(cacheKey);
+            const setupKey = this.getPendingSetupKey(scriptUri, metadata, packages, options);
+            const pending = this.pendingSetups.get(setupKey);
             if (pending) {
                 return await pending;
             }
 
-            const creation = this.createOrReuseEnvironment({
-                cacheKey,
-                packages,
-                metadata,
-                selectedBase,
-            });
-            this.pendingCreations.set(cacheKey, creation);
+            const setup = this.createForScript(scriptUri, metadata, packages, options);
+            this.pendingSetups.set(setupKey, setup);
             try {
-                return await creation;
+                return await setup;
             } finally {
-                if (this.pendingCreations.get(cacheKey) === creation) {
-                    this.pendingCreations.delete(cacheKey);
+                if (this.pendingSetups.get(setupKey) === setup) {
+                    this.pendingSetups.delete(setupKey);
                 }
             }
         } catch (error) {
             this.log.error(`Failed to set up inline-script environment: ${getErrorMessage(error)}`);
             return undefined;
         }
+    }
+
+    private async createForScript(
+        scriptUri: Uri,
+        metadata: InlineScriptMetadata,
+        packages: readonly string[],
+        options?: CreateEnvironmentOptions,
+    ): Promise<PythonEnvironment | undefined> {
+        let selectedBase = await this.selectBaseInterpreter(metadata);
+        if (!selectedBase && options?.quickCreate !== true) {
+            selectedBase = await this.installAndSelectBaseInterpreter(metadata);
+        }
+        if (!selectedBase) {
+            this.log.warn(`No compatible Python is available for inline-script environment creation: ${scriptUri.fsPath}.`);
+            return undefined;
+        }
+
+        const cacheKey = computeCacheKey({
+            dependencies: packages,
+            interpreterPath: selectedBase.canonicalPath,
+        });
+        const pending = this.pendingCreations.get(cacheKey);
+        if (pending) {
+            return await pending;
+        }
+
+        const creation = this.createOrReuseEnvironment({
+            cacheKey,
+            packages,
+            metadata,
+            selectedBase,
+        });
+        this.pendingCreations.set(cacheKey, creation);
+        try {
+            return await creation;
+        } finally {
+            if (this.pendingCreations.get(cacheKey) === creation) {
+                this.pendingCreations.delete(cacheKey);
+            }
+        }
+    }
+
+    private getPendingSetupKey(
+        scriptUri: Uri,
+        metadata: InlineScriptMetadata,
+        packages: readonly string[],
+        options: CreateEnvironmentOptions | undefined,
+    ): string {
+        const normalizedPackages = Array.from(new Set(packages.map(normalizeDependency))).sort();
+        return JSON.stringify([
+            normalizePath(scriptUri.fsPath),
+            metadata.requiresPython?.trim() ?? '',
+            options?.quickCreate === true ? 'quick' : 'interactive',
+            normalizedPackages,
+        ]);
     }
 
     async refresh(_scope: RefreshEnvironmentsScope): Promise<void> {
--- a/src/managers/builtin/inlineScript/envManager.ts
+++ b/src/managers/builtin/inlineScript/envManager.ts
@@ -3,6 +3,7 @@
 
 import * as fs from 'fs-extra';
 import * as path from 'path';
+import { clean as cleanPep440, satisfies as satisfiesPep440 } from '@renovatebot/pep440';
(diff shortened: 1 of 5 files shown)


src/managers/builtin/inlineScript/envManager.ts
```diff
@@ -239,6 +260,191 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
...
+            return satisfiesPep440(version, requiresPython, {
+                prereleases: /(?:(?:a|alpha|b|beta|c|rc|pre|preview)[._-]?\d+|dev[._-]?\d+)/i.test(
+                    requiresPython,
+                ),
+            });
+        } catch (error) {
+            this.log.warn(`Unable to evaluate requires-python '${requiresPython}': ${getErrorMessage(error)}`);
+            return false;
+        }
+    }
```
edvilme, line 441: I would probably extract this function off the class since it is not bound to an instance and can be reused elsewhere

StellaHuang95:

src/managers/builtin/inlineScript/envManager.ts
```diff
@@ -192,12 +196,24 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
...
+            ...globalEnvironments.filter(
+                (environment) =>
+                    BASE_INTERPRETER_MANAGER_IDS.has(environment.envId.managerId) &&
+                    (environment.envId.managerId !== CONDA_MANAGER_ID || environment.name === 'base'),
+            ),
+            ...[...this.directlyResolvedBaseInterpreters.values()].filter(
+                (environment) =>
+                    !metadata.requiresPython ||
+                    this.matchesInstallConstraint(metadata.requiresPython, environment.version),
+            ),
```
edvilme, line 254:
```suggestion
            ...this.directlyResolvedBaseInterpreters.values().filter(
                (environment) =>
                    !metadata.requiresPython ||
                    this.matchesInstallConstraint(metadata.requiresPython, environment.version),
            ),
```

src/managers/builtin/inlineScript/envManager.ts
```diff
@@ -214,10 +230,15 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
         );
         let candidates = derivedChecks
             .filter((candidate) => !candidate.derived)
-            .map((candidate) => candidate.environment);
+            .map((candidate) => candidate.environment)
+            .filter(
+                (candidate) =>
+                    !metadata.requiresPython ||
+                    this.matchesInstallConstraint(metadata.requiresPython, candidate.version),
+            );
```
edvilme: Maybe thes can be combined into something like
```typescript 
let candidates = derivedChecks
    .filter((candidate) => 
        !candidate.derived 
        && (!metadata.requiresPython || this.matchesInstallConstraint(metadata.requiresPython, candidate.version)
    )
    .map((candidate) => candidate.environment)
```

src/managers/builtin/inlineScript/envManager.ts
```diff
@@ -214,10 +230,15 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
...
 
         while (candidates.length > 0) {
-            const environment = pickCompatibleInterpreter(candidates, metadata.requiresPython);
+            const environment = pickCompatibleInterpreter(candidates, undefined);
```
edvilme, line 280:
```suggestion
            const environment = pickCompatibleInterpreter(candidates);
```

src/managers/builtin/inlineScript/envManager.ts
```diff
@@ -239,6 +260,191 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
...
+    private async installAndSelectBaseInterpreter(
+        metadata: InlineScriptMetadata,
+    ): Promise<SelectedBaseInterpreter | undefined> {
+        const run = this.baseInterpreterInstallationQueue.then(() =>
+            this.installAndSelectBaseInterpreterSerially(metadata),
+        );
+        this.baseInterpreterInstallationQueue = run.then(
+            () => undefined,
+            () => undefined,
+        );
```
edvilme, line 313: I am not entirely sure what does this mean

src/managers/builtin/inlineScript/envManager.ts
```diff
@@ -239,6 +260,191 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
...
+        if (requiresPython && !version) {
+            this.log.warn(
+                'Cannot install a Python for this inline script because no compatible install version could be selected.',
+            );
+            return undefined;
+        }
+
+        const installedPath = await this.installPythonAndRefresh(requiresPython, version);
+        if (!installedPath) {
+            return undefined;
```
edvilme, line 337: Maybe we can try catching this?

src/managers/builtin/inlineScript/envManager.ts
```diff
@@ -239,6 +260,191 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
...
+        }
+        if (!selected) {
+            this.log.warn(
+                'Python was installed for an inline script, but no compatible base interpreter was discovered after refreshing environments.',
+            );
+        }
+        return selected;
+    }
+
+    private async selectInstallablePythonVersion(
```
edvilme, line 381: Maybe this is a followup PR, but I think we could create a `PythonVersion` class that can handle extracting lower/upper bounds, testing regexes, comparing, ensuring versions match constraints, etc.

src/managers/builtin/inlineScript/envManager.ts
StellaHuang95: Not sure what's a good way to extract the function further. It currently uses `this.log` for errors, so it is instance bound. The `satisfiesPep440` is already an extracted helper.

src/managers/builtin/inlineScript/envManager.ts
StellaHuang95: `Map.values()` returns a `MapIterator`, not an array, and `filter()` are not included in this project’s current TypeScript setup (TypeScript 5.3.3 and  "lib": ["ES2020"]). I am getting compile errors with this change.

src/managers/builtin/inlineScript/envManager.ts
StellaHuang95: It can be combined but it's mainly for readability concerns, and I don't see any meaningful performance difference. I personally don't prefer compact forms but I don't mind changing it.

src/managers/builtin/inlineScript/envManager.ts
StellaHuang95: This is intentional just to call the helper to validate the already compatible candidates.

src/managers/builtin/inlineScript/envManager.ts
StellaHuang95: will add a comment

src/managers/builtin/inlineScript/envManager.ts
StellaHuang95: `installPythonAndRefreshalready` already catches prompt/install errors and refresh failures, while `undefined` is the expected result for a declined, cancelled, or handled failed installation.

src/managers/builtin/inlineScript/envManager.ts
edvilme: Ahh you're right, sorry

src/managers/builtin/inlineScript/envManager.ts
edvilme: Ahh okay

edvilme (dismissed):
:)

eleanorjboyd (changes requested):
Thanks for adding this fallback. I found two related gaps that need addressing:

1. In `InlineScriptEnvManager.create`, fallback interpreter selection and installation happen before the per-script entry is added to `pendingCreations`. Concurrent `create()` calls for the same script can therefore both display prompts and launch competing uv/Python installations. Please move the fallback work inside the deduplicated operation.

2. The new uv-version-lookup installation path is not effectively exercised. The manager tests stub `ensureUvForInlineScriptVersionLookup()` to return true, while the installer tests cover only dismissing its prompt. Please add coverage for accepted installation success, installation failure, uv remaining unavailable/restart-required, cancellation, and concurrent same-script creation.

StellaHuang95:
> Thanks for adding this fallback. I found two related gaps that need addressing:
> 
> 1. In `InlineScriptEnvManager.create`, fallback interpreter selection and installation happen before the per-script entry is added to `pendingCreations`. Concurrent `create()` calls for the same script can therefore both display prompts and launch competing uv/Python installations. Please move the fallback work inside the deduplicated operation.
> 2. The new uv-version-lookup installation path is not effectively exercised. The manager tests stub `ensureUvForInlineScriptVersionLookup()` to return true, while the installer tests cover only dismissing its prompt. Please add coverage for accepted installation success, installation failure, uv remaining unavailable/restart-required, cancellation, and concurrent same-script creation.


Thanks, both gaps are addressed.
1. I added request-level coalescing before interpreter selection and uv fallback. Identical concurrent requests now share the entire setup result, including successful installation, decline, and failure, so the same script cannot show repeated prompts or retry the same installation. The key includes the normalized script path,  requires-python , effective package set, and quick-create mode, so requests with different requirements/options are not incorrectly combined. The existing cache-key coalescing remains in place for equivalent final environments across different scripts.
2. I added direct coverage for the uv version-lookup bootstrap path: accepted installation success, installation failure, uv remaining unavailable with restart guidance, and cancelled installation. I also added concurrent same-script tests for both successful and declined fallback.

eleanorjboyd (approved):

src/managers/builtin/inlineScript/envManager.ts
```diff
@@ -78,7 +77,10 @@ type CacheEntryInspection =
 
 /** Manages extension-owned PEP 723 script environments. */
 export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
+    private readonly pendingSetups = new Map<string, Promise<PythonEnvironment | undefined>>();
     private readonly pendingCreations = new Map<string, Promise<PythonEnvironment | undefined>>();
+    private readonly directlyResolvedBaseInterpreters = new Map<string, PythonEnvironment>();
+    private baseInterpreterInstallationQueue: Promise<void> = Promise.resolve();
```
edvilme, line 83: What is this used for?

edvilme (approved):

PR #1697, opened 2026-08-06
Add inline script environment persistence (PEP 723 PR 7/16)

StellaHuang95:
>
> This replaces the earlier closed draft #1653 with the finalized implementation rebased on `main`.

### Roadmap context

This is **PR 7 of 16** in the PEP 723 inline-script roadmap. It adds durable per-script environment associations to the internal manager.

|---|---|---|
| | PR 6: `create()` uv-install fallback | open (#1696) |
| | **PR 7: persistence (`get` / `set` + Memento)** | **this PR** |
| | PR 8: activation-time discovery | follow-up |
| | PR 9: route PEP 723 scripts to the inline manager | follow-up |

### Why this PR

PR 5 can build or reuse an inline-script environment, but the manager does not remember that the resulting environment belongs to a particular script. After an extension-host restart, the in-memory association is gone.

This PR implements the persistence portion of Q4 in the design:

- maintain an independent environment association for each script;
- persist script-to-environment executable paths in workspace Memento;
- lazily and safely rehydrate those associations;
- re-check current `requires-python` metadata before returning an environment;
- report changes so the central environment API can update its last-known state.

### What this PR does

**Implements per-script `set()`**

- Accepts one or more local `file:` URIs and rejects invalid or mixed scopes atomically.
- Validates that selected environments are owned inline-script cache entries.
- Persists a normalized script path → environment executable path mapping under a dedicated Memento key.
- Supports assigning and unassigning individual scripts or batches.
- Updates in-memory state and emits `onDidChangeEnvironment` only for effective changes.
- Leaves the existing `create()` behavior separate: creation alone does not implicitly establish a persisted association.

**Implements per-script `get()`**

- Reads current PEP 723 metadata before returning an association.
- Keeps unreadable or temporarily invalid script metadata from destructively clearing state.
- Returns an in-memory association when valid.
- Lazily reconstructs persisted environments after restart instead of resolving every script during activation.
- Re-checks `requires-python` against the reconstructed Python version before returning it.

**Safely rehydrates persisted associations**

- Requires an absolute executable path.
- Preserves associations while their cache entry is locked or being created.
- Verifies that the executable exists and is a regular file.
- Resolves it into a `PythonEnvironment` and confirms that it belongs to the expected extension-owned cache entry.
- Removes only definitively stale associations; transient filesystem or resolver failures remain retryable.
- Emits a change event when a slow rehydration eventually succeeds, including when the public API's initial one-second wait has already elapsed.

**Validates warm in-memory associations**

- Periodically revalidates cached associations without performing full resolution on every lookup.
- Detects executables deleted while VS Code remains open.
- Detects an environment rebuilt at the same cache path with a different Python version.
- Preserves busy/locked entries instead of misclassifying them as stale.
- Coalesces simultaneous validations for the same script.
- Retains the existing environment object when resolution produces only a new generated ID for the same Python, avoiding false changes and duplicate ID-keyed resources.

**Protects persistence and selection from races**

- Serializes Memento read-modify-write operations so concurrent script selections cannot lose one another.
- Uses per-script association revisions so an older rehydration cannot overwrite a newer selection or unset.
- Removes stale persisted values conditionally, only if the inspected path is still current.
- Keeps failed persistence writes from changing in-memory state or emitting success-shaped events.
- Does not globally serialize unrelated environment operations.

**Updates central active-environment tracking**

- Keys inline-script selections by normalized script path rather than containing project, so two scripts in one workspace can retain different environments.
- Uses per-scope revisions and manager identity checks so slow refreshes cannot overwrite newer selections.
- Ensures failed selections and failed refreshes do not discard a valid in-flight refresh.
- Groups same-manager batch unsets and calls the manager once with the complete URI array.
- Updates central cache entries and events only after the manager operation succeeds.
- Attributes inline-script change events to the script URI rather than the containing project URI.

### Example

Given two scripts in the same workspace:

```text
tools/report.py → Python 3.12 inline environment
tools/import.py → Python 3.13 inline environment
```

PR 7 stores and retrieves those associations independently. Selecting the environment for `import.py` does not overwrite the last-known environment for `report.py`.

After restart:

```text
get(report.py)
→ read persisted executable
→ verify cache ownership and current metadata
→ resolve environment
→ cache and return it
```

If `report.py` later changes from `requires-python = ">=3.11"` to `">=3.13"`, its persisted Python 3.12 environment is no longer returned as compatible.

### Persistence and failure semantics

| Condition | Behavior |
|---|---|
| Executable exists and cache ownership is valid | Rehydrate and return |
| Cache entry is locked/in progress | Preserve association; retry later |
| Resolver fails transiently | Preserve association; retry later |
| Executable is definitively missing and unlocked | Remove stale association and notify |
| A newer selection wins during rehydration | Discard the stale result |
| Memento write fails | Keep previous in-memory/persisted selection and propagate the error |

### Tests

Coverage includes:

- assign, retrieve, unset, and batch persistence;
- restart-time lazy rehydration and delayed success events;
- metadata compatibility changes;
- missing, malformed, unowned, busy, and transient cache states;
- warm deletion and same-path rebuild detection;
- concurrent persistence, rehydration, validation, selection, and unset races;
- failed Memento writes;
- strict URI-scope validation;
- independent same-project script selections;
- stale and failed central refresh ordering;
- atomic same-manager batch unsets.

`npm run compile-tests`, `npm run lint`, the full unit suite, and the focused persistence/central-manager suites are clean.

### Performance

- Rehydration is lazy rather than activation-blocking.
- Warm associations are cached and validation is throttled.
- Same-script rehydration and validation work is coalesced.
- Queues cover only shared persistence and mutation ordering; unrelated script reads and environment-manager operations remain independent.

### User impact

**No default-path user impact yet.** This completes an internal Phase 2 manager capability. Automatic routing and user-facing entry points arrive in later roadmap PRs.

Once routing is wired, script-specific selections will survive extension-host restarts and remain independent even for multiple scripts in the same workspace.

### Merge order

The core persistence behavior depends on the merged manager skeleton (#1610). This branch is rebased on current `main`; PR 8 and PR 9 build on this capability.

### DIFF
--- a/src/features/envManagers.ts
+++ b/src/features/envManagers.ts
@@ -65,6 +65,7 @@ export class PythonEnvironmentManagers implements EnvironmentManagers {
      */
     private readonly _activeSelection = new Map<string, PythonEnvironment | undefined>();
     private readonly _selectionRevisions = new Map<string, number>();
+    private readonly _selectionOperationCounters = new Map<string, number>();
 
     private _onDidChangeEnvironmentManager = new EventEmitter<DidChangeEnvironmentManagerEventArgs>();
     private _onDidChangePackageManager = new EventEmitter<DidChangePackageManagerEventArgs>();
@@ -117,7 +118,7 @@ export class PythonEnvironmentManagers implements EnvironmentManagers {
                 );
             }),
             mgr.onDidChangeEnvironment((e: DidChangeEnvironmentEventArgs) => {
-                if (e.old?.envId.id === e.new?.envId.id) {
+                if (this.isSameEnvironment(e.old, e.new)) {
                     return;
                 }
 
@@ -362,11 +363,12 @@ export class PythonEnvironmentManagers implements EnvironmentManagers {
         }
         const project = scope ? this.pm.get(scope) : undefined;
         const key = this.getActiveSelectionKey(scope, manager, project);
+        const operation = this.beginSelectionOperation(key);
+        const inlineClearOperation =
+            scope instanceof Uri && manager.id !== INLINE_SCRIPT_MANAGER_ID
+                ? this.beginSelectionOperation(this.getInlineScriptSelectionKey(scope))
+                : undefined;
         await manager.set(scope, environment);
-        if (scope instanceof Uri) {
-            this.clearInlineActiveSelection(scope, manager);
-        }
-        this.bumpSelectionRevision(key);
 
         // Only persist to settings when explicitly requested
         if (shouldPersistSettings && scope) {
@@ -393,8 +395,14 @@ export class PythonEnvironmentManagers implements EnvironmentManagers {
             );
         }
 
+        if (scope instanceof Uri) {
+            this.clearInlineActiveSelection(scope, manager, inlineClearOperation);
+        }
+        if (!this.commitSelectionOperation(key, operation)) {
+            return;
+        }
         const oldEnv = this._activeSelection.get(key);
-        if (oldEnv?.envId.id !== environment?.envId.id) {
+        if (!this.isSameEnvironment(oldEnv, environment)) {
             this._activeSelection.set(key, environment);
             await new Promise<void>((resolve, reject) => {
                 setImmediate(() => {
@@ -443,33 +451,35 @@ export class PythonEnvironmentManagers implements EnvironmentManagers {
             const settings: EditAllManagerSettings[] = [];
             const events: DidChangeEnvironmentEventArgs[] = [];
             if (Array.isArray(scope) && scope.every((s) => s instanceof Uri)) {
+                const selections = scope.map((uri) => this.beginPendingSelection(uri, manager));
                 await manager.set(scope, environment);
-                scope.forEach((uri) => {
-                    this.clearInlineActiveSelection(uri, manager);
-                    const project = this.pm.get(uri);
-                    this.bumpSelectionRevision(this.getActiveSelectionKey(uri, manager, project));
-                });
-                scope.forEach((uri) => {
-                    const m = this.getEnvironmentManager(uri);
-                    const project = this.pm.get(uri);
+                selections.forEach((selection) => {
+                    const m = this.getEnvironmentManager(selection.scope);
                     // Always add settings when persisting, OR when manager differs
                     if (
                         (shouldPersistSettings || manager.id !== m?.id) &&
-                        this.canPersistManagerSettingForScope(uri, manager, project)
+                        this.canPersistManagerSettingForScope(selection.scope, manager, selection.project)
                     ) {
                         settings.push({
-                            project,
+                            project: selection.project,
                             envManager: manager.id,
                             packageManager: manager.preferredPackageManagerId,
                         });
                     }
-
-                    const key = this.getActiveSelectionKey(uri, manager, project);
-                    const oldEnv = this._activeSelection.get(key);
-                    if (oldEnv?.envId.id !== environment?.envId.id) {
-                        this._activeSelection.set(key, environment);
+                });
+                if (shouldPersistSettings) {
+                    await setAllManagerSettings(settings);
+                }
+                selections.forEach((selection) => {
+                    this.clearInlineActiveSelection(selection.scope, manager, selection.inlineClearOperation);
+                    if (!this.commitSelectionOperation(selection.key, selection.operation)) {
+                        return;
+                    }
+                    const oldEnv = this._activeSelection.get(selection.key);
+                    if (!this.isSameEnvironment(oldEnv, environment)) {
+                        this._activeSelection.set(selection.key, environment);
                         events.push({
-                            uri: this.getActiveSelectionUri(uri, manager, project),
+                            uri: this.getActiveSelectionUri(selection.scope, manager, selection.project),
                             new: environment,
                             old: oldEnv,
                         });
@@ -477,8 +487,8 @@ export class PythonEnvironmentManagers implements EnvironmentManagers {
                 });
             } else if (typeof scope === 'string' && scope === 'global') {
                 const m = this.getEnvironmentManager(undefined);
+                const operation = this.beginSelectionOperation('global');
                 await manager.set(undefined, environment);
-                this.bumpSelectionRevision('global');
                 // Always add settings when persisting, OR when manager differs
                 if (shouldPersistSettings || manager.id !== m?.id) {
                     settings.push({
@@ -488,15 +498,16 @@ export class PythonEnvironmentManagers implements EnvironmentManagers {
                     });
                 }
 
-                const oldEnv = this._activeSelection.get('global');
-                if (oldEnv?.envId.id !== environment?.envId.id) {
-                    this._activeSelection.set('global', environment);
-                    events.push({ uri: undefined, new: environment, old: oldEnv });
+                if (shouldPersistSettings) {
+                    await setAllManagerSettings(settings);
                 }
-            }
-            // Only persist to settings when explicitly requested
-            if (shouldPersistSettings) {
-                await setAllManagerSettings(settings);
+                if (this.commitSelectionOperation('global', operation)) {
+                    const oldEnv = this._activeSelection.get('global');
+                    if (!this.isSameEnvironment(oldEnv, environment)) {
+                        this._activeSelection.set('global', environment);
+                        events.push({ uri: undefined, new: environment, old: oldEnv });
+                    }
+                }
             }
             if (events.length > 0) {
                 await new Promise<void>((resolve, reject) => {
@@ -521,21 +532,19 @@ export class PythonEnvironmentManagers implements EnvironmentManagers {
                     }
                 });
                 for (const [manager, uris] of groupedScopes) {
+                    const selections = uris.map((uri) => this.beginPendingSelection(uri, manager));
                     await manager.set(uris);
-                    uris.forEach((uri) => {
-                        const project = this.pm.get(uri);
-                        this.bumpSelectionRevision(this.getActiveSelectionKey(uri, manager, project));
-                    });
                     await Promise.all(
-                        uris.map(async (uri) => {
-                            const project = this.pm.get(uri);
-                            const newEnv = await manager.get(uri);
-                            const key = this.getActiveSelectionKey(uri, manager, project);
-                            const oldEnv = this._activeSelection.get(key);
-                            if (oldEnv?.envId.id !== newEnv?.envId.id) {
-                                this._activeSelection.set(key, newEnv);
+                        selections.map(async (selection) => {
+                            const newEnv = await manager.get(selection.scope);
+                            if (!this.commitSelectionOperation(selection.key, selection.operation)) {
+                                return;
+                            }
+                            const oldEnv = this._activeSelection.get(selection.key);
+                            if (!this.isSameEnvironment(oldEnv, newEnv)) {
+                                this._activeSelection.set(selection.key, newEnv);
                                 events.push({
-                                    uri: this.getActiveSelectionUri(uri, manager, project),
+                                    uri: this.getActiveSelectionUri(selection.scope, manager, selection.project),
                                     new: newEnv,
                                     old: oldEnv,
                                 });
@@ -546,13 +555,15 @@ export class PythonEnvironmentManagers implements EnvironmentManagers {
             } else if (typeof scope === 'string' && scope === 'global') {
                 const manager = this.getEnvironmentManager(undefined);
                 if (manager) {
+                    const operation = this.beginSelectionOperation('global');
                     await manager.set(undefined);
-                    this.bumpSelectionRevision('global');
                     const newEnv = await manager.get(undefined);
-                    const oldEnv = this._activeSelection.get('global');
-                    if (oldEnv?.envId.id !== newEnv?.envId.id) {
-                        this._activeSelection.set('global', newEnv);
-                        events.push({ uri: undefined, new: newEnv, old: oldEnv });
+                    if (this.commitSelectionOperation('global', operation)) {
+                        const oldEnv = this._activeSelection.get('global');
+                        if (!this.isSameEnvironment(oldEnv, newEnv)) {
+                            this._activeSelection.set('global', newEnv);
+                            events.push({ uri: undefined, new: newEnv, old: oldEnv });
+                        }
                     }
                 }
             }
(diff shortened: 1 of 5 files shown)


eleanorjboyd:
AI agent review findings:

1. Selecting an inline-script environment currently uses the normal project settings persistence path, which can save `inline-script` as the containing project's default manager. Subsequent ordinary-file and unassociated-script lookups then route to `InlineScriptEnvManager` and can return no environment. Conversely, loose-script associations have no persisted manager routing and become unreachable after restart. This routing needs to remain per-script rather than changing the project-wide manager setting.

2. `validateCachedAssociation()` captures the association revision only after awaiting the busy and filesystem checks, while still validating the earlier cached environment. An explicit selection during either await can therefore update the revision, after which the stale cached result is treated as current and may overwrite the newer selection. Please capture the revision with the original cached read, before any await, and add a regression test for that interleaving.

I validated the PR with `npm run compile-tests`, `npm run compile`, and the focused inline-script/last-known-environment tests (78 passing, 1 pending).

_Reviewed and posted by Eleanor's AI agent._

StellaHuang95:
> AI agent review findings:
> 
> 1. Selecting an inline-script environment currently uses the normal project settings persistence path, which can save `inline-script` as the containing project's default manager. Subsequent ordinary-file and unassociated-script lookups then route to `InlineScriptEnvManager` and can return no environment. Conversely, loose-script associations have no persisted manager routing and become unreachable after restart. This routing needs to remain per-script rather than changing the project-wide manager setting.
> 2. `validateCachedAssociation()` captures the association revision only after awaiting the busy and filesystem checks, while still validating the earlier cached environment. An explicit selection during either await can therefore update the revision, after which the stale cached result is treated as current and may overwrite the newer selection. Please capture the revision with the original cached read, before any await, and add a regression test for that interleaving.
> 
> I validated the PR with `npm run compile-tests`, `npm run compile`, and the focused inline-script/last-known-environment tests (78 passing, 1 pending).
> 
> _Reviewed and posted by Eleanor's AI agent._

Thanks @eleanorjboyd, both findings are valid.
1. Per-script routing/settings:  pm.get(scriptUri)  can currently return the containing project, causing  setEnvironment()  to persist  inline-script  as that whole project’s manager. I’ll update both single and batch selection paths so an inline-script manager setting is persisted only when the resolved project URI exactly matches the script URI. Otherwise, only the inline manager’s per-script Memento association will be updated.
The loose-script restart-routing gap is real, but the roadmap intentionally separates that work: PR7 persists and rehydrates the association inside the manager, PR9 routes known scripts to that manager, and PR10 registers scripts as individual projects so settings can be persisted per script. I’ll prevent the incorrect project-wide write in this PR without pulling all PR9/PR10 routing into it.
2. Cached-association revision: Agreed. The revision is captured too late, after asynchronous lock/filesystem checks, so validation of an old cached environment can accidentally adopt a newer selection’s revision. I’ll capture the revision together with the cached environment before the first  await , pass it through validation and stale cleanup, and add a regression test where a new selection occurs while validation is paused.

eleanorjboyd:
will give a thumbs up to after the prior one merges as I assume that will create merge conflicts

eleanorjboyd (dismissed):

rchiodo:

src/managers/builtin/inlineScript/envManager.ts
```diff
@@ -234,6 +260,543 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
...
+                this.log.warn(message);
+                throw new Error(message);
+            }
+            environmentPath = environment.environmentPath.fsPath;
+        }
+
+        const updates: {
+            readonly uri: Uri;
+            readonly scriptPath: string;
+            readonly before: PythonEnvironment | undefined;
```
rchiodo: > **Issue** · Please address or respond

The long five-field `updates` shape and repeated script-reference shape violate `typescript-named-types-at-bottom`; `PersistedAssociationChange` is also declared near the top. Extract named types such as `ScriptReference` and `PendingScriptUpdate`, and place all newly introduced named types at the bottom of the file.

[verified]

src/features/envManagers.ts
```diff
@@ -354,20 +388,19 @@ export class PythonEnvironmentManagers implements EnvironmentManagers {
...
-                    `packageManager=${this.getPackageManager(environment)?.id ?? 'UNDEFINED'}, ` +
-                    `settingsPersisted=${!!(project && this.getPackageManager(environment))}`,
+                    `packageManager=${packageManager?.id ?? 'UNDEFINED'}, ` +
+                    `settingsPersisted=${!!canPersistSettings}`,
             );
         }
 
-        const key = project ? project.uri.toString() : 'global';
         const oldEnv = this._activeSelection.get(key);
         if (oldEnv?.envId.id !== environment?.envId.id) {
```
rchiodo: > **Warning** · Non-blocking recommendation

A selection can publish after a newer selection when it stalls in `setAllManagerSettings`, because the revision is bumped before that await but never checked before updating `_activeSelection`. Capture the revision and verify it immediately before publishing, with a regression test where the first settings write resolves last.

[verified]

rchiodo: > **Warning** · Non-blocking recommendation

Central state compares only `envId.id`, although the inline manager and its tests establish that equal IDs can refer to different executable paths. Reuse one inline-environment identity helper at every central publication site and add central-manager coverage for same-ID, different-path selections.

[verified]

src/managers/builtin/inlineScript/envManager.ts
```diff
@@ -234,6 +260,543 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
...
+        expectedEnvironmentPath: string,
+        revision: number,
+        scriptUri?: Uri,
+    ): Promise<void> {
+        await this.enqueueSelection(async () => {
+            if (!this.isCurrentAssociationRevision(scriptPath, revision)) {
+                return;
+            }
+            try {
+                await this.updatePersistedAssociations([{ scriptPath, expectedEnvironmentPath }]);
```
rchiodo, line 682: > **Warning** · Non-blocking recommendation

Persisted deletion normalizes executable paths, but in-memory cleanup requires exact string equality; casing differences on Windows can therefore leave a stale cached environment. Normalize both paths in this condition and add a case-variant stale-cleanup test.

[verified]

src/managers/builtin/inlineScript/envManager.ts
```diff
@@ -217,12 +243,12 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
         return [];
```
rchiodo, line 235: > **Warning** · Non-blocking recommendation

Every script selection shares one `selectionQueue`, serializing ownership checks and association reads for unrelated scripts even though persistence has its own queue. While this simplifies atomicity, use per-script sequencing with deterministic multi-key acquisition if independent-operation performance is a requirement.

[verified]

src/managers/builtin/inlineScript/envManager.ts
```diff
@@ -81,6 +100,13 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
     private readonly pendingCreations = new Map<string, Promise<PythonEnvironment | undefined>>();
     private readonly directlyResolvedBaseInterpreters = new Map<string, PythonEnvironment>();
     private baseInterpreterInstallationQueue: Promise<void> = Promise.resolve();
+    private readonly pendingRehydrations = new Map<string, Promise<PythonEnvironment | undefined>>();
+    private readonly fsPathToEnv = new Map<string, PythonEnvironment>();
+    private readonly fsPathToPersistedEnvPath = new Map<string, string>();
+    private readonly cachedAssociationValidatedAt = new Map<string, number>();
```
rchiodo, line 98: > **Warning** · Non-blocking recommendation

This manager now owns persistence schema, transactional queues, revisions, rehydration, filesystem validation, ownership checks, compatibility, and event publication. Extract a focused association repository/coordinator before later roadmap work expands this lifecycle further.

[verified]

src/managers/builtin/inlineScript/envManager.ts
```diff
@@ -234,6 +260,543 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
...
+            readonly shouldNotify: boolean;
+        }[] = [];
+        for (const script of scripts) {
+            const before = await this.getAssociationForMutation(script.scriptPath);
+            const hadPersistedAssociation = this.fsPathToPersistedEnvPath.has(script.scriptPath);
+            const hasSamePersistedEnvironment =
+                environmentPath !== undefined &&
+                normalizePath(this.fsPathToPersistedEnvPath.get(script.scriptPath) ?? '') ===
+                    normalizePath(environmentPath);
+            const needsPersistence = environment ? !hasSamePersistedEnvironment : hadPersistedAssociation;
```
rchiodo, line 280: > **Info** · Optional note

Unsetting a cold persisted association can emit an event whose `old` and `new` values are both `undefined`. Clarify the event contract and either resolve the prior value, emit an explicit invalidation shape, or add an integration assertion proving consumers support this payload.

[verified]

src/managers/builtin/inlineScript/envManager.ts
StellaHuang95: Addressed in `1573f35`. I added `ScriptReference` and `PendingScriptUpdate`, reused them instead of the repeated anonymous shapes, and moved the newly introduced association types to the bottom of the file.

src/features/envManagers.ts
StellaHuang95: Addressed in `1573f35`. Central selection operations now use per-scope operation tokens and commit only if no newer successful operation has published. The check is applied across single, batch, global, unset, and refresh publication paths. I also made no-op refreshes revision-neutral so they cannot defeat a pending manager switch, and added regressions where the first settings write resolves after a newer selection.

StellaHuang95: Addressed in `1573f35`. Central publication now uses one identity helper everywhere: inline environments compare manager ID plus normalized environment path, while other managers compare manager ID plus environment ID. Added central coverage for equal IDs at different inline paths.

src/managers/builtin/inlineScript/envManager.ts
StellaHuang95: Addressed in `1573f35`. The in-memory cleanup condition now normalizes both executable paths before comparison. Added a Windows regression that rehydrates a case-variant persisted path, deletes the warm executable, and verifies all persisted and in-memory state is cleared.

src/managers/builtin/inlineScript/envManager.ts
StellaHuang95: Thanks, this is a valid tradeoff. I am keeping the shared `selectionQueue` in PR7 because selections are explicit/infrequent, persistence still requires globally ordered read-modify-write operations, and the queue keeps batch atomicity and stale repair straightforward. Per-script sequencing with deterministic multi-key acquisition is better handled as a measured performance follow-up rather than adding locking complexity here.

src/managers/builtin/inlineScript/envManager.ts
StellaHuang95: I agree with the direction, but I am deferring the extraction from PR7. Moving persistence, revisions, rehydration, and event coordination into a repository would add substantial structural churn to an already large correctness-focused PR without changing behavior. A focused follow-up before later lifecycle work expands the manager further would be the safer seam.

src/managers/builtin/inlineScript/envManager.ts
StellaHuang95: Thanks. I am leaving this payload unchanged in PR7 and treating the cold-unset case as an invalidation signal: there is a persisted association to clear, but no materialized old `PythonEnvironment`. Current consumers refresh by URI and treat `old`/`new` as optional. I agree the contract should be made explicit when PR9 adds routing-cache consumers; that is the right point to decide whether to retain an invalidation signal, suppress it, or introduce a dedicated shape.

src/features/envManagers.ts
```diff
@@ -593,24 +650,128 @@ export class PythonEnvironmentManagers implements EnvironmentManagers {
...
+    }
+
+    private isSameEnvironment(
+        first: PythonEnvironment | undefined,
+        second: PythonEnvironment | undefined,
+    ): boolean {
+        if (first === second) {
+            return true;
+        }
+        if (!first || !second || first.envId.managerId !== second.envId.managerId) {
```
rchiodo, line 758: > **Issue** · Please address or respond

`isSameEnvironment` treats inline environments at the same executable path as identical even when their Python version changed. A same-path rebuild event is therefore discarded, leaving `_activeSelection` stale. Include version-relevant identity while still ignoring generated-ID-only changes.

src/features/envManagers.ts
```diff
@@ -467,49 +522,51 @@ export class PythonEnvironmentManagers implements EnvironmentManagers {
...
                         const oldEnv = this._activeSelection.get('global');
-                        if (oldEnv?.envId.id !== newEnv?.envId.id) {
+                        if (!this.isSameEnvironment(oldEnv, newEnv)) {
                             this._activeSelection.set('global', newEnv);
                             events.push({ uri: undefined, new: newEnv, old: oldEnv });
                         }
-                    };
-                    promises.push(setAndAddEvent());
+                    }
                 }
```
rchiodo, line 570: > **Issue** · Please address or respond

A completed manager group updates `_activeSelection`, but a later group rejection exits before the accumulated active-environment events are fired. Publish completed groups consistently, or defer central mutations until the batch publication semantics are settled.

src/managers/builtin/inlineScript/envManager.ts
```diff
@@ -234,6 +252,538 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
...
+        this.cachedAssociationValidatedAt.set(scriptPath, Date.now());
+        this._onDidChangeEnvironment.fire({ uri: scriptUri, old: undefined, new: resolved });
+        return resolved;
+    }
+
+    private async inspectAssociationOwnership(environment: PythonEnvironment): Promise<CacheEnvironmentInspection> {
+        if (environment.envId.managerId !== INLINE_SCRIPT_MANAGER_ID || !path.isAbsolute(environment.sysPrefix)) {
+            return 'uncertain';
+        }
+        const cacheRoot = getScriptEnvCacheRoot(this.globalStorageUri);
```
rchiodo, line 626: > **Warning** · Non-blocking recommendation

Cold rehydration calls the resolver and ownership inspection outside error handling. A transient rejection can escape from `get()` instead of returning `undefined` while preserving the association for retry. Catch and log those transient failures.

src/features/envManagers.ts
StellaHuang95: Addressed in `2e6ce79`. Inline central identity now compares normalized environment path plus Python version, while still ignoring generated-ID-only changes. Added coverage proving same-path version rebuilds publish and same-path/same-version ID regeneration does not.

src/features/envManagers.ts
StellaHuang95: Addressed in `2e6ce79`. Multi-manager unsets now publish each successfully completed manager group immediately, before processing the next group. Added a regression where the first group clears and publishes successfully while the second group rejects.

src/managers/builtin/inlineScript/envManager.ts
StellaHuang95: Addressed in `2e6ce79`. Cold rehydration now catches and logs both resolver and ownership-inspection rejections, returns `undefined`, and preserves the persisted association for retry. Added direct rejection-then-success tests for both stages.

src/managers/builtin/inlineScript/envManager.ts
```diff
@@ -234,6 +252,554 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
...
+                normalizePath(this.fsPathToPersistedEnvPath.get(script.scriptPath) ?? '') ===
+                    normalizePath(environmentPath);
+            const needsPersistence = environment ? !hasSamePersistedEnvironment : hadPersistedAssociation;
+            const shouldNotify =
+                (!this.isSameEnvironment(before, environment) && !hasSamePersistedEnvironment) ||
+                (!environment && hadPersistedAssociation);
+            const hasPendingRehydration = this.pendingRehydrations.has(script.scriptPath);
+            const cached = this.fsPathToEnv.get(script.scriptPath);
+            const needsMemoryUpdate = environment ? cached !== environment : cached !== undefined;
+            if (needsPersistence || shouldNotify || hasPendingRehydration || needsMemoryUpdate) {
```
rchiodo, line 287: > **Warning** · Non-blocking recommendation

`isSameEnvironment` and `hasSamePersistedEnvironment` compare only the executable path. Selecting a rebuilt environment at the same path with a different Python version updates the cached environment without emitting a manager change event. Include version in manager-level identity while continuing to suppress generated-ID-only changes, and add direct manager-subscriber coverage.

rchiodo (approved):
Approved via Review Center.

src/features/envManagers.ts
```diff
@@ -354,20 +390,25 @@ export class PythonEnvironmentManagers implements EnvironmentManagers {
...
                     `project=${project?.uri?.toString() ?? 'none'}, ` +
-                    `packageManager=${this.getPackageManager(environment)?.id ?? 'UNDEFINED'}, ` +
-                    `settingsPersisted=${!!(project && this.getPackageManager(environment))}`,
+                    `packageManager=${packageManager?.id ?? 'UNDEFINED'}, ` +
+                    `settingsPersisted=${!!canPersistSettings}`,
             );
         }
 
-        const key = project ? project.uri.toString() : 'global';
+        if (scope instanceof Uri) {
```
rchiodo, line 398: > **Issue** · Please address or respond

Switching away from the inline manager only removes central in-memory routing; it does not unset the inline manager's persisted script association. A later restart or inline discovery can therefore resurrect the superseded selection. Clear the durable inline association after the replacement manager succeeds.

edvilme (approved):

PR #1702, opened 2026-08-07
Add engagement-based Marketplace feedback prompt

StellaHuang95:
## Summary

Adds a one-time prompt inviting established Python Environments users to leave a review on the Visual Studio Marketplace.

The prompt uses neutral engagement and successful-workflow signals so it appears after repeated, intentional use rather than during initial setup or passive extension activation.

The notification contains one action: **Review on Marketplace**.

## Prompt eligibility

The prompt is eligible only when all of the following conditions are met:

- The extension was first observed at least 14 days ago.
- The extension was active on at least 9 distinct days.
- The user successfully selected or created an environment on at least 2 distinct days.
- A successful selection or creation occurred during the current session.
- A 60-second quiet period has elapsed since the latest successful action.
- The VS Code window is focused.
- VS Code is running as a desktop application.
- `telemetry.feedback.enabled` is enabled.

Automatic discovery, initial environment resolution, cancelled pickers, and failed operations do not count as successful actions.

## End-to-end flow

1. Extension activation records the current UTC date as an active day.
2. Explicit environment selection and creation commands return the environment when they complete successfully.
3. A successful result records the current date as a successful-action day.
4. Once the age, active-day, and successful-action-day requirements are met, a 60-second quiet timer starts.
5. Eligibility and window state are checked again when the timer completes.
6. The extension atomically claims a marker in its global storage before showing the notification.
7. The user can select **Review on Marketplace** to open the extension's Ratings and Reviews page.
8. Dismissing the notification performs no external action.

## Once-only behavior

The prompt is claimed before it is displayed by creating an exclusive marker file in extension global storage.

This prevents multiple VS Code windows from displaying the prompt concurrently. The marker remains after either selecting the Marketplace action or dismissing the notification, so the prompt is not shown again.

If display conditions change while the marker is being claimed, such as the window losing focus, the claim is released so the prompt can be reconsidered later.

## Data and privacy

The implementation stores only:

- The first-observed timestamp.
- Up to 30 distinct active dates.
- Up to 30 distinct dates with successful selection or creation.

It does not store environment names, environment paths, project information, commands, or package information. No new telemetry is emitted by the prompt.

## Scope

The implementation is intentionally limited to five files:

- Localized prompt strings.
- Extension command wiring.
- Environment-selection success reporting.
- The feedback prompt service.
- Focused unit tests.

Terminal, task, package-management, discovery, and environment-manager internals are unchanged.

## Testing

- TypeScript compilation passes.
- ESLint passes.
- Production webpack packaging passes.
- Feedback prompt unit tests: 7 passing.
- Full unit suite: 1,755 passing and 8 pending.
- The remaining 3 failures are existing unrelated Venv/Conda test failures.

### DIFF
--- a/src/features/feedback/feedbackPromptService.ts
+++ b/src/features/feedback/feedbackPromptService.ts
@@ -232,6 +232,10 @@ export class FeedbackPromptService implements Disposable {
     }
 }
 
+/**
+ * Atomically creates the marker only when it does not already exist.
+ * Success acquires the prompt claim; EEXIST means another extension host already claimed or displayed it.
+ */
 async function claimPrompt(globalStoragePath: string): Promise<boolean> {
     await fs.mkdir(globalStoragePath, { recursive: true });
     try {
@@ -246,6 +250,8 @@ async function claimPrompt(globalStoragePath: string): Promise<boolean> {
     }
 }
 
+// Remove the claim only when display conditions change before the prompt is shown.
+// After the prompt is displayed or dismissed, the marker remains to enforce once-only behavior.
 async function releasePromptClaim(globalStoragePath: string): Promise<void> {
     try {
         await fs.unlink(path.join(globalStoragePath, PROMPT_CLAIM_FILE));
--- a/src/features/feedback/feedbackPromptService.ts
+++ b/src/features/feedback/feedbackPromptService.ts
@@ -1,0 +1,257 @@
+import { promises as fs } from 'fs';
+import * as path from 'path';
+import { Disposable, env, Memento, UIKind, window } from 'vscode';
+import { launchBrowser } from '../../common/env.apis';
+import { FeedbackStrings } from '../../common/localize';
+import { traceError } from '../../common/logging';
+import { showInformationMessage } from '../../common/window.apis';
+import { getConfiguration } from '../../common/workspace.apis';
+
+export const FEEDBACK_PROMPT_STATE_KEY = 'python-envs:feedbackPrompt:v2';
+
+const MARKETPLACE_REVIEWS_URL =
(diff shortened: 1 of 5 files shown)


StellaHuang95:
The prompt looks like this. image

src/features/feedback/feedbackPromptService.ts
```diff
@@ -0,0 +1,257 @@
...
+}
+
+async function claimPrompt(globalStoragePath: string): Promise<boolean> {
+    await fs.mkdir(globalStoragePath, { recursive: true });
+    try {
+        const handle = await fs.open(path.join(globalStoragePath, PROMPT_CLAIM_FILE), 'wx');
+        await handle.close();
+        return true;
+    } catch (error) {
+        if ((error as NodeJS.ErrnoException).code === 'EEXIST') {
```
edvilme, line 246: Maybe add a comment explaining how the PROMPT_CLAIM_FILE logic works, I am guessing it checks if it exists, and if it doesn't, returns false; and when releasing, it removes it?

edvilme (dismissed):

src/features/feedback/feedbackPromptService.ts
StellaHuang95: Sure just added comments

eleanorjboyd:
AI agent review finding:

After `showPromptIfEligible()` acquires the persistent once-only claim, a rejection from `showInformationMessage()` leaves the claim marker in place. Future extension hosts then treat the prompt as already shown even though the user may never have seen it. Please release the claim when notification display fails, while retaining it after a notification was successfully presented, and add a rejection regression test.

I validated the PR with `npm run compile-tests`, `npm run compile`, and the focused `FeedbackPromptService` tests (7 passing).

_Reviewed and posted by Eleanor's AI agent._

PR #1708, opened 2026-08-12
Respect terminal cwd when resolving local venvs

### LINKED ISSUE #1631: Auto-activation forces a single environment across an entire multi-project workspace root, injecting activate commands into unrelated terminals
## Summary

`python-envs.terminal.autoActivationType` (default `"command"`) auto-injects a `source .../activate` command into **every new integrated terminal**, based on a single environment resolved for the entire workspace root — even when that root folder contains many independent subprojects, each with its own `.venv`, and the new terminal's cwd is a *different* subproject than the one whose environment got selected.

## Repro

1. Open a folder (e.g. `~/AILab`) that is not itself a Python project, but contains several subfolders that each are (e.g. `AILab/ProjectA/.venv`, `AILab/ProjectB/.venv`, `AILab/ProjectC/.venv`).
2. Work in `ProjectA` for a while (open files, run terminals) — the extension resolves/selects `ProjectA/.venv` as "the" environment for the workspace root `AILab` (log: `Changed environment from undefined to .venv (...) for: /Users/silent/AILab`).
3. Open a brand new integrated terminal cwd'd in `ProjectC` (a completely unrelated subproject with its own `.venv`).
4. Observed: the terminal automatically runs `source AILab/ProjectA/.venv/bin/activate` — activating the *wrong* project's environment — without the user typing anything. The extension log shows this is driven by `workspaceSearchPaths` (default `[".venv", "*/.venv"]`) resolving one environment for the whole root rather than scoping per-subfolder/per-terminal-cwd.

Example from my logs:
```
2026-07-11 19:29:06.742 [info] Python API: Changed environment from undefined to .venv (3.12.13.final.0) for: /Users/silent/AILab
2026-07-11 19:29:11.749 [error] Shell execution timed out:  source /Users/silent/AILab/Medical_Agent/.venv/bin/activate
2026-07-11 19:29:11.749 [info] Terminal is activated: /Users/silent/AILab/Medical_Agent/.venv/bin/python
```
This also repeatedly logs `Shell execution timed out`, suggesting the injection itself is unreliable/retried.

## Impact

- Wrong environment silently activated in terminals for unrelated subprojects — confusing at best (mismatched dependencies, wrong Python version) and could cause real harm if a script in one project runs against another project's environment/dependencies.
- Because this fires for *every* integrated terminal regardless of the extension/tool that opened it, it also leaks into third-party terminal-integration extensions (e.g. Claude Code's integrated terminal panels), where the injected activation line was mistaken for injected/leaked chat input since it appears in the terminal buffer unprompted.
- The setting is `scope: "machine"`, so it cannot be scoped per-workspace-folder to opt individual multi-project roots out — the only escape is fully disabling auto-activation (`"off"`) machine-wide, losing the feature everywhere, even in genuine single-project workspaces where it'd be wanted.

## Expected behavior

- Auto-activation should be scoped to the environment resolved for the **specific terminal's cwd / the specific subfolder project**, not a single environment resolved for the entire (possibly multi-project) workspace root.
- Alternatively, `workspaceSearchPaths`/environment resolution should not silently pick one `.venv` out of several sibling `*/.venv` matches to represent an entire root folder that isn't itself a Python project.
- `autoActivationType` should ideally support workspace-folder scope, not just `machine`, so it can be disabled for specific multi-project roots without losing it elsewhere.

## Environment

- `ms-python.vscode-python-envs` v1.36.0 (darwin-arm64)
- VS Code, macOS
- Workspace: single large folder containing multiple independent Python subprojects, each with their own `.venv`

StellaHuang95:
## Summary

- resolve terminal environments from the terminal cwd before using workspace-wide fallbacks
- reject a selected local venv when it belongs to a sibling directory, then use one nearest, unambiguous venv that contains the cwd
- fail closed instead of activating another project's environment when the cwd project has no safe match
- keep terminal activation menu context synchronized with the currently active terminal

Fixes #1631

## Context

The issue uses a single VS Code workspace folder as a container for several independent Python projects:

```text
AILab/
  ProjectA/.venv
  ProjectB/.venv
  ProjectC/.venv
```

With the default `workspaceSearchPaths` (`[".venv", "*/.venv"]`), all three environments are discovered. However, the workspace root is automatically registered as the only Python project unless the nested projects are explicitly added. A terminal opened in ProjectC could therefore receive ProjectA's activation command.

Besides activating the wrong dependencies and Python version, this affects every integrated terminal opened by other extensions because auto-activation runs on the terminal-open event.

## Root cause

The failure was caused by two scope-collapsing decisions:

1. Nested environment paths resolve to the automatically registered workspace-root project through ancestor project lookup. `VenvManager` can therefore associate one sorted or persisted sibling venv with that root.
2. `getEnvironmentForTerminal()` previously preferred project-wide environment consensus before checking terminal cwd. With exactly one registered project, it returned the root's selected environment immediately, so the cwd heuristic never ran.

Simply moving the existing cwd lookup earlier is insufficient: `api.getEnvironment(ProjectC)` still maps ProjectC back to the root project and returns the same sibling environment.

### Cwd-aware terminal resolution

The terminal resolver now:

1. reads shell-integration cwd first, then terminal creation cwd
2. finds the most-specific registered project containing that cwd
3. preserves the project's selected environment unless it is a built-in venv located on a divergent sibling branch
4. for a sibling venv only, searches already-known environments for candidates that:
   - belong to the built-in venv manager
   - are valid and have an absolute `sysPrefix`
   - live inside the containing project
   - have an owner directory that is an ancestor of the terminal cwd
5. selects exactly one nearest candidate

If no candidate exists, multiple candidates are equally near, lookup fails, or lookup exceeds one second, the resolver returns no environment rather than activating the wrong one. The losing timeout is cancelled after a successful lookup.

A known cwd project with no selected environment also stops resolution, preventing another project's sole environment from leaking through the previous multi-project fallback.

### Compatibility boundaries

This is intentionally terminal-only. It does not change central environment selection, persisted manager state, debugger/Pylance resolution, status-bar selection, task execution, or the public environment API.

| Scenario | Behavior |
| --- | --- |
| `workspace/.venv` with a terminal in `workspace/src` | Existing root venv remains selected |
| Registered nested project with an explicit environment | Existing project environment remains selected |
| Conda/Poetry/Pyenv or another external environment | Existing selected environment remains selected |
| Terminal cwd outside all registered projects | Existing fallback behavior remains |
| Terminal cwd unavailable | Existing fallback behavior remains |
| Selected venv is under ProjectA and cwd is under ProjectC | Use ProjectC's unique nearest venv, or activate nothing |
| Cwd project has no selected environment | Activate nothing; do not borrow another project's environment |

### Terminal activation UI

Safe resolution can intentionally return `undefined`. The activation menu contexts are now cleared in that case instead of retaining the previous terminal's Activate/Deactivate state. Async context updates also verify that the originating terminal is still active, preventing slower lookups or background activation events from overwriting the currently focused terminal's global menu state.

## Tests

Added direct coverage for:

- normal workspace-root venvs and nested cwd paths
- sibling ProjectA/ProjectB/ProjectC venv resolution
- shell-integration cwd precedence
- ambiguous equally near venvs
- missing local venvs
- global venv exclusion
- workspace-root terminals with a selected descendant venv
- projects with no selected environment
- preventing another project's environment from leaking into the cwd project
- environment-list rejection and timeout
- external environments, explicit nested projects, outside-workspace cwd, and unavailable cwd
- clearing activation menu contexts when no environment resolves

Validation performed:

- `npm run compile-tests`
- `npm run compile`
- ESLint on all changed TypeScript files
- all 97 tracked unit-test files: **1,525 passing, 5 expected pending**

## Out of scope

The shell-execution timeout also reported in #1631 is a separate activation-state issue. This PR addresses only selecting the correct environment for a terminal cwd.

### DIFF
--- a/src/features/terminal/utils.ts
+++ b/src/features/terminal/utils.ts
@@ -1,7 +1,14 @@
 import * as path from 'path';
 import { Disposable, env, ExtensionTerminalOptions, tasks, Terminal, TerminalOptions, Uri } from 'vscode';
-import { PythonEnvironment, PythonProject, PythonProjectEnvironmentApi, PythonProjectGetterApi } from '../../api';
-import { traceVerbose } from '../../common/logging';
+import {
+    PythonEnvironment,
+    PythonEnvironmentApi,
+    PythonProject,
+    PythonProjectEnvironmentApi,
+    PythonProjectGetterApi,
+} from '../../api';
+import { VENV_MANAGER_ID } from '../../common/constants';
+import { traceError, traceVerbose } from '../../common/logging';
 import { timeout } from '../../common/utils/asyncUtils';
 import { createSimpleDebounce } from '../../common/utils/debounce';
 import { onDidChangeTerminalShellIntegration, onDidWriteTerminalData } from '../../common/window.apis';
@@ -10,6 +17,7 @@ import { identifyTerminalShell } from '../common/shellDetector';
 import { shellIntegrationSupportedShells } from './shells/common/shellUtils';
 
 export const SHELL_INTEGRATION_TIMEOUT = 500; // 0.5 seconds
+const LOCAL_VENV_LOOKUP_TIMEOUT_MS = 1000;
 
 /**
  * Use `terminal.integrated.shellIntegration.timeout` setting if available.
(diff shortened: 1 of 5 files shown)


src/features/terminal/utils.ts
```diff
@@ -191,13 +199,152 @@ async function getDistinctProjectEnvs(
...
+
+    return unique[0];
+}
+
+interface CwdEnvironmentResolution {
+    environment?: PythonEnvironment;
+    stopFallback: boolean;
+}
+
+async function getEnvironmentForCwd(
```
StellaHuang95, line 301: Previously, a new terminal could activate a sibling project’s venv because workspace-level selection was checked before terminal cwd. The resolver now uses cwd, rejects sibling venvs, and selects the nearest unambiguous local venv, or activates nothing if unsafe.

eleanorjboyd (approved):

edvilme (approved):

Issue #1584, opened 2026-06-15
[Python Environments] Environments not detected


franciswairagu:
Type: <b>Bug</b>

## Description
Jupyter can't detect any conda env I have been using

## Steps to Reproduce
1. 
2. 
3. 

## Actual Behavior

<details>
<summary>Environment Information</summary>

```
Extension Version: 1.34.0
Python Extension Version: 2026.4.0

Registered Environment Managers (6):
  - ms-python.python:system (Global)
  - ms-python.python:venv (venv)
  - ms-python.python:pyenv (PyEnv)
  - ms-python.python:pipenv (Pipenv)
  - ms-python.python:poetry (Poetry)
  - ms-python.python:conda (Conda)

Total Available Environments: 6
Environment Details:
  1. Python 3.13.12.final.0 (3.13.12.final.0) - /usr/bin/python
  2. Python 2.7.18.final.0 (2.7.18.final.0) - /usr/bin/python2
  3. dsf (no-python) (no-python) - /home/wairagu/anaconda3/envs/dsf
  4. module4 (3.14.4) (3.14.4) - /home/wairagu/anaconda3/envs/module4
  5. base (3.13.9) (3.13.9) - /home/wairagu/anaconda3
  6. python311 (3.11.15) (3.11.15) - /home/wairagu/anaconda3/envs/python311

Python Projects (1):
  1. /home/wairagu/Desktop/Moringa/my labs
     Environment: base (3.13.9)

Extension Settings:
  Default Environment Manager: ms-python.python:conda
  Default Package Manager: ms-python.python:conda
Auto-activation is "command". Activation based on first 'py-env.terminal.autoActivationType' setting which is 'command' and 'python.terminal.activateEnvironment' if the first is undefined which is 'true'.

```

</details>

Extension version: 1.34.0
VS Code version: Code 1.124.2 (6928394f91b684055b873eecb8bc281365131f1c, 2026-06-12T00:12:36Z)
OS version: Linux x64 6.19.11+kali-amd64
Modes:

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|AMD Ryzen 5 5500U with Radeon Graphics (12 x 4042)|
[...]

PR #1716, opened 2026-08-17
Report an empty version for conda environments without Python

LH-and-FPGA:
## Summary

A conda environment that has no Python interpreter makes **all** conda environments
disappear from the interpreter picker and the Jupyter kernel picker.

This happens with a prefix created as a toolchain rather than a Python environment,
e.g. `conda create -n cuda cuda-toolkit`. Reproduced on Linux with 5 conda
environments, one of them interpreter-less; only `base` survived.

Likely the same root cause as #1584 — that report's environment list shows a third
entry rendered as `(no-python)`, and it was closed as `info-needed`.

## Root cause

`getCondaWithoutPython` describes an interpreter-less environment with a placeholder
in a field that consumers parse:

```ts
// src/managers/conda/condaUtils.ts
version: 'no-python',
```

`version` is declared `readonly version: string` in `src/api.ts`, so a sentinel
compiles fine — but `ms-python.python` parses it as a PEP 440 version:

```js
// ms-python.python 2026.7.2026080801, out/client/extension.js
function f(e){ const t = l.parseBasicVersionInfo(`ignored-${e}`);
  if(!t){ if("" === e) return [v(), ""];        // "unknown" — degrades gracefully
          throw Error(`invalid version ${e}`) } }  // 'no-python' lands here
```

The empty string is the one unparseable value it accepts. Any other placeholder
throws.

**Why one bad environment takes down the rest.** This repository is careful —
`condaUtils.ts` wraps each environment in its own try/catch during conversion. But
the sentinel is *valid data*, so it passes that guard and is published in the batch
fired from `condaEnvManager.ts`. The consumer has no per-item guard:

```js
onDidChangeEnvironments(e){ e.forEach(e => { … this.addEnv(e.environment) }) }
```

`Array.forEach` cannot resume after a throw, so every environment ordered *after* the
offending one is silently dropped. Observed stack:

```
at f                              ← throw Error(`invalid version no-python`)
at _                              ← convert PythonEnvironmentInfo
at S.addEnv
at Array.forEach (<anonymous>)    ← no try/catch
at S.onDidChangeEnvironments
at T.fire
```

## The fix

**1. Don't publish a sentinel as a version** (`condaUtils.ts`, `condaEnvManager.ts`)

`version: 'no-python'` → `version: ''`. This is already how the rest of the repository
represents an unknown version:

- `src/features/interpreterSelection.ts` — `version: resolved.version ?? ''`
- `src/common/inlineScript/interpreter.ts` — `env.version.length === 0` means not usable

Conda was the only one of the eight managers surfacing a sentinel. The `(no-python)`
marker stays in `displayName` / `shortDisplayName`, which are displayed but never
parsed, so the UI is unchanged. The two internal `version === 'no-python'` checks now
go through an exported `isCondaEnvWithoutPython` predicate.

This also fixes a smaller bug: `pickPythonVersion` builds its list with
`.map(e => e.version).filter(Boolean)`, and `'no-python'` is truthy, so "Select the
version of Python to install" offered `no-python` as a choice.

**2. Make `sortEnvironments` a total order** (`managers/common/utils.ts`)

```ts
return a.version ? 1 : -1;
```

Comparing a real version against an unparseable non-empty one returns `1` in both
directions, which breaks the antisymmetry `Array.prototype.sort` requires, so the
result is an implementation-defined permutation. Now: valid versions compare
descending, a known version sorts before an unknown one, and two unknowns fall back to
a string comparison.

`getLatest` had a related gap — seeded with `candidates[0]`, both operands had to parse
for the seed to ever be replaced, so an unparseable seed always won.

## Verification

Measured against the real `@renovatebot/pep440`, on the reported environment set
(`base 3.13.13`, an interpreter-less prefix, `git 3.14.6`, `lh 3.14.7`, `vllm 3.14.6`):

| | before | after |
|---|---|---|
| antisymmetry violations over a 9-version vocabulary | 36 | 0 |
| distinct sort results over 120 input permutations | 16 | 1 |
| `getLatest` result | the interpreter-less env | `lh` |
| environments published before the consumer throws | `base` only | all 5 |

## Tests

- `npm run unittest`: 1582 passing, 0 failing (1577 before this change).
- New `utils.sortEnvironments.unit.test.ts` — descending order, unknown versions last,
  and stability across all input permutations.
- New `condaUtils.noPythonEnv.unit.test.ts` — an interpreter-less prefix is still
  discovered, reports `version: ''`, and keeps its `(no-python)` display name.
- Three of the four new tests fail on `main` and pass with this change; the fourth is a
  baseline that passes on both.
- Updated the three existing fixtures that constructed environments with `'no-python'`.

`src/api.ts` is unchanged, so no API version bump or changelog entry is needed.


LH-and-FPGA:

heejaechang:

src/managers/conda/condaUtils.ts
```diff
@@ -765,6 +765,10 @@ function getCondaWithoutPython(name: string, prefix: string, conda: string): Pyt
     };
 }
 
+export function isCondaEnvWithoutPython(environment: PythonEnvironment): boolean {
+    return environment.version === '';
+}
```
heejaechang, line 785: > **Warning** · Non-blocking recommendation

`version === ''` also represents an unknown version generally, so it cannot reliably prove that Python is absent. Preserve a Conda-specific discriminator (or check executable availability), and cover an environment with unavailable version metadata that still has a runnable interpreter.

heejaechang (dismissed):
Approved via Review Center.

heejaechang:

src/managers/conda/condaUtils.ts
LH-and-FPGA: Do I need to separate "no version" and "no interpreter" into two different flag?

eleanorjboyd: I think these need to remain two distinct states, but not necessarily as two new public flags. `""` is already the generic "version metadata unavailable" value: `tryResolveInterpreterPath` creates a runnable environment with `version: resolved.version ?? ""` and a real executable. If that resolves to the Conda manager, this predicate sends an already-runnable interpreter through the install-Python flow and can reject it as the base fallback. Please classify no-interpreter independently—e.g. through the existing `error`/capability representation—and add a regression test with a valid executable plus an empty version.

edvilme:
Hello @LH-and-FPGA 
Thanks for your contribution. Since your last commit, we have introduced `PythonVersion` and `PythonVersionSpecifier` for parsing and comparing Python versions, which relates to your changes. Please solve the merge conflicts and we can make another review. Thx!

LH-and-FPGA:
Hello @edvilme. Thank you for your message, I've updated my branch and resolved the conflicts.

eleanorjboyd:
@StellaHuang95 can you manage and help shepherd this. Thanks!

PR #1704, opened 2026-08-07
test: cover registered package manager lifecycles

edvilme:
## Summary

Adds a package-manager-centric integration baseline that intentionally precedes and de-risks #1686, so the package-manager command refactor is exercised against behavior established on `main`.

- drives one stateful install/list/direct-package/uninstall lifecycle per active profile
- uses unique disposable projects and manager-owned disposable environments
- exercises the live registered manager instances through a runtime-gated integration-test bridge
- guards registry completeness so every registered package-manager ID has an active fixture or explicit deferral
- covers normal Pip execution and Conda when their runtime prerequisites are available
- records an uncached baseline instead of assuming a newly created environment is empty
- restores workspace-scoped configuration from `inspect()` snapshots and performs guarded failure-safe cleanup
- defers Poetry pending a Poetry-owned project/lockfile lifecycle
- defers uv-backed Pip because changing the machine-scoped selection reliably within one extension host was not stable on `main`, while available-version lookup would also introduce `uv tool run pip` network seeding
- pins the disposable integration-test user profile to normal Pip execution

## Validation

- `npm run compile`
- `npm run compile-tests`
- `npm run lint`
- `npm run unittest`
- targeted `packageManagement.integration.test.js`: 3 passing, 2 prerequisite skips locally
  - Pip skipped because quick create selected Python 3.15.0 alpha, whose bundled Pip metadata is incomplete
  - Conda skipped because Conda is not installed
- reviewer specialist: clean, no Critical or Important findings

The active Pip and Conda fixtures require package-index/network access when their runtime prerequisites are present.

Fixes #1701

PR #1717, opened 2026-08-17
test: add cross-platform package manager lifecycle coverage

### LINKED ISSUE #1650: Package list shows "No packages found" when pip emits an upgrade notice (JSON parse failure)

edvilme:
Follow-up to #1704.

## Summary

- Adds shared conformance coverage for Pip, Conda, and Poetry headless package operations, ensuring failures reject without prompting or showing error UI.
- Runs real Pip and Conda install, list, uninstall, and version-lookup coverage across the full OS/Python CI matrix.
- Creates test-owned ephemeral venv and Conda environments in the integration suite, with ownership checks, bounded processes, and guaranteed cleanup; CI only provisions the required tools.
- Distinguishes unsupported package version lookup with an internal `PackageVersionLookupNotSupportedError`. Supported profiles require a defined, non-empty result, while command, network, and parsing failures propagate.
- Keeps Poetry lifecycle coverage explicitly deferred until CI has a controlled Poetry installation.

## Validation

- Lint, TypeScript compilation, and unit tests
- Pip lifecycle and version lookup in a real VS Code integration host
- Full GitHub Actions OS/Python matrix


edvilme:
CI diagnosis and fix pushed in df6d1a2. The opt-in live-network suite introduced by #1704 fails identically on #1704, current main, and this PR: Pip version lookup returns undefined, then Conda quick-create hangs until the shared 300s Mocha timeout. Because PR #1717's intended coverage is deterministic headless conformance, the fix removes only the automatic VSC_PYTHON_PACKAGE_NETWORK_TEST invocation from PR/push workflows; it retains the tests behind their explicit opt-in gate, regular integration profile coverage, and the new deterministic unit conformance. Local validation: lint, compile-tests, 1628 unit tests, workflow YAML/Prettier, and targeted Package Manager profile coverage integration test all pass.

StellaHuang95:

src/test/managers/common/packageManagerHeadlessConformance.unit.test.ts
```diff
@@ -0,0 +1,76 @@
...
+        }
+
+        assert.ok(pipPicker.notCalled);
+        assert.ok(condaPicker.notCalled);
+        assert.ok(poetryInput.notCalled);
+    });
+
+    test('rejects failures without showing error notifications', async () => {
+        const operationError = new Error('package operation failed');
+        sinon.stub(windowApis, 'withProgress').callsFake(async (_options, task) => task({} as never, {} as never));
```
StellaHuang95: > **Issue** · Please address or respond

📍 src/test/managers/common/packageManagerHeadlessConformance.unit.test.ts:60

`runHeadless` promises no user interaction, but this stub allows every manager to invoke progress UI. Assert that `withProgress` is not called and update Pip, Conda, and Poetry to execute the operation directly when headless.

[verified]

.github/workflows/pr-check.yml
```diff
@@ -335,14 +335,6 @@ jobs:
         if: runner.os != 'Linux'
```
StellaHuang95, line 344: > **Warning** · Non-blocking recommendation

📍 .github/workflows/pr-check.yml:335

Removing this job and its `push-check.yml` counterpart leaves the guarded network lifecycle tests unreachable in CI; the new mocked suite cannot detect regressions in real install, list, uninstall, or version lookup commands. Retain required coverage or move it to an explicit scheduled/manual workflow before deleting both jobs.

[verified]

src/test/managers/common/packageManagerHeadlessConformance.unit.test.ts
```diff
@@ -0,0 +1,76 @@
...
+        sinon.stub(builtinUtils, 'managePackages').rejects(operationError);
+        sinon.stub(condaUtils, 'managePackages').rejects(operationError);
+        sinon.stub(poetryUtils, 'getPoetry').resolves(undefined);
+        const showErrorMessage = sinon.stub(windowApis, 'showErrorMessage').resolves(undefined);
+        const showErrorMessageWithLogs = sinon.stub(errorUtils, 'showErrorMessageWithLogs').resolves();
+
+        for (const manager of createManagers()) {
+            await assert.rejects(
```
StellaHuang95, line 71: > **Warning** · Non-blocking recommendation

📍 src/test/managers/common/packageManagerHeadlessConformance.unit.test.ts:64

[unverified] The test fails before post-operation package refresh runs, leaving refresh failures uncovered. Add a case where package management succeeds but refresh rejects, then verify rejection occurs without error UI.

[verified]

src/test/managers/common/packageManagerHeadlessConformance.unit.test.ts
edvilme: Fixed in efe2f9c. Headless Pip, Conda, and Poetry operations now execute directly without `withProgress`, while preserving the same package operation, refresh, logging, and rejection behavior. The conformance test now asserts that progress UI is never invoked.

.github/workflows/pr-check.yml
edvilme: Addressed in efe2f9c without restoring the unstable suite to required PR/push checks. Added a dedicated `Package Manager Network Check` workflow that can be explicitly dispatched and runs the guarded lifecycle suite with `VSC_PYTHON_PACKAGE_NETWORK_TEST=1` on Ubuntu/Python 3.12.

src/test/managers/common/packageManagerHeadlessConformance.unit.test.ts
edvilme: Added in efe2f9c. The conformance suite now covers successful package management followed by a rejected package refresh for Pip, Conda, and Poetry, asserting that the refresh error propagates without progress or error UI.

src/test/managers/common/packageManagerHeadlessConformance.unit.test.ts
```diff
@@ -0,0 +1,110 @@
...
+            .stub(
+                PoetryPackageManager.prototype as unknown as {
+                    runPoetryManage: () => Promise<void>;
+                },
+                'runPoetryManage',
+            )
+            .resolves();
+        sinon.stub(builtinUtils, 'refreshPipPackages').rejects(refreshError);
+        sinon.stub(CondaPackageManager.prototype, 'getPackages').rejects(refreshError);
+        sinon.stub(PoetryPackageManager.prototype, 'getPackages').rejects(refreshError);
```
StellaHuang95: > **Warning** · Non-blocking recommendation

This stubs Pip's refresh helper itself to reject, bypassing its production failure boundary. Exercise the real refresh path—or make headless refresh failures propagate at that boundary—so the test verifies the intended contract rather than a manufactured rejection.

src/test/managers/common/packageManagerHeadlessConformance.unit.test.ts
```diff
@@ -0,0 +1,110 @@
...
+        const operationError = new Error('package operation failed');
+        const withProgress = sinon.stub(windowApis, 'withProgress');
+        sinon.stub(builtinUtils, 'managePackages').rejects(operationError);
+        sinon.stub(condaUtils, 'managePackages').rejects(operationError);
+        sinon.stub(poetryUtils, 'getPoetry').resolves(undefined);
+        const showErrorMessage = sinon.stub(windowApis, 'showErrorMessage').resolves(undefined);
+        const showErrorMessageWithLogs = sinon.stub(errorUtils, 'showErrorMessageWithLogs').resolves();
+
+        for (const manager of createManagers()) {
+            await assert.rejects(
```
StellaHuang95, line 71: > **Warning** · Non-blocking recommendation

The Poetry case rejects because `getPoetry()` returns `undefined`, while Pip and Conda reject from the package-operation stub. Stub Poetry's actual management operation to reject the same expected error and assert its identity so this covers consistent failure propagation.

StellaHuang95 (dismissed):
Approved via Review Center.

.github/workflows/pr-check.yml
edvilme: Updated in cd739a5 based on follow-up discussion: the separate manual workflow has been removed. `VSC_PYTHON_PACKAGE_NETWORK_TEST=1` is now set on the existing integration job in both PR and push workflows, so the normal and network package-manager suites run together for every OS/Python matrix entry.

src/test/managers/common/packageManagerHeadlessConformance.unit.test.ts
```diff
@@ -0,0 +1,110 @@
...
+    const environment = {
+        envId: { id: 'test-environment', managerId: 'test-manager' },
+        environmentPath: Uri.joinPath(Uri.file(__dirname), 'path', 'to', 'environment'),
+    } as PythonEnvironment;
+
+    teardown(() => {
+        sinon.restore();
+    });
+
+    function createManagers(): PackageManager[] {
```
StellaHuang95: > **Issue** · Please address or respond

`createManagers` is declared before its callers, contrary to the documented local-helper placement rule. Move it below the final `test(...)` block in this suite.

[verified]

src/test/managers/common/packageManagerHeadlessConformance.unit.test.ts
```diff
@@ -0,0 +1,110 @@
...
+        sinon
+            .stub(
+                PoetryPackageManager.prototype as unknown as {
+                    runPoetryManage: () => Promise<void>;
+                },
+                'runPoetryManage',
+            )
+            .resolves();
+        sinon.stub(builtinUtils, 'refreshPipPackages').rejects(refreshError);
+        sinon.stub(CondaPackageManager.prototype, 'getPackages').rejects(refreshError);
```
StellaHuang95: > **Issue** · Please address or respond

This forces `refreshPipPackages` to reject even though production catches refresh failures and returns cached packages, so the shared test asserts a contract Pip does not implement. Exercise Pip's real failure boundary and test suppression, or intentionally change production so headless refresh failures propagate.

[verified]

src/test/managers/common/packageManagerHeadlessConformance.unit.test.ts
```diff
@@ -0,0 +1,110 @@
...
+        sinon.stub(condaUtils, 'managePackages').rejects(operationError);
+        sinon.stub(poetryUtils, 'getPoetry').resolves(undefined);
+        const showErrorMessage = sinon.stub(windowApis, 'showErrorMessage').resolves(undefined);
+        const showErrorMessageWithLogs = sinon.stub(errorUtils, 'showErrorMessageWithLogs').resolves();
+
+        for (const manager of createManagers()) {
+            await assert.rejects(
+                manager.manage(environment, { install: ['requests'], runHeadless: true }),
+            );
+        }
```
StellaHuang95, line 75: > **Issue** · Please address or respond

Poetry rejects because its executable lookup returns `undefined`, while Pip and Conda reject `operationError`; accepting any rejection lets unrelated failures satisfy the test. Make Poetry's package operation reject `operationError` and assert `error === operationError` for every manager.

[verified]

.github/workflows/pr-check.yml
```diff
@@ -252,6 +252,8 @@ jobs:
     name: Integration Tests
     runs-on: ${{ matrix.os }}
     needs: [smoke-tests]
+    env:
+      VSC_PYTHON_PACKAGE_NETWORK_TEST: '1'
```
StellaHuang95, line 257: > **Warning** · Non-blocking recommendation

Moving the network-test flag to job scope expands the previously targeted Linux/Python 3.12 run to every OS/Python matrix entry, multiplying CI cost and external-network flake exposure. Restrict it to an explicit matrix subset or a single required job, and mirror the correction in `push-check.yml`.

[verified]

.github/workflows/pr-check.yml
edvilme: Follow-up fix in 13cd44d after enabling the network suites across the full matrix: CI now sets `CONDA_PLUGINS_AUTO_ACCEPT_TOS=yes` so non-interactive Conda environment creation does not hang, and the Pip lifecycle profile temporarily disables `alwaysUseUv` so it exercises Pip rather than failing during UV tool bootstrap. The original scoped setting is restored in teardown.

.github/workflows/pr-check.yml
edvilme: This is intended to make sure the commands work on every matrix entry (they might be different in different versions of python/pip or on different OSes)

src/test/managers/common/packageManagerHeadlessConformance.unit.test.ts
```diff
@@ -0,0 +1,110 @@
...
+
+        assert.ok(withProgress.notCalled);
+        assert.ok(showErrorMessage.notCalled);
```
StellaHuang95, line 79: > **Issue** · Please address or respond

📍 src/test/managers/common/packageManagerHeadlessConformance.unit.test.ts:63

Poetry still fails during executable lookup, while the bare `assert.rejects` accepts any error. Stub Poetry's management operation to reject `operationError` and assert exact error identity for every manager; the documented no-partial-assertions rule makes this must-fix.

[verified]

src/test/managers/common/packageManagerHeadlessConformance.unit.test.ts
```diff
@@ -0,0 +1,110 @@
...
+                    runPoetryManage: () => Promise<void>;
+                },
+                'runPoetryManage',
+            )
+            .resolves();
+        sinon.stub(builtinUtils, 'refreshPipPackages').rejects(refreshError);
+        sinon.stub(CondaPackageManager.prototype, 'getPackages').rejects(refreshError);
+        sinon.stub(PoetryPackageManager.prototype, 'getPackages').rejects(refreshError);
+        sinon.stub(PipPackageManager.prototype, 'getDirectPackageNames').resolves(undefined);
+        sinon.stub(PoetryPackageManager.prototype, 'getDirectPackageNames').resolves(undefined);
```
StellaHuang95, line 136: > **Warning** · Non-blocking recommendation

📍 src/test/managers/common/packageManagerHeadlessConformance.unit.test.ts:91

The Pip and Poetry stubs replace the refresh methods whose production failure handling this test needs to exercise, manufacturing rejection paths rather than proving the real contract. Split into manager-specific cases that fail the underlying command boundary and assert actual suppression or deliberately change production to propagate refresh failures.

[verified]

.github/workflows/pr-check.yml
```diff
@@ -252,6 +252,9 @@ jobs:
     name: Integration Tests
     runs-on: ${{ matrix.os }}
     needs: [smoke-tests]
+    env:
+      CONDA_PLUGINS_AUTO_ACCEPT_TOS: 'yes'
+      VSC_PYTHON_PACKAGE_NETWORK_TEST: '1'
```
StellaHuang95, line 257: > **Warning** · Non-blocking recommendation

📍 .github/workflows/pr-check.yml:255

The network flag remains job-scoped, expanding the former Linux/Python 3.12 run across the complete matrix; the added TOS and Pip configuration mitigate hangs but not cost or external-service flakiness. Restrict network coverage to an explicit matrix subset and mirror the change in `push-check.yml`.

[verified]

.github/workflows/pr-check.yml
```diff
@@ -252,6 +252,9 @@ jobs:
     name: Integration Tests
     runs-on: ${{ matrix.os }}
     needs: [smoke-tests]
+    env:
+      CONDA_PLUGINS_AUTO_ACCEPT_TOS: 'true'
+      VSC_PYTHON_PACKAGE_NETWORK_TEST: '1'
```
StellaHuang95, line 257: > **Warning** · Non-blocking recommendation

The network flag remains job-scoped, running external package lifecycles across every OS/Python matrix entry. Restrict it to an explicit subset or focused job and mirror that scope in `push-check.yml`.

[verified]

src/test/integration/packageManager.integration.test.ts
```diff
@@ -124,11 +132,25 @@ for (const profile of profiles) {
...
+            if (profile.reuseExistingEnvironment) {
+                environment = (await api.getEnvironments('global')).find(
+                    (candidate) => candidate.envId.managerId === profile.environmentManagerId,
+                );
+                assert.ok(environment, `No existing ${profile.name} environment is available`);
+                return;
+            }
+
             const projectUri = vscode.Uri.joinPath(
                 workspaceUri,
```
StellaHuang95: > **Warning** · Non-blocking recommendation

Reusing the first global Conda environment can leave newly installed dependencies after teardown removes only the target package. Compare the complete package set before and after the suite, or restore isolation by creating and deleting a dedicated environment.

[verified]

src/test/integration/packageManager.integration.test.ts
```diff
@@ -124,11 +132,25 @@ for (const profile of profiles) {
...
                 this.skip();
                 return;
             }
 
+            if (profile.reuseExistingEnvironment) {
+                environment = (await api.getEnvironments('global')).find(
+                    (candidate) => candidate.envId.managerId === profile.environmentManagerId,
+                );
+                assert.ok(environment, `No existing ${profile.name} environment is available`);
+                return;
```
StellaHuang95: > **Issue** · Please address or respond

Reusing the first global Conda environment can leave installed packages, transitive dependency changes, or partial solver changes behind if the test fails before its in-test uninstall. Please use an isolated environment or snapshot and restore the full package set in teardown.

.github/workflows/pr-check.yml
StellaHuang95, line 257: > **Warning** · Non-blocking recommendation

This makes package-network tests run across the entire integration OS/Python matrix. That substantially increases external-network cost and flake exposure; please restrict them to an explicit matrix subset or focused job, and mirror the change in `push-check.yml`.

src/test/managers/common/packageManagerHeadlessConformance.unit.test.ts
```diff
@@ -0,0 +1,110 @@
...
+        sinon.stub(builtinUtils, 'managePackages').rejects(operationError);
+        sinon.stub(condaUtils, 'managePackages').rejects(operationError);
+        sinon.stub(poetryUtils, 'getPoetry').resolves(undefined);
+        const showErrorMessage = sinon.stub(windowApis, 'showErrorMessage').resolves(undefined);
+        const showErrorMessageWithLogs = sinon.stub(errorUtils, 'showErrorMessageWithLogs').resolves();
+
+        for (const manager of createManagers()) {
+            await assert.rejects(
+                manager.manage(environment, { install: ['requests'], runHeadless: true }),
+            );
```
StellaHuang95, line 74: > **Issue** · Please address or respond

Poetry rejects during executable lookup here, while the bare `assert.rejects` accepts any error. This does not prove that `operationError` propagates from Poetry's package-management operation. Stub that operation to reject `operationError` and assert the exact error identity for every manager.

[verified]

src/test/managers/common/packageManagerHeadlessConformance.unit.test.ts
```diff
@@ -0,0 +1,110 @@
...
+        sinon.stub(PipPackageManager.prototype, 'getDirectPackageNames').resolves(undefined);
+        sinon.stub(PoetryPackageManager.prototype, 'getDirectPackageNames').resolves(undefined);
+        const showErrorMessage = sinon.stub(windowApis, 'showErrorMessage').resolves(undefined);
+        const showErrorMessageWithLogs = sinon.stub(errorUtils, 'showErrorMessageWithLogs').resolves();
+
+        for (const manager of createManagers()) {
+            await assert.rejects(
+                manager.manage(environment, { install: ['requests'], runHeadless: true }),
+                (error: unknown) => error === refreshError,
+            );
```
StellaHuang95: > **Issue** · Please address or respond

This shared refresh test injects rejection above manager-specific failure boundaries; Pip catches `refreshPipPackages` failures in production. Split this into manager-specific cases that fail the underlying command boundary and assert each manager's actual suppression or propagation contract.

[verified]

src/test/integration/packageManager.integration.test.ts
StellaHuang95: > **Issue** · Please address or respond

The suite mutates an arbitrary global Conda environment, but teardown neither owns that environment nor restores its complete package state. Create and delete a dedicated environment, or snapshot and restore the full package set even when the lifecycle test fails midway.

[verified]

src/test/integration/packageManager.integration.test.ts
```diff
@@ -215,14 +245,19 @@ for (const profile of profiles) {
...
-            const versions = await api.getPackageAvailableVersions(environment!, 'requests');
-            assert.ok(versions, `${profile.name} unexpectedly failed to retrieve package versions`);
+            const versions = await api.getPackageAvailableVersions(environment!, profile.packageName);
+            // The API currently returns undefined for both unsupported lookups and command/network failures.
+            // Skip until those outcomes can be distinguished by the API contract.
+            if (versions === undefined) {
+                this.skip();
+                return;
+            }
             assert.ok(versions.length > 0, 'No package versions available');
```
StellaHuang95, line 226: > **Issue** · Please address or respond

Returning `undefined` from a supported version lookup now skips the test, so command or network regressions can leave the network job green. Skip only profiles that explicitly lack lookup support; for supported profiles, require a defined non-empty result.

[verified]

src/test/integration/packageManager.integration.test.ts
```diff
@@ -124,11 +123,28 @@ for (const profile of profiles) {
...
+                );
+                if (!environment) {
+                    this.skip();
+                    return;
+                }
+                return;
+            }
+
             const projectUri = vscode.Uri.joinPath(
                 workspaceUri,
```
StellaHuang95: > **Issue** · Please address or respond

📍 src/test/integration/packageManager.integration.test.ts:141
The Conda suite still mutates the first matching global environment without owning or fully restoring it. Use a dedicated environment or snapshot and restore the complete package state in teardown, including after partial failures.

[verified]

src/test/integration/packageManager.integration.test.ts
```diff
@@ -215,14 +239,19 @@ for (const profile of profiles) {
...
+            // The API currently returns undefined for both unsupported lookups and command/network failures.
+            // Skip until those outcomes can be distinguished by the API contract.
+            if (versions === undefined) {
+                this.skip();
+                return;
+            }
             assert.ok(versions.length > 0, 'No package versions available');
         });
 
         suiteTeardown(async () => {
```
StellaHuang95, line 229: > **Issue** · Please address or respond

📍 src/test/integration/packageManager.integration.test.ts:248
Profiles that declare version-lookup support still convert `undefined` into a skip, allowing command or network regressions to leave CI green. Skip only unsupported profiles; supported profiles must return a defined, non-empty result.

[verified]

src/test/managers/common/packageManagerHeadlessConformance.unit.test.ts
edvilme: Addressed in 0462436. The Pip case now drives the real refresh path through the child-process boundary and verifies the production suppression contract.

src/test/managers/common/packageManagerHeadlessConformance.unit.test.ts
edvilme: Addressed in 0462436. Poetry now fails from its spawned package-management operation with the shared operationError, and the test asserts exact error identity.

src/test/managers/common/packageManagerHeadlessConformance.unit.test.ts
edvilme: Addressed in 0462436. createManagers is now below the test blocks with the other local helper.

src/test/managers/common/packageManagerHeadlessConformance.unit.test.ts
edvilme: Addressed in 0462436. Refresh coverage is manager-specific: Pip exercises its real command failure boundary and verifies that the failure is suppressed.

src/test/managers/common/packageManagerHeadlessConformance.unit.test.ts
edvilme: Addressed in 0462436. Poetry now emits operationError from the spawned management command, and every manager must reject with that exact object.

src/test/managers/common/packageManagerHeadlessConformance.unit.test.ts
edvilme: Addressed in 0462436. The Poetry operation uses the real process boundary and assert.rejects validates error === operationError for every manager.

src/test/managers/common/packageManagerHeadlessConformance.unit.test.ts
edvilme: Addressed in 0462436. The shared artificial refresh test was replaced with manager-specific tests against the underlying Pip, Conda, and Poetry command boundaries and their actual suppression/propagation contracts.

.github/workflows/pr-check.yml
edvilme: Keeping the full OS/Python matrix intentionally. Package-manager discovery and command behavior vary across both dimensions, so this PR is intended to exercise every integration entry; push-check.yml mirrors the same scope.

.github/workflows/pr-check.yml
edvilme: Keeping the full OS/Python matrix intentionally. The matrix is the requested coverage boundary for OS-, Python-, Pip-, and Conda-specific behavior, and both workflows remain aligned.

src/test/integration/packageManager.integration.test.ts
edvilme: Addressed in 0462436. CI now creates a dedicated Conda clone under RUNNER_TEMP, the test selects that exact normalized path, and an always-running workflow step removes it.

src/test/integration/packageManager.integration.test.ts
edvilme: Addressed in 0462436. The suite no longer mutates an arbitrary global environment; it uses a CI-owned Conda clone that is unconditionally cleaned up by the workflow.

.github/workflows/pr-check.yml
edvilme: Keeping the full OS/Python matrix intentionally. This is the requested coverage scope because package-manager behavior differs by OS and Python version; push-check.yml mirrors it.

src/test/managers/common/packageManagerHeadlessConformance.unit.test.ts
edvilme: Addressed in 0462436. Poetry now rejects from the spawned package operation with operationError, and exact identity is asserted for all managers.

src/test/managers/common/packageManagerHeadlessConformance.unit.test.ts
edvilme: Addressed in 0462436. Refresh tests are split by manager and fail each underlying command boundary, asserting Pip/Poetry suppression and Conda propagation.

src/test/integration/packageManager.integration.test.ts
edvilme: Addressed in 0462436. The Conda lifecycle now targets an isolated workflow-owned clone and cleanup runs even when the integration test fails.

src/test/integration/packageManager.integration.test.ts
edvilme: Addressed in 0462436. Only profiles that explicitly lack lookup support skip; supported profiles now require a defined, non-empty version result, with the API ambiguity documented for follow-up.

src/test/integration/packageManager.integration.test.ts
edvilme: Addressed in 0462436. The Conda profile discovers the exact CI-owned environment path and the workflow removes that isolated environment in an always step.

src/test/integration/packageManager.integration.test.ts
edvilme: Addressed in 0462436. Supported version lookups now fail on undefined and require at least one returned version; unsupported profiles remain the only skip path.

.github/workflows/pr-check.yml
StellaHuang95, line 257: > **Warning** · Non-blocking recommendation

This job-scoped flag enables external package operations for every OS/Python matrix entry, materially increasing CI cost and exposure to network/service flakiness. Please restrict this coverage to an intentional matrix subset or focused job, with the same scope in `push-check.yml`.

src/test/integration/packageManager.integration.test.ts
```diff
@@ -198,11 +243,14 @@ for (const profile of profiles) {
...
-                assert.ok(packages, 'Unable to list packages after uninstallation');
-                assert.ok(
-                    !packages.some((pkg) => pkg.name.toLowerCase() === packageName),
+                await waitForCondition(
+                    async () => {
+                        packages = await api.getPackages(environment!, { skipCache: true });
+                        return packages !== undefined && !packages.some((pkg) => pkg.name.toLowerCase() === packageName);
+                    },
+                    30_000,
                     'Package not uninstalled',
```
StellaHuang95, line 184: > **Warning** · Non-blocking recommendation

`waitForCondition` catches predicate exceptions, so persistent `getPackages` failures can become a generic 30-second “Package not installed/uninstalled” timeout. Preserve and report the last command error rather than masking it as a polling timeout.

StellaHuang95 (dismissed):

src/test/integration/packageManager.integration.test.ts
```diff
@@ -121,71 +113,49 @@ for (const profile of profiles) {
...
+            let packages: Package[] | undefined;
+            await waitForCondition(
+                async () => {
+                    packages = await api.getPackages(environment!, { skipCache: true });
+                    return packages?.some((pkg) => pkg.name.toLowerCase() === packageName) ?? false;
+                },
+                30_000,
                 'Package not installed',
+                1_000,
             );
```
StellaHuang95, line 166: > **Warning** · Non-blocking recommendation

If `waitForCondition` retries rejected `getPackages` calls, a persistent command failure is replaced by a generic 30-second timeout. Preserve the last refresh error or fail immediately on rejection; add a focused rejecting-`getPackages` test to lock this down.

src/test/managers/common/packageManagerHeadlessConformance.unit.test.ts
```diff
@@ -0,0 +1,176 @@
...
+            setImmediate(() => {
+                process.emit('exit', 1, null);
+                process.emit('close', 1, null);
+            });
+            return process as unknown as ReturnType<typeof childProcessApis.spawnProcess>;
+        });
+        sinon.stub(PipPackageManager.prototype, 'getDirectPackageNames').resolves(undefined);
+        const showErrorMessage = sinon.stub(windowApis, 'showErrorMessage').resolves(undefined);
+        const showErrorMessageWithLogs = sinon.stub(errorUtils, 'showErrorMessageWithLogs').resolves();
+        const manager = createManagers()[0];
```
StellaHuang95: > **Warning** · Non-blocking recommendation

Selecting managers with `[0]`, `[1]`, and `[2]` can silently redirect manager-specific tests when the factory order changes. Return named manager fields or select by manager identity.

src/test/integration/environmentFixture.ts
```diff
@@ -0,0 +1,506 @@
...
+                } catch (error) {
+                    cleanupErrors.push(toError(error));
+                }
+            }
+            if (ownershipVerified && (await pathExists(prefix))) {
+                cleanupErrors.push(
+                    new Error(`${request.name} API removal left the environment on disk: ${prefix.fsPath}`),
+                );
+                try {
+                    await request.provider.remove(prefix);
```
StellaHuang95: > **Warning** · Non-blocking recommendation

`withTimeout` rejects without cancelling `api.removeEnvironment`, after which fallback removal can start against the same prefix. Ensure the first removal has terminated before fallback deletion; verify with a deferred removal promise.

StellaHuang95 (dismissed):

src/test/integration/packageManager.integration.test.ts
```diff
@@ -189,9 +189,11 @@ for (const profile of profiles) {
             }
 
             const versions = await api.getPackageAvailableVersions(environment!, profile.packageName);
-            // The API does not yet distinguish an unsupported lookup from a command or network failure.
-            // Supported profiles must remain strict until that result contract can be made explicit.
-            assert.ok(versions, `${profile.name} unexpectedly failed to retrieve package versions`);
+            // Accept undefined until the API can distinguish unsupported lookups from command or network failures.
+            // TODO: Add that result distinction and make supported lookups strict in a follow-up PR.
+            if (versions === undefined) {
```
edvilme: Note to reviewers: This will be changed in #1686

src/test/integration/packageManager.integration.test.ts
```diff
@@ -215,55 +188,26 @@ for (const profile of profiles) {
...
+            const versions = await api.getPackageAvailableVersions(environment!, profile.packageName);
+            // Accept undefined until the API can distinguish unsupported lookups from command or network failures.
+            // TODO: Add that result distinction and make supported lookups strict in a follow-up PR.
+            if (versions === undefined) {
+                return;
+            }
             assert.ok(versions.length > 0, 'No package versions available');
         });
 
         suiteTeardown(async () => {
```
StellaHuang95, line 229: > **Issue** · Please address or respond

Profiles that declare version-lookup support should require a defined, non-empty result. Returning early on `undefined` lets command or network regressions pass while this coverage reports success.

.github/workflows/pr-check.yml
```diff
@@ -252,6 +252,9 @@ jobs:
...
     strategy:
```
StellaHuang95, line 258: > **Warning** · Non-blocking recommendation

This enables external package lifecycle tests for every OS/Python matrix entry rather than the former focused Linux 3.12 run. Please scope these network-dependent tests to a focused required job or subset, and mirror that scope in `push-check.yml`.

src/test/integration/environmentFixture.ts
```diff
@@ -0,0 +1,506 @@
...
+                    await assertPathMissing(prefix, `${request.name} environment was not removed`);
+                } catch (error) {
+                    cleanupErrors.push(toError(error));
+                }
```
StellaHuang95, line 147: > **Warning** · Non-blocking recommendation

`withTimeout` does not cancel `api.removeEnvironment`. After timing out, provider cleanup can mutate the same prefix while the original removal is still running. Avoid concurrent cleanup of the fixture path or make the removal cancellation-aware.

src/test/integration/environmentFixture.ts
```diff
@@ -0,0 +1,506 @@
...
+    let markerWritten = false;
+    let projectRootCreated = false;
+
+    const dispose = async (): Promise<void> => {
+        if (disposed) {
+            return;
+        }
+        disposed = true;
+        const cleanupErrors: Error[] = [];
+
```
StellaHuang95, line 95: > **Warning** · Non-blocking recommendation

Setting `disposed` before cleanup completes makes every later `dispose()` call a no-op after a partial cleanup failure. Preserve retryability, such as by tracking an in-flight disposal promise and marking completion only after successful cleanup.

src/test/integration/packageManager.integration.test.ts
edvilme: This is intentional for this PR. `getPackageAvailableVersions` currently uses `undefined` for both an unsupported lookup and a command or network failure, so requiring a defined result would fail without identifying which condition occurred. The test still requires a non-empty list whenever a result is returned. #1686 will add the result distinction so supported lookups can become strict.

.github/workflows/pr-check.yml
edvilme: The full OS/Python matrix is intentional for this PR. Package-manager invocation, environment discovery, and cleanup differ materially across Windows, macOS, and Linux, and the earlier failures were platform-specific. `pr-check.yml` and `push-check.yml` use the same matrix so both validation paths remain consistent.

src/test/integration/environmentFixture.ts
edvilme: Fixed in ebb222d. Cleanup now tracks whether API removal has settled, gives a timed-out removal a bounded settlement grace period, and skips provider/root deletion while that removal is still active. Direct cleanup only runs after the API promise has settled, so both paths cannot mutate the prefix concurrently.

src/test/integration/environmentFixture.ts
edvilme: Fixed in ebb222d. `dispose()` now shares an in-flight cleanup promise across concurrent calls, retains it after successful completion, and clears it after failure so a later call can retry any partial cleanup.

.github/workflows/pr-check.yml
edvilme: The full OS/Python matrix is intentional for this PR because package-manager invocation, discovery, and cleanup differ by platform and earlier failures were platform-specific. `push-check.yml` mirrors the same scope. The network exposure is accepted to make that coverage required rather than best-effort.

src/test/integration/packageManager.integration.test.ts
edvilme: Fixed in 0601d34. `waitForCondition` now has an explicit fail-fast mode, and package installation/removal polling uses it so a rejected `getPackages` call propagates its original error immediately instead of becoming a generic timeout. A focused unit test verifies error identity.

src/test/integration/packageManager.integration.test.ts
edvilme: Fixed in 0601d34. Package polling disables predicate-error retries, preserving and immediately surfacing the original `getPackages` rejection. The new unit test asserts the exact rejected error is retained.

src/test/integration/environmentFixture.ts
edvilme: Fixed in ebb222d. The fixture tracks API-removal settlement, waits through a bounded grace period after timeout, and skips provider/root deletion if the API operation remains active. Direct cleanup cannot race the original removal.

src/test/managers/common/packageManagerHeadlessConformance.unit.test.ts
edvilme: Fixed in 0601d34. The factory now returns named `pip`, `conda`, and `poetry` fields plus an `all` collection, so manager-specific tests no longer depend on array ordering.

src/test/integration/packageManager.integration.test.ts
```diff
@@ -215,55 +190,26 @@ for (const profile of profiles) {
...
-            const versions = await api.getPackageAvailableVersions(environment!, 'requests');
-            assert.ok(versions, `${profile.name} unexpectedly failed to retrieve package versions`);
+            const versions = await api.getPackageAvailableVersions(environment!, profile.packageName);
+            // Accept undefined until the API can distinguish unsupported lookups from command or network failures.
+            // TODO: Add that result distinction and make supported lookups strict in a follow-up PR.
+            if (versions === undefined) {
+                return;
+            }
             assert.ok(versions.length > 0, 'No package versions available');
         });
```
StellaHuang95, line 227: > **Issue** · Please address or respond

Profiles that declare version-lookup support still report success when the result is `undefined`, allowing command or network regressions to pass with no assertion. Require a defined, non-empty result; deferring this with a TODO does not resolve the documented no-partial-assertions rule violation.

[verified]

src/test/integration/environmentFixture.ts
```diff
@@ -0,0 +1,546 @@
...
+        if (cleanupErrors.length > 0) {
+            throw new Error(cleanupErrors.map((error) => error.message).join('\n'));
+        }
+    };
+
+    const dispose = (): Promise<void> => {
+        if (!disposePromise) {
+            disposePromise = cleanup().catch((error) => {
+                disposePromise = undefined;
```
StellaHuang95, line 202: > **Warning** · Non-blocking recommendation

[unverified] Clearing `disposePromise` permits a retry, but the original API removal may still be running after both timeouts, so the retry can start a second removal concurrently. Persist the outstanding removal promise across attempts and add a delayed-removal test confirming retries do not overlap.

[verified]

src/test/integration/packageManager.integration.test.ts
```diff
@@ -121,71 +113,50 @@ for (const profile of profiles) {
...
+            let packages: Package[] | undefined;
+            await waitForCondition(
+                async () => {
+                    packages = await api.getPackages(environment!, { skipCache: true });
+                    return packages?.some((pkg) => pkg.name.toLowerCase() === packageName) ?? false;
+                },
+                30_000,
                 'Package not installed',
+                1_000,
+                false,
```
StellaHuang95: > **Warning** · Non-blocking recommendation

Passing `retryOnError=false` makes lifecycle polling abort on the first transient `getPackages` rejection instead of using the 30-second retry window. Keep retrying transient list failures, or add a focused test demonstrating that immediate rejection is the intended contract.

[verified]

src/test/integration/environmentFixture.ts
edvilme: Fixed in 76c45af. API removal now uses a single-flight operation retained across disposal retries, so a still-running removal promise is reused rather than starting another removal. A delayed-operation unit test verifies retries share the same promise and invoke the operation once.

src/test/integration/packageManager.integration.test.ts
edvilme: Fixed in 26250c5. Unsupported lookup now throws `PackageVersionLookupNotSupportedError`, which is the only condition the integration test skips. A returned `undefined`, empty list, command failure, network failure, or parse failure now fails the test.

src/test/integration/packageManager.integration.test.ts
edvilme: Fixed in 76c45af. Lifecycle polling now keeps retrying transient `getPackages` rejections through the polling window and, if failures persist to timeout, rejects with the last original refresh error. Unit coverage verifies both retry count and error identity.

src/test/integration/packageManager.integration.test.ts
```diff
@@ -198,72 +168,47 @@ for (const profile of profiles) {
...
+                    30_000,
                     'Package not uninstalled',
+                    1_000,
+                    true,
+                    true,
                 );
             }
         });
 
         test(`${profile.name} Package Manager should list available package versions`, async function () {
```
StellaHuang95, line 192: > **Issue** · Please address or respond

Both configured profiles are expected to support version lookup, but any `PackageVersionLookupNotSupportedError` skips the test, so incorrect manager resolution or lost capability can leave CI green. Restore an explicit capability expectation and permit this skip only for profiles declared unsupported.

[verified]

edvilme: Fixed in a0fe872. Each profile now explicitly declares whether lookup is supported for the active tool version. Conda always requires support; Pip requires it for 21.2+, and an indeterminate Pip version fails. `PackageVersionLookupNotSupportedError` only skips when the profile explicitly expects unsupported lookup.

src/features/pythonApi.ts
```diff
@@ -326,7 +327,9 @@ export class PythonEnvironmentApiImpl implements PythonEnvironmentApi {
         await waitForEnvManagerId([context.envId.managerId]);
         const manager = this.envManagers.getPackageManager(context);
         if (!manager) {
-            return Promise.resolve(undefined);
+            throw new PackageVersionLookupNotSupportedError(
+                `No package manager supports version lookup for: ${context.envId.id}`,
+            );
         }
         return manager.getPackageAvailableVersions(context, packageName);
```
StellaHuang95: > **Issue** · Please address or respond

This changes the public API from resolving `undefined` to rejecting when no package manager supports lookup, breaking existing consumers despite the compatibility claim in `api.ts`. Preserve `undefined` at the existing API boundary or introduce the rejecting behavior through a versioned API.

[verified]

src/features/pythonApi.ts
```diff
@@ -326,7 +327,9 @@ export class PythonEnvironmentApiImpl implements PythonEnvironmentApi {
         await waitForEnvManagerId([context.envId.managerId]);
         const manager = this.envManagers.getPackageManager(context);
         if (!manager) {
-            return Promise.resolve(undefined);
+            throw new PackageVersionLookupNotSupportedError(
+                `No package manager supports version lookup for: ${context.envId.id}`,
+            );
         }
```
StellaHuang95, line 348: > **Issue** · Please address or respond

This still changes the public contract from resolving `undefined` to rejecting when lookup is unavailable, so existing consumers can receive unexpected unhandled rejections. Preserve `undefined` at the public boundary, or formally update and export the public error contract and adjust all consumers; make the corresponding correction in `InternalPackageManager`.

[verified]

StellaHuang95:
The public package-version lookup API now rejects where it previously resolved undefined, creating a breaking consumer contract change.

src/features/pythonApi.ts
edvilme: This is addressed in #1728

src/features/pythonApi.ts
edvilme: Fixed in #1728

eleanorjboyd (dismissed):
No approval-blocking issues found. The headless package operations suppress progress/error UI while preserving operational failures, and the integration cleanup uses randomized ownership markers plus strict path-containment checks before destructive removal. I also confirmed that both PR and push CI enable the network lifecycle coverage.

Non-blocking follow-up: additional focused cleanup tests for ownership mismatch, out-of-root prefixes, partial setup, and removal timeout would strengthen the fixture coverage.

eleanorjboyd (approved):

src/managers/builtin/utils.ts
```diff
@@ -200,7 +200,7 @@ async function execPipList(environment: PythonEnvironment, log?: LogOutputChanne
     try {
         return await runPython(
             environment.execInfo.run.executable,
-            ['-m', 'pip', 'list', '--format=json', ...(args ?? [])],
+            ['-m', 'pip', 'list', '--format=json', '--disable-pip-version-check', ...(args ?? [])],
```
edvilme, line 203: Fixes #1650

src/test/testUtils.ts
```diff
@@ -50,22 +52,34 @@ export async function waitForCondition(
...
+    rejectWithLastError: boolean = false,
 ): Promise<void> {
     return new Promise<void>((resolve, reject) => {
         const startTime = Date.now();
+        let lastError: unknown;
 
         const checkCondition = async () => {
             try {
                 const result = await condition();
+                lastError = undefined;
```
StellaHuang95, line 65: > **Warning** · Non-blocking recommendation

A mixed sequence of rejection followed by `false` clears `lastError`, so `rejectWithLastError` emits a generic timeout instead of the operational failure. Preserve the last rejection across later false results and add a reject-then-false regression test.

[verified]

src/test/integration/environmentFixture.ts
```diff
@@ -0,0 +1,562 @@
...
+                await vscode.workspace.fs.delete(projectUri, { recursive: true, useTrash: false });
+            } catch (error) {
+                cleanupErrors.push(toError(error));
+            }
+        }
+
+        if (cleanupErrors.length > 0) {
+            throw new Error(cleanupErrors.map((error) => error.message).join('\n'));
+        }
+    };
```
StellaHuang95, line 197: > **Warning** · Non-blocking recommendation

This fixture combines provisioning, configuration, ownership enforcement, timeout handling, subprocess termination, and cleanup state, while direct unit coverage exercises only the single-flight helper. Split the bounded-process/termination machinery from the environment lease and add focused tests for ownership rejection, partial setup, removal timeout, and fallback cleanup.

[verified]

StellaHuang95:
GitHub cannot anchor PR review comments to unchanged lines in the diff. Falling back to a general PR comment for [.github/workflows/pr-check.yml:L350](https://github.com/microsoft/vscode-python-environments/blob/eb998c545d6967cb20fb98fdfcb3f9bcf48d4ad9/.github/workflows/pr-check.yml#L350).

Network package operations and Conda provisioning now affect every OS/Python integration job, mirrored in `push-check.yml`, making unrelated integration coverage dependent on external indexes and channels. Move this coverage to a dedicated package-manager matrix or otherwise bound the added network exposure.

[verified]

StellaHuang95:
GitHub cannot anchor PR review comments to unchanged lines in the diff. Falling back to a general PR comment for [src/managers/builtin/pipPackageManager.ts:L128](https://github.com/microsoft/vscode-python-environments/blob/eb998c545d6967cb20fb98fdfcb3f9bcf48d4ad9/src/managers/builtin/pipPackageManager.ts#L128).

The headless/progress/error-notification policy is independently duplicated in the Pip, Conda, and Poetry managers. Extract a small shared operation runner so new managers cannot silently drift from the conformance contract.

[verified]

.github/instructions/testing-workflow.instructions.md
```diff
@@ -16,6 +16,10 @@ This guide covers the full testing lifecycle:
 4. **🛠️ Fixing Problems** - Resolve compilation and runtime issues
 5. **✅ Validation** - Ensure coverage and resilience
 
+## Learnings
+
```
StellaHuang95, line 20: > **Info** · Optional note

The learning ends with an unexplained `(1)` citation marker. Remove it or add the referenced footnote.

[verified]

StellaHuang95 (approved):

heejaechang (approved):
Approved via Review Center.

PR #1722, opened 2026-08-18
Add inline script activation-time discovery (PEP 723 PR 8/16)

StellaHuang95:
>
> Builds on the merged persistence work in #1697 and is rebased on current `main`.

### Roadmap context

This is **PR 8 of 16** in the PEP 723 inline-script roadmap. It makes extension-owned cached environments discoverable after activation; automatic script routing remains a separate follow-up.

|---|---|---|
| | PR 6: `create()` uv-install fallback | merged (#1696) |
| | PR 7: persistence (`get` / `set` + Memento) | merged (#1697) |
| | **PR 8: activation-time discovery** | **this PR** |
| | PR 9: route PEP 723 scripts to the inline manager | follow-up |

### Why this PR

PR 7 persists and lazily rehydrates a specific script's selected inline environment when the manager is asked for that script. The manager still has no global inventory, however: `getEnvironments()` returns `[]`, and cached environments are not published after an extension-host restart unless a script-specific lookup happens to reconstruct one.

This PR implements the activation-discovery portion of Q4 in the design:

- walk the versioned global cache without blocking extension activation;
- validate each cache entry before exposing it as a `PythonEnvironment`;
- publish add/remove events as valid entries appear or become definitively invalid;
- preserve previously known entries through locks and transient filesystem failures;
- retry bounded transient work without creating a permanent watcher or polling loop.

### What this PR does

**Defers discovery until after manager registration**

- Starts activation discovery with `setImmediate()` after the manager is registered.
- Leaves the feature-gate-off path unchanged: no manager registration, cache scan, timer, or filesystem work.
- Keeps extension activation non-blocking.

**Publishes a global discovered collection**

- `getEnvironments('all')` returns a copy of the validated cache collection.
- Other scopes remain empty because inline cache entries live in extension global storage rather than belonging to one workspace.
- Results are sorted deterministically.
- Collection reconciliation emits precise `EnvironmentChangeKind.add` and `remove` events only when manager identity, executable path, or Python version materially changes.

**Validates every cache entry before publication**

A candidate must:

1. be a normal directory rather than a symlink;
2. be unlocked and physically contained beneath the expected cache root;
3. have a valid `.meta.json` sidecar;
4. have an available cached launcher and base interpreter;
5. resolve into a real `PythonEnvironment`;
6. prove direct-child cache ownership through `sysPrefix`; and
7. match the Python release recorded in the sidecar.

Discovery is read-only. It does not delete, rebuild, or rewrite invalid entries.

**Distinguishes definitive invalidity from uncertainty**

- Missing, malformed, unowned, or version-mismatched entries are omitted.
- Locked entries and transient I/O/resolver/ownership failures preserve the previously published environment and request another activation pass.
- A missing cache root is treated as an empty cache.
- An unavailable cache root preserves the current collection and remains retryable.

**Handles concurrent cache changes**

- Activation scans compare an initial and final directory snapshot so an entry created during the scan triggers a follow-up pass.
- Final cache-root absence publishes an empty collection immediately; transient final-read failures preserve the prior collection until a retry.
- Per-entry filesystem fingerprints detect a cache directory rebuilt under the same key even when its name is unchanged.
- Locks use the cache entry name/hash as their identity.
- Published entries use that same cache-key identity rather than mixing lexical cache paths with canonical `sysPrefix` paths, avoiding false removal through symlink/junction path differences.
- Lock probing uses `lstat`; only `ENOENT` means unlocked. `EIO`, access failures, and other uncertain states fail closed.

**Coalesces refresh work without weakening activation discovery**

- Concurrent compatible refreshes share one scan.
- Activation joining an in-flight explicit refresh receives one snapshot-aware follow-up rather than accepting the explicit pass's weaker single-snapshot semantics.
- Explicit `refresh()` cancels activation retries, performs one settled pass, and schedules no delayed work afterward.

**Retries activation discovery only while useful**

- Follow-up delays are bounded at 1 second, 5 seconds, and 30 seconds.
- Retries are requested for locks, transient failures, or a changed cache snapshot.
- No permanent filesystem watcher or unbounded polling loop is introduced.
- Disposal cancels pending timers and prevents in-flight scans from publishing late results.

**Strengthens the Windows usability guard**

- Windows cache validation now checks both the cached environment launcher and the base interpreter referenced by `pyvenv.cfg`.
- A surviving base interpreter no longer makes an environment with a missing `Scripts\python.exe` appear usable.

### Discovery semantics

| Cache state | Behavior |
|---|---|
| Valid sidecar, launcher, interpreter, ownership, and version | Publish environment |
| Entry is locked or being built | Preserve previous publication; retry |
| Filesystem/resolver state is temporarily unavailable | Preserve previous publication; retry |
| Entry appears during the scan | Schedule snapshot-aware follow-up |
| Missing or malformed sidecar | Omit/remove from collection |
| Missing launcher/base interpreter | Omit/remove from collection |
| Ownership or recorded-version mismatch | Omit/remove from collection |
| Cache root is absent | Publish empty collection |

### Example

```text
extension activation
→ register InlineScriptEnvManager
→ defer one event-loop turn
→ scan <globalStorage>/script-envs-v1
→ validate sidecar + launcher + ownership + version
→ publish valid cached environments through getEnvironments('all')
→ retry only if the scan observed a lock, transient state, or snapshot change
```

This inventory does not associate an environment with a script. PR 7 owns persisted script associations, and PR 9 will use those associations for automatic per-file routing.

### Tests

Coverage includes:

- deferred feature-gated activation startup;
- valid cache discovery and `all`-scope publication;
- missing, malformed, unavailable, non-directory, and symlinked entries;
- missing and non-regular Windows launchers;
- lock preservation, including unavailable (`EIO`) lock probes;
- canonical `sysPrefix` versus lexical cache-root identities;
- add/remove event reconciliation;
- concurrent refresh coalescing;
- activation joining an explicit refresh;
- cache entries created during a scan;
- builds completing after the short retry window;
- bounded retry exhaustion;
- explicit single-pass refresh behavior; and
- disposal during in-flight scans and pending retries.

Validation on the rebased branch:

- `npm run compile-tests`
- `npm run compile`
- `npm run lint`
- focused activation-discovery/cache-launcher/registration suites

The full Windows unit run reaches 1613 passing and 5 pending; the existing concurrent `writeMetaJson` rename test can still intermittently fail with `EPERM` on Windows. That writer is unchanged by this PR and the same failure is reproducible on `main`.

### Performance

- Activation is deferred and never waits for discovery.
- Cache scans are coalesced.
- Retries are bounded and stop after a stable pass.
- Explicit refresh remains single-pass.
- There is no persistent watcher, unbounded polling, or per-document work.

### User impact

**No default-path user impact.** The manager and discovery remain behind the undeclared, default-off `python-envs.inlineScripts.enabled` flag.

With the internal flag manually enabled, valid cached inline environments become available through the manager after restart. This PR does not automatically select one for a script and introduces no public command, setting, picker item, project registration, cache deletion, TTL cleanup, or telemetry.

### Scope and follow-up

- automatic PEP 723 script routing (PR 9);
- exact script project registration (PR 10);
- CodeLens or bulk setup UX (PRs 11-12);
- cache clearing or TTL eviction (PRs 13-14); or
- lifecycle telemetry (PR 15).

PR 7 (#1697) is merged, and this branch is rebased on current `main`. Automatic routing can follow independently after this manager-discovery layer lands.

### DIFF
--- a/src/managers/builtin/inlineScript/envManager.ts
+++ b/src/managers/builtin/inlineScript/envManager.ts
@@ -11,6 +11,7 @@ import {
     DidChangeEnvironmentEventArgs,
     DidChangeEnvironmentsEventArgs,
     EnvironmentManager,
+    EnvironmentChangeKind,
     GetEnvironmentScope,
     GetEnvironmentsScope,
     IconPath,
@@ -60,6 +61,7 @@ import { normalizePath } from '../../../common/utils/pathUtils';
 import { compareReleaseSegments, parseReleaseSegments } from '../../../common/utils/pep440Release';
 import { getVenvPythonPath } from '../../../common/utils/virtualEnvironment';
 import { NativePythonFinder } from '../../common/nativePythonFinder';
+import { sortEnvironments } from '../../common/utils';
 import { resolveSystemPythonEnvironmentPath } from '../utils';
 import * as uvPythonInstaller from '../uvPythonInstaller';
 import {
@@ -78,6 +80,7 @@ const BASE_INTERPRETER_MANAGER_IDS = new Set([
 const CACHE_LOCK_TIMEOUT_MS = 5 * 60 * 1000;
 const CACHE_LOCK_RETRY_MS = 500;
 const CACHED_ASSOCIATION_VALIDATION_INTERVAL_MS = 5_000;
+const DISCOVERY_RETRY_DELAYS_MS = [1_000, 5_000] as const;
 /** Workspace-state key for PEP 723 script path to environment executable associations. */
 export const INLINE_SCRIPT_ENVS_KEY = `${ENVS_EXTENSION_ID}:inline-script:SCRIPT_ENVIRONMENTS`;
 
@@ -124,17 +127,23 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
     private readonly pendingCreations = new Map<string, Promise<PythonEnvironment | undefined>>();
     private readonly directlyResolvedBaseInterpreters = new Map<string, PythonEnvironment>();
     private baseInterpreterInstallationQueue: Promise<void> = Promise.resolve();
+    private collection: PythonEnvironment[] = [];
     private readonly pendingRehydrations = new Map<string, Promise<PythonEnvironment | undefined>>();
     private readonly fsPathToEnv = new Map<string, PythonEnvironment>();
     private readonly fsPathToPersistedEnvPath = new Map<string, string>();
     private readonly cachedAssociationValidatedAt = new Map<string, number>();
     private readonly associationRevisions = new Map<string, number>();
+    private pendingRefresh: Promise<boolean> | undefined;
+    private activationDiscoveryActive = false;
+    private discoveryRetryAttempt = 0;
+    private discoveryRetryTimer: ReturnType<typeof setTimeout> | undefined;
     private persistenceQueue: Promise<void> = Promise.resolve();
     private selectionQueue: Promise<void> = Promise.resolve();
     private cacheMaintenanceQueue: Promise<void> = Promise.resolve();
     private cacheMaintenanceBarrier: Deferred<void> | undefined;
     private pendingCacheMaintenances = 0;
     private activeCreateOperations = 0;
+    private disposed = false;
 
     private readonly _onDidChangeEnvironments = new EventEmitter<DidChangeEnvironmentsEventArgs>();
     public readonly onDidChangeEnvironments: Event<DidChangeEnvironmentsEventArgs> =
@@ -289,10 +298,17 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
     }
 
     async refresh(_scope: RefreshEnvironmentsScope): Promise<void> {
-        return;
+        if (this.disposed) {
+            return;
+        }
+        this.stopActivationDiscovery();
+        await this.getOrStartRefreshPass();
     }
 
-    async getEnvironments(_scope: GetEnvironmentsScope): Promise<PythonEnvironment[]> {
+    async getEnvironments(scope: GetEnvironmentsScope): Promise<PythonEnvironment[]> {
+        if (scope === 'all') {
+            return Array.from(this.collection);
+        }
         return [];
     }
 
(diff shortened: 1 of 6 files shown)


rchiodo:

src/managers/builtin/inlineScript/envManager.ts
```diff
@@ -247,6 +269,333 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
...
+        void followUp.then(
+            () => {
+                if (this.pendingSnapshotRefresh === followUp) {
+                    this.pendingSnapshotRefresh = undefined;
+                }
+            },
+            () => {
+                if (this.pendingSnapshotRefresh === followUp) {
+                    this.pendingSnapshotRefresh = undefined;
+                }
```
rchiodo, line 399: > **Warning** · Non-blocking recommendation

The final snapshot treats an absent cache root as retryable, unlike the initial read which correctly treats `ENOENT` as an empty cache. When the root remains absent, activation will unnecessarily schedule all 1s/5s/30s retries. Classify definitive stale-path errors here consistently with the initial read.

rchiodo (dismissed):
Approved via Review Center.

src/managers/builtin/inlineScript/envManager.ts
```diff
@@ -247,6 +269,333 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
...
+            },
+        );
+        return followUp;
+    }
+
+    private startSnapshotRefreshAfter(sharedPass: DiscoveryRefreshPass): Promise<boolean> {
+        return sharedPass.promise.then(() => {
+            if (this.disposed || !this.activationDiscoveryActive) {
+                return false;
```
edvilme: I don't think this means the `startSnapshotRefreshAfter` will return false. Maybe it would be better to construct a `Promise` object with `resolve` and `reject` parameters?

StellaHuang95: It does return false in this case, it waits for sharedPass.promise to finish and if the condition is met, finish the new promise with false. Maybe I will just use aysnc and await to be more clear.

StellaHuang95: Thanks for raising this. The Promise returned by `.then()` did adopt the callback's `false` or nested `Promise<boolean>`, so the existing behavior was correct and an explicit Promise constructor was not necessary. I agree the control flow was not immediately obvious, so a48066a rewrites it with `async`/`await` without changing the logic. The snapshot-follow-up and overlapping-refresh cancellation tests remain green.

StellaHuang95: oops, Copilot automatically replied to it as well.

src/managers/builtin/inlineScript/envManager.ts
```diff
@@ -247,6 +269,332 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
...
+            () => {
+                if (this.pendingSnapshotRefresh === followUp) {
+                    this.pendingSnapshotRefresh = undefined;
+                }
+            },
+            () => {
+                if (this.pendingSnapshotRefresh === followUp) {
+                    this.pendingSnapshotRefresh = undefined;
+                }
+            },
```
rchiodo, line 400: > **Issue** · Please address or respond

The final snapshot treats every `readdir` error as retryable. If the cache root is absent here, `ENOENT` is a definitive empty snapshot (as it is for the initial read), but this schedules all bounded retries. Classify stale-path errors consistently and compare an empty final snapshot with `entryNames`; retry only when the snapshot changed or the failure is genuinely uncertain.

StellaHuang95: Addressed in 8383af0. The final cache-root read now uses the same stale-path classification as the initial read: `ENOENT` and other definitive stale-path errors become an empty snapshot, while genuinely uncertain errors still request a retry. The empty final snapshot is compared with the initial entries, so a removed populated root triggers one follow-up but an empty/absent root schedules no bounded retries. Added a regression for initial `[]` followed by final `ENOENT`.

rchiodo (dismissed):

src/managers/builtin/inlineScript/envManager.ts
```diff
@@ -247,6 +269,339 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
...
+
+    private async inspectDiscoveredCacheEntry(
+        cacheRoot: Uri,
+        envDir: Uri,
+    ): Promise<DiscoveredCacheEntryResult> {
+        try {
+            const stat = await fs.lstat(envDir.fsPath);
+            if (!stat.isDirectory() || stat.isSymbolicLink()) {
+                return { kind: 'skip' };
+            }
```
rchiodo, line 578: > **Warning** · Non-blocking recommendation

If a populated cache root disappears after inspection, the definitive empty final snapshot requests a retry but `nextByKey` is still published first. This can temporarily publish environments whose cache directories have already been removed. Reconcile a definitively absent final root to an empty collection before publishing, and add a populated-then-`ENOENT` regression test.

src/managers/builtin/inlineScript/envManager.ts
```diff
@@ -247,6 +269,339 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
...
+        }
+
+        if (this.disposed) {
+            return false;
+        }
+
+        // Preserve previously known entries when a refresh cannot safely classify
+        // them because a build is in progress or the filesystem is transiently unavailable.
+        this.replaceDiscoveredEnvironments(sortEnvironments(Array.from(nextByKey.values())));
```
rchiodo, line 564: > **Warning** · Non-blocking recommendation

The final snapshot compares only entry names, so replacing an entry under the same cache key during inspection is treated as stable. If inspection observes a transient invalid state before the replacement completes, the valid replacement can remain omitted without a follow-up scan. Detect per-entry changes or recheck entry state, with a same-key rebuild race regression test.

eleanorjboyd (approved):

eleanorjboyd (dismissed):

src/managers/builtin/inlineScript/envManager.ts
```diff
@@ -283,6 +305,339 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
...
+
+    private async inspectDiscoveredCacheEntry(
+        cacheRoot: Uri,
+        envDir: Uri,
+    ): Promise<DiscoveredCacheEntryResult> {
+        try {
+            const stat = await fs.lstat(envDir.fsPath);
+            if (!stat.isDirectory() || stat.isSymbolicLink()) {
+                return { kind: 'skip' };
+            }
```
rchiodo, line 578: > **Issue** · Please address or respond

When the root existed for the initial snapshot but is definitively absent for the final `readdir`, this still publishes entries resolved from the now-deleted root and only schedules a retry. A missing cache root is defined as an empty cache, so reconcile this case to an empty collection before publishing and add a populated-then-`ENOENT` regression test.

src/managers/builtin/inlineScript/envManager.ts
```diff
@@ -283,6 +305,339 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
...
+        }
+
+        if (this.disposed) {
+            return false;
+        }
+
+        // Preserve previously known entries when a refresh cannot safely classify
+        // them because a build is in progress or the filesystem is transiently unavailable.
+        this.replaceDiscoveredEnvironments(sortEnvironments(Array.from(nextByKey.values())));
+        return shouldRetry;
```
rchiodo, line 565: > **Issue** · Please address or respond

The final snapshot compares only entry names, so replacing an entry under the same cache key during inspection is considered stable. If inspection observed the old transiently invalid entry, the valid replacement is omitted and no retry is scheduled. Track per-entry identity or revalidate changed entries, with a same-key rebuild regression test.

src/managers/builtin/inlineScript/envManager.ts
```diff
@@ -315,6 +337,339 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
...
+
+    private async inspectDiscoveredCacheEntry(
+        cacheRoot: Uri,
+        envDir: Uri,
+    ): Promise<DiscoveredCacheEntryResult> {
+        try {
+            const stat = await fs.lstat(envDir.fsPath);
+            if (!stat.isDirectory() || stat.isSymbolicLink()) {
+                return { kind: 'skip' };
+            }
```
rchiodo, line 578: > **Issue** · Please address or respond

When the final cache-root read reports `ENOENT`, this still publishes entries resolved from the initial snapshot, even though the root is now definitively empty; when that read fails transiently, it also publishes an unverified staged collection. Reconcile final absence to an empty collection, preserve the prior collection on a transient final read, and add populated-root regressions for both cases.

src/managers/builtin/inlineScript/envManager.ts
```diff
@@ -315,6 +337,339 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
...
+
+        // Preserve previously known entries when a refresh cannot safely classify
+        // them because a build is in progress or the filesystem is transiently unavailable.
+        this.replaceDiscoveredEnvironments(sortEnvironments(Array.from(nextByKey.values())));
+        return shouldRetry;
+    }
+
+    private async inspectDiscoveredCacheEntry(
+        cacheRoot: Uri,
+        envDir: Uri,
```
rchiodo, line 570: > **Issue** · Please address or respond

The final snapshot compares only entry names, so replacing an entry during inspection under the same cache key is treated as stable. If the first inspection skips the entry while it is being rebuilt and the replacement is complete by the final read, activation discovery stops without ever publishing it. Track per-entry identity or revalidate entries before publication, with a same-key rebuild regression test.

src/managers/builtin/inlineScript/envManager.ts
StellaHuang95: Addressed in bd7e082. A definitive missing final cache root now clears the staged collection before publication, so a previously populated root emits the removal immediately. It still requests one bounded follow-up when the initial snapshot was nonempty, allowing a concurrently recreated root to be discovered. Added a populated-initial-snapshot → final-`ENOENT` regression.

src/managers/builtin/inlineScript/envManager.ts
StellaHuang95: Addressed in bd7e082. Final `ENOENT` publishes an empty collection immediately. A transient final root read now returns retryable without calling `replaceDiscoveredEnvironments`, preserving the prior collection and avoiding unverified add/remove events. Added populated-root regressions for both final absence and final `EIO`.

src/managers/builtin/inlineScript/envManager.ts
StellaHuang95: Addressed in bd7e082. Discovery now captures a cheap filesystem fingerprint (`dev`, `ino`, birth/creation, change, and modification times) from the existing initial `lstat`, then rechecks each same-name entry before publication. A changed, missing, or unverifiable generation preserves the prior collection and schedules a bounded retry instead of publishing staged results. The regression replaces a directory under the same cache key while inspection is gated and verifies the follow-up publishes the new generation.

src/managers/builtin/inlineScript/envManager.ts
StellaHuang95: Addressed in bd7e082 with per-entry filesystem fingerprints rather than a second full environment resolution pass. Same-name directory replacement is detected before publication, staged results are withheld, and the next bounded pass validates and publishes the replacement. Added the requested same-key rebuild regression.

rchiodo (approved):

heejaechang (approved):
Approved via Review Center.

PR #1724, opened 2026-08-18
Add clear script environment cache command (PEP 723 PR 13/16)

StellaHuang95:
>
> Builds on the merged persistence work in #1697 and is rebased on current `main`.

### Roadmap context

This is **PR 13 of 16** in the PEP 723 inline-script roadmap. It adds the explicit, user-confirmed cache lifecycle operation that the later TTL work will reuse.

| Phase 5: Lifecycle and polish | PR | Status |
|---|---|---|
| | PR 7: persistence (`get` / `set` + Memento) | merged (#1697) |
| | **PR 13: clear inline-script cache** | **this PR** |
| | PR 14: opportunistic 14-day TTL eviction | follow-up |
| | PR 15: lifecycle telemetry | separate |
| | PR 16: status-bar decision | resolved; no code PR |

### Why this PR

The extension can create extension-owned inline-script environments and persist script associations, but it has no complete way to remove that state. Clearing only the files would leave Memento associations and `pythonProjects` entries pointing at deleted interpreters; clearing only settings would leave disk usage behind.

This PR adds one coordinated lifecycle operation covering:

- extension-owned cache entries;
- persisted and in-memory script associations;
- active selection events; and
- generated inline-script project settings.

Because this is destructive and the cache is shared by extension-host processes, the implementation is intentionally fail-closed around path ownership and locks.

### What this PR does

**Adds an internal, confirmation-gated clear command**

- Registers `python-envs.clearScriptEnvCache` only while the hidden inline-script feature flag is enabled.
- Does not contribute the command to `package.json` or the Command Palette before rollout.
- Shows a modal warning covering cached environments, associations, and project entries.
- Cancelling the prompt performs no filesystem, state, or settings changes.
- Runs cache cleanup before settings removal, so failed/partial cache cleanup does not silently rewrite project configuration.

**Keeps generic cache clearing behavior safe**

- The existing public `python-envs.clearCache` command continues to clear existing non-inline managers.
- It skips the preview inline manager because the generic path has no inline-specific confirmation or project-settings lifecycle.
- The dedicated command invokes the inline manager directly and performs the complete cleanup transaction.

**Serializes in-process maintenance**

- Adds a manager-local maintenance queue and barrier.
- `create()`, `get()`, and `set()` cannot observe or mutate half-cleared state.
- A clear request refuses to begin when creation already started.
- A creation request that arrives after clear begins waits for maintenance to settle.
- Multiple maintenance requests are serialized without globally serializing unrelated managers.

**Coordinates deletion across extension hosts**

- Acquires and holds each cache entry's cross-process lock through deletion.
- Classifies locks as missing, held, retained, stale, orphaned, malformed, or unavailable.
- Uses PID liveness to distinguish a live owner from a stale owner.
- Makes retained markers generation-specific by preserving the owner's PID/nonce.
- Reclaims only the exact stale/retained generation marker that was inspected.
- If another process replaces that generation before the atomic claim, reclamation loses safely and touches nothing.
- Ambiguous legacy fixed `retained` markers remain recognizable but are conservatively not reclaimed.

**Validates every destructive path**

Before deleting an entry, cleanup verifies that:

- global storage and `script-envs-v1` are normal directories rather than symlinks/junctions;
- the versioned cache root is the expected direct child of global storage;
- neither path is a filesystem root or dangerously shallow;
- physical `realpath` containment matches the expected ownership boundary;
- the cache root has not changed since the cleanup snapshot; and
- the target entry is a normal direct-child directory inside that same physical root.

The entry lock is acquired first, then root and entry ownership are revalidated immediately before removal.

**Keeps state consistent through partial failures**

- Attempts cache entries independently and records successful removals.
- Aggregates and surfaces deletion/persistence failures rather than returning success-shaped output.
- Invalidates only associations whose environment was removed or is definitively missing.
- Preserves associations for cache entries that could not safely be removed.
- Cancels pending rehydration for invalidated scripts.
- Clears warm environment and validation caches.
- Advances association revisions so stale async work cannot restore removed selections.
- Emits `onDidChangeEnvironment` only for selections actually invalidated.

**Removes generated inline project settings safely**

- Resolves `pythonProjects` entries independently from global, workspace, and workspace-folder sources.
- Removes only entries whose manager is the inline-script manager.
- Preserves non-inline duplicate entries and higher-precedence overrides.
- Handles same relative paths across multiple workspace roots.
- Aggregates global/workspace updates so each shared scope is written once.
- Unloads only loaded projects that have no remaining configuration source.

### Cleanup semantics

| Condition | Behavior |
|---|---|
| Entry is unlocked and physically owned | Lock, revalidate, delete |
| Lock belongs to a live process | Refuse that deletion |
| Exact stale/retained generation can be claimed | Reclaim, acquire a fresh lock, delete |
| Lock is unavailable, malformed, orphaned, or legacy-ambiguous | Preserve entry and surface failure |
| Root or entry is redirected/outside ownership boundary | Refuse deletion |
| One entry fails after another was removed | Preserve valid survivors; invalidate removed associations; report aggregate failure |
| Persistence update fails after disk cleanup | Keep in-memory state consistent and surface the persistence error |
| Cache is already absent | Clear stale associations safely; remain idempotent |

### Example

```text
Clear Script Environment Cache
→ modal confirmation
→ enter manager maintenance barrier
→ verify physical cache root
→ acquire exact per-entry lock
→ revalidate ownership immediately before deletion
→ delete safe entries
→ reconcile Memento + in-memory selections + events
→ remove generated inline pythonProjects settings
```

### Tests

Coverage includes:

- prompt cancellation and command ordering;
- generic clear behavior with the preview manager absent/present;
- in-process create/clear ordering;
- live, stale, retained, orphaned, malformed, unavailable, and legacy lock states;
- exact-generation reclamation and delayed-reclaimer/new-creator races;
- holding entry locks through deletion;
- unsafe, shallow, redirected, symlinked, and root-swapped paths;
- successful, missing-cache, idempotent, and partial-failure cleanup;
- persistence failures and pending-rehydration races;
- global/workspace/workspace-folder setting precedence;
- multi-root projects and same-path entries; and
- default-off command registration.

Validation on the rebased branch:

- `npm run compile-tests`
- `npm run compile`
- `npm run lint`
- focused lock/cache-clear/settings/command suites: 38 passing

The full Windows unit run reaches 1638 passing and 5 pending; the existing concurrent `writeMetaJson` rename test can still intermittently fail with `EPERM` on Windows. That writer is unchanged by this PR and the same failure is reproducible on `main`.

### Performance

- No activation scan, timer, or background maintenance is added.
- All work is initiated by the internal clear command.
- Per-entry locks avoid globally serializing independent environment creation across extension hosts.
- The maintenance barrier exists only inside the enabled inline manager and is active only during cleanup.

### User impact

**No default-path user impact.** The manager and command remain behind the undeclared, default-off `python-envs.inlineScripts.enabled` flag, and the command is not publicly contributed.

When the internal flag is manually enabled, the existing generic cache command still behaves as before for non-inline managers. Inline cleanup is available only through the dedicated confirmed lifecycle.

### Scope and follow-up

- automatic routing or project registration;
- activation-time discovery;
- silent/opportunistic deletion;
- TTL expiration; or
- lifecycle telemetry.

PR 14 will reuse this safety and state-cleanup foundation to remove entries whose `lastUsedAt` exceeds the planned 14-day TTL.

### DIFF
--- a/src/common/lockfile.apis.ts
+++ b/src/common/lockfile.apis.ts
@@ -162,6 +162,31 @@ export async function inspectFileLock(filePath: string, options?: InspectFileLoc
     return 'orphaned';
 }
 
+/**
+ * Move a stale or retained lock out of the lock name before a replacement owner is acquired.
+ * The rename prevents a newly-created lock from being removed based on an earlier inspection.
+ */
+export async function reclaimFileLock(filePath: string): Promise<boolean> {
+    const lockPath = getFileLockPath(filePath);
+    const state = await inspectFileLock(filePath);
+    if (state !== 'stale' && state !== 'retained') {
+        return false;
+    }
+
+    const quarantinedLockPath = `${lockPath}.reclaimed-${process.pid}-${crypto.randomBytes(16).toString('hex')}`;
+    try {
+        await fsapi.rename(lockPath, quarantinedLockPath);
+    } catch (error) {
+        if (hasErrorCode(error, 'ENOENT')) {
+            return false;
+        }
+        throw error;
+    }
+
+    await fsapi.remove(quarantinedLockPath);
+    return true;
+}
+
 export async function getProcessLiveness(pid: number): Promise<ProcessLiveness> {
     try {
         process.kill(pid, 0);
@@ -196,12 +221,16 @@ function hasErrorCode(error: unknown, code: string): boolean {
 }
 
 function parseOwnerPid(entry: string): number | undefined {
-    const match = entry.match(/^owner-(\d+)-/);
+    const match = entry.match(new RegExp(`^${escapeRegExp(FILE_LOCK_OWNER_MARKER_PREFIX)}(\\d+)-`));
     if (!match) {
         return undefined;
     }
     const pid = Number(match[1]);
     return Number.isSafeInteger(pid) && pid > 0 ? pid : undefined;
+}
+
+function escapeRegExp(value: string): string {
+    return value.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
 }
 
 function createLockError(message: string, code: string, lockPath: string): NodeJS.ErrnoException {
--- a/src/common/lockfile.apis.ts
+++ b/src/common/lockfile.apis.ts
@@ -16,13 +16,31 @@ export interface AcquiredFileLock {
     readonly retain: () => Promise<void>;
 }
 
+export const FILE_LOCK_DIR_SUFFIX = '.lock';
+export const FILE_LOCK_OWNER_MARKER_PREFIX = 'owner-';
+export const FILE_LOCK_RETAINED_MARKER = 'retained';
+
+export type ProcessLiveness = 'live' | 'dead' | 'unavailable';
+export type FileLockState = 'missing' | 'held' | 'retained' | 'stale' | 'orphaned' | 'malformed' | 'unavailable';
+
+export interface InspectFileLockOptions {
+    readonly checkProcessLiveness?: (pid: number) => Promise<ProcessLiveness>;
+}
+
 type LockState = 'held' | 'released' | 'retained';
+
+export function getFileLockPath(filePath: string): string {
+    return `${path.resolve(filePath)}${FILE_LOCK_DIR_SUFFIX}`;
+}
 
 /** Acquire an atomic lock released only explicitly; interrupted operations remain locked. */
 export async function acquireFileLock(filePath: string, options: AcquireFileLockOptions): Promise<AcquiredFileLock> {
-    const lockPath = `${path.resolve(filePath)}.lock`;
-    const ownerMarker = path.join(lockPath, `owner-${process.pid}-${crypto.randomBytes(16).toString('hex')}`);
-    const retainedMarker = path.join(lockPath, 'retained');
+    const lockPath = getFileLockPath(filePath);
+    const ownerMarker = path.join(
+        lockPath,
+        `${FILE_LOCK_OWNER_MARKER_PREFIX}${process.pid}-${crypto.randomBytes(16).toString('hex')}`,
+    );
+    const retainedMarker = path.join(lockPath, FILE_LOCK_RETAINED_MARKER);
     const deadline = Date.now() + options.timeoutMs;
 
     while (true) {
@@ -100,9 +118,68 @@ export async function acquireFileLock(filePath: string, options: AcquireFileLock
     }
 }
 
+export async function inspectFileLock(filePath: string, options?: InspectFileLockOptions): Promise<FileLockState> {
+    const lockPath = getFileLockPath(filePath);
+
+    let stat;
+    try {
+        stat = await fsapi.lstat(lockPath);
+    } catch (error) {
+        if (hasErrorCode(error, 'ENOENT')) {
+            return 'missing';
+        }
+        throw error;
+    }
+
+    if (!stat.isDirectory() || stat.isSymbolicLink()) {
+        return 'malformed';
+    }
+
+    const entries = await fsapi.readdir(lockPath);
+    const ownerEntries = entries.filter((entry) => entry.startsWith(FILE_LOCK_OWNER_MARKER_PREFIX));
+    const retainedEntries = entries.filter((entry) => entry === FILE_LOCK_RETAINED_MARKER);
+    const unknownEntries = entries.filter(
+        (entry) => !entry.startsWith(FILE_LOCK_OWNER_MARKER_PREFIX) && entry !== FILE_LOCK_RETAINED_MARKER,
+    );
+
+    if (unknownEntries.length > 0 || ownerEntries.length > 1 || retainedEntries.length > 1) {
+        return 'malformed';
+    }
+    if (retainedEntries.length === 1) {
+        return 'retained';
+    }
+    if (ownerEntries.length === 1) {
+        const ownerPid = parseOwnerPid(ownerEntries[0]);
+        if (ownerPid === undefined) {
+            return 'malformed';
+        }
+        const liveness = await (options?.checkProcessLiveness ?? getProcessLiveness)(ownerPid);
+        if (liveness === 'dead') {
+            return 'stale';
+        }
+        return liveness === 'live' ? 'held' : 'unavailable';
+    }
+    return 'orphaned';
+}
+
+export async function getProcessLiveness(pid: number): Promise<ProcessLiveness> {
+    try {
+        process.kill(pid, 0);
+        return 'live';
+    } catch (error) {
+        if (hasErrorCode(error, 'ESRCH')) {
+            return 'dead';
+        }
+        if (hasErrorCode(error, 'EPERM') || hasErrorCode(error, 'EACCES')) {
+            return 'unavailable';
+        }
+        return 'unavailable';
+    }
+}
+
 async function isRetainedLock(lockPath: string): Promise<boolean> {
     try {
-        await fsapi.lstat(path.join(lockPath, 'retained'));
+        await fsapi.lstat(path.join(lockPath, FILE_LOCK_RETAINED_MARKER));
         return true;
     } catch (error) {
         if (hasErrorCode(error, 'ENOENT')) {
@@ -116,6 +193,15 @@ function hasErrorCode(error: unknown, code: string): boolean {
     return (
         typeof error === 'object' && error !== null && 'code' in error && (error as NodeJS.ErrnoException).code === code
     );
+}
+
+function parseOwnerPid(entry: string): number | undefined {
+    const match = entry.match(/^owner-(\d+)-/);
+    if (!match) {
+        return undefined;
+    }
+    const pid = Number(match[1]);
+    return Number.isSafeInteger(pid) && pid > 0 ? pid : undefined;
 }
 
 function createLockError(message: string, code: string, lockPath: string): NodeJS.ErrnoException {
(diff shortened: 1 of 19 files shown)


rchiodo:

src/features/settings/settingHelpers.ts
```diff
@@ -349,6 +471,178 @@ export interface EditProjectSettings {
...
+    return settings?.map((setting) => ({ ...setting }));
+}
+
+export async function removeInlineScriptPythonProjectSettings(
+    currentProjects: readonly PythonProject[],
+): Promise<PythonProject[]> {
+    const currentProjectsByUri = new Map(currentProjects.map((project) => [project.uri.toString(), project] as const));
+    const workspaceEntries: Array<readonly [WorkspaceFolder, EditProjectSettings[]]> = [];
+    for (const workspaceFolder of workspaceApis.getWorkspaceFolders() ?? []) {
+        const edits: EditProjectSettings[] = getResolvedPythonProjectSettings(workspaceFolder)
```
rchiodo, line 538: > **Warning** · Non-blocking recommendation

When there are no workspace folders, this returns before inspecting configuration, so global `pythonProjects` entries for the inline manager remain after clearing the cache. Handle global configuration independently of workspace-folder iteration and add an empty-window regression test.

rchiodo (dismissed):
Approved via Review Center.

src/common/lockfile.apis.ts
```diff
@@ -16,13 +16,33 @@ export interface AcquiredFileLock {
...
+
 /** Acquire an atomic lock released only explicitly; interrupted operations remain locked. */
 export async function acquireFileLock(filePath: string, options: AcquireFileLockOptions): Promise<AcquiredFileLock> {
-    const lockPath = `${path.resolve(filePath)}.lock`;
-    const ownerMarker = path.join(lockPath, `owner-${process.pid}-${crypto.randomBytes(16).toString('hex')}`);
-    const retainedMarker = path.join(lockPath, 'retained');
+    const lockPath = getFileLockPath(filePath);
+    const ownerMarker = path.join(
+        lockPath,
+        `${FILE_LOCK_OWNER_MARKER_PREFIX}${process.pid}-${crypto.randomBytes(16).toString('hex')}`,
```
edvilme: Should we store this in a variable for debugging?

edvilme: Or maybe extract it to a function like the other one here?

src/features/settings/settingHelpers.ts
```diff
@@ -349,6 +471,178 @@ export interface EditProjectSettings {
...
+                envManager: INLINE_SCRIPT_MANAGER_ID,
+            }));
+
+        if (edits.length > 0) {
+            workspaceEntries.push([workspaceFolder, edits]);
+        }
+    }
+
+    if (workspaceEntries.length === 0) {
+        return [];
```
rchiodo, line 558: > **Issue** · Please address or respond

When no workspace folders are open, this function returns without inspecting or updating global `pythonProjects`, leaving stale inline-script entries after cache deletion. Handle global configuration independently of workspace-folder iteration and add an empty-window regression test.

StellaHuang95: Addressed in 826fdd8. Global `pythonProjects` cleanup now runs independently of workspace-folder iteration: it reads user/global settings up front, removes all inline-manager entries, preserves non-inline entries, and awaits the global update even when no workspace folders or resolved workspace edits exist. Added an empty-window regression with one inline and one non-inline global entry.

src/managers/builtin/inlineScript/envManager.ts
```diff
@@ -129,46 +147,53 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
...
     ): Promise<PythonEnvironment | undefined> {
+        this.activeCreateOperations += 1;
         try {
-            const scriptUri = this.getScriptUri(scope);
-            if (!scriptUri) {
-                this.log.warn('Inline-script environment creation requires exactly one local file URI.');
-                return undefined;
-            }
+            return await this.waitForCacheMaintenance(async () => {
+                try {
```
rchiodo, line 171: > **Warning** · Non-blocking recommendation

`activeCreateOperations` is incremented before waiting on the maintenance barrier. A create queued behind one clear makes a subsequent queued clear reject even though that create has not started. Increment the counter after the barrier admits the create, and cover clear → create → clear ordering.

src/features/settings/settingHelpers.ts
```diff
@@ -10,6 +15,123 @@ import { sendTelemetryEvent } from '../../common/telemetry/sender';
...
+    const resolvedWorkspaceFolder = setting.workspace
+        ? allWorkspaceFolders.find((candidate) => candidate.name === setting.workspace)
+        : workspaceFolder;
+    return resolvedWorkspaceFolder
+        ? Uri.file(path.resolve(resolvedWorkspaceFolder.uri.fsPath, setting.path))
+        : undefined;
+}
+
+function getResolvedPythonProjectSettings(
+    workspaceFolder: WorkspaceFolder,
```
rchiodo, line 69: > **Info** · Optional note

Workspace-name and relative-path resolution is duplicated between these helpers, creating a maintainability risk if project-setting identity rules evolve. Consolidate this behind one canonical resolver.

src/features/envManagers.ts
```diff
@@ -320,12 +320,16 @@ export class PythonEnvironmentManagers implements EnvironmentManagers {
...
+                    .map((manager) => manager.clearCache()),
+            );
             return;
         }
 
         const manager = this.getEnvironmentManager(scope);
-        if (manager) {
+        if (manager && manager.id !== INLINE_SCRIPT_MANAGER_ID) {
             await manager.clearCache();
         }
```
rchiodo, line 334: > **Warning** · Non-blocking recommendation

An explicit `clearCache(INLINE_SCRIPT_MANAGER_ID)` request now succeeds without clearing anything. Reject with a lifecycle-required error, or represent the unsupported operation explicitly in the manager API so callers do not receive success-shaped output for a skipped clear.

rchiodo (dismissed):

eleanorjboyd (dismissed):

rchiodo (approved):

heejaechang (approved):
Approved via Review Center.

src/managers/builtin/inlineScript/envManager.ts
```diff
@@ -1327,6 +1386,414 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
...
+        const envDirPath = path.join(originalPhysicalCacheRootPath, entryName);
+        let lock: AcquiredFileLock | undefined;
+        try {
+            lock = await this.acquireCacheEntryLockForClear(envDirPath);
+            const currentPhysicalCacheRootPath = await this.getPhysicalOwnedCacheRootPath(cacheRoot);
+            if (!currentPhysicalCacheRootPath) {
+                return undefined;
+            }
+            if (
+                normalizePath(currentPhysicalCacheRootPath) !== normalizePath(originalPhysicalCacheRootPath)
```
eleanorjboyd, line 1500: **High — guard against same-path cache-root replacement**\n\nThis revalidation compares only canonical path strings. If another process removes and recreates the cache root at the same pathname after the initial enumeration, the comparison still succeeds, and cleanup can lock and delete an entry from the replacement root generation. Path containment protects where deletion happens, but not which filesystem generation owns that path. Please coordinate cleanup and creation with a root-generation lock/identity token and revalidate that identity immediately before recursive deletion.

src/common/lockfile.apis.ts
```diff
@@ -100,10 +107,141 @@ export async function acquireFileLock(filePath: string, options: AcquireFileLock
...
+    ) {
+        return false;
+    }
+
+    const claimedMarker = path.join(
+        lockPath,
+        `.reclaim-${process.pid}-${crypto.randomBytes(16).toString('hex')}-${snapshot.marker}`,
+    );
+    try {
+        await fsapi.rename(path.join(lockPath, snapshot.marker), claimedMarker);
```
eleanorjboyd, line 204: **Medium — make interrupted reclaim markers recoverable**\n\nAfter this rename, a crash—or an error during the following unlink/rmdir—can leave `.reclaim-*` behind. `inspectFileLockSnapshot()` treats that name as unknown, so every later inspection reports `malformed` and the cache entry can no longer be reclaimed automatically. Please either model reclaim markers as a recognized generation with PID/liveness recovery, or restore the original observed marker when reclamation does not complete.

src/features/settings/settingHelpers.ts
```diff
@@ -349,6 +471,175 @@ export interface EditProjectSettings {
...
+            ...globalValueRemaining,
+            ...workspaceValueRemaining,
+            ...((folderRemainingSettings.get(workspaceFolder.uri.toString()) ?? [])),
+        ];
+        edits.filter(
+            (edit) =>
+                existingSettings.some((projectSetting) => matchesProjectSettingEdit(projectSetting, edit, workspaceFolder)) &&
+                !hasProjectSetting(remainingSettings, edit.project, workspaceFolder),
+        ).forEach((edit) => {
+            removedProjects.set(edit.project.uri.toString(), edit.project);
```
eleanorjboyd, line 632: **Medium — do not remove an intrinsic workspace-root project**\n\nAn inline-script override such as `{ path: ".", envManager: "ms-python.python:inline-script" }` resolves to the workspace root. Once the override is removed, this code adds that URI to `removedProjects`; the command then calls `wm.remove()`, even though workspace roots are intrinsic projects created independently of settings. This temporarily drops the workspace root from the in-memory project model until a later refresh. Please exclude workspace-folder roots from `loadedProjectsToRemove` (or recompute removal from the effective project set after edits).

eleanorjboyd (commented):
Post-merge review: three concurrency/project-model findings from the clear-cache implementation.

PR #1725, opened 2026-08-18
Fix flaky Python project environment identity assertions

StellaHuang95:
## Context

The macOS 3.12 integration job in recent prs failed in `Integration: Python Projects > File in project uses project environment`

The assertion showed the same Python version with two different generated suffixes:

```text
actual:   Python 3.13.14-cnykpnoi58
expected: Python 3.13.14-ul2l97bdet
```

## Root cause

`PythonEnvironmentApiImpl.createPythonEnvironmentItem()` appends `Math.random()` when creating `envId.id`. A manager may rematerialize the same logical interpreter from its persisted path during `getEnvironment()`, producing a fresh object and random ID.

The Python Projects integration suite compared `envId.id` to verify:

- candidate environments differ;
- the expected change event arrived;
- project readback returns the selected environment;
- clearing starts from the selected environment;
- a file inherits its containing project's environment.

Those checks can fail even when the manager, executable path, and Python version are all correct.

## Fix

Use one logical-environment identity throughout this suite:

```text
manager ID + normalized environment path + version
```

The shared helper now drives candidate selection, event filtering, and all project/file readback assertions. The event helper still subscribes before `setEnvironment()`, so this does not reintroduce the settings-persistence race that the event-driven test was designed to avoid.

The event-specific test also reuses the URI-and-environment-filtered helper instead of accepting any later event with a non-empty `new` value.

## Why this is test-only

This suite verifies project association and inheritance, not the long-term stability contract of `envId.id`. Changing production ID generation here would be a broad public-API and cache-semantics change: environment IDs key package caches, watchers, terminals, tree items, and active-selection equality. Managers also do not consistently represent `environmentPath` as the same kind of path.

Production `envId` stability is worth evaluating separately. This focused fix keeps the integration assertion strict while avoiding an unrelated cross-cutting behavior change.

## Validation

- `npm run compile-tests`
- ESLint on the changed test
- targeted `Integration: Python Projects` suite: **8 passing**
- targeted suite repeated **7 times** successfully in the VS Code integration harness
- independent code review completed with no significant findings

A full local integration run reached and passed the changed Python Projects suite, then failed later in unrelated environment-creation tests. The separate Ubuntu package-network failure from the same CI run is not addressed by this PR.

### DIFF
--- a/src/test/integration/pythonProjects.integration.test.ts
+++ b/src/test/integration/pythonProjects.integration.test.ts
@@ -20,16 +20,49 @@
 import * as assert from 'assert';
 import * as vscode from 'vscode';
 import { PythonEnvironment, PythonEnvironmentApi } from '../../api';
+import { normalizePath } from '../../common/utils/pathUtils';
 import { ENVS_EXTENSION_ID } from '../constants';
-import { TestEventHandler, waitForCondition } from '../testUtils';
+import { waitForCondition } from '../testUtils';
 
 const ENV_CHANGE_TIMEOUT_MS = 15_000;
+
+function getEnvironmentIdentity(environment: PythonEnvironment) {
+    return {
+        managerId: environment.envId.managerId,
+        path: normalizePath(environment.environmentPath.fsPath),
+        version: environment.version,
+    };
+}
+
+function isSameEnvironment(
+    first: PythonEnvironment | undefined,
+    second: PythonEnvironment | undefined,
+): boolean {
+    if (!first || !second) {
+        return first === second;
+    }
+    const firstIdentity = getEnvironmentIdentity(first);
+    const secondIdentity = getEnvironmentIdentity(second);
+    return (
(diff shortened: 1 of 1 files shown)


rchiodo:

src/test/integration/pythonProjects.integration.test.ts
```diff
@@ -20,16 +20,49 @@
...
 
 const ENV_CHANGE_TIMEOUT_MS = 15_000;
 
+function getEnvironmentIdentity(environment: PythonEnvironment) {
+    return {
+        managerId: environment.envId.managerId,
+        path: normalizePath(environment.environmentPath.fsPath),
+        version: environment.version,
+    };
+}
```
edvilme, line 35: Could/should we somehow hash them so we only have to compare the hashes?

edvilme (approved):

rchiodo (approved):
Approved via Review Center.

PR #1729, opened 2026-08-19
Route PEP 723 scripts to inline environments (PEP 723 PR 9/16)

StellaHuang95:
### Roadmap context

This is **PR 9 of 16** in the PEP 723 inline-script roadmap and the repository's intentional routing cutover. Earlier manager PRs create, discover, validate, and persist environments; this PR makes normal per-file environment lookup use them when a saved script association is proven current.

| Phase 2/3 | PR | Status |
|---|---|---|
| | PR 7: per-script persistence | merged (#1697) |
| | PR 8: activation-time cache discovery | merged (#1722) |
| | **PR 9: automatic per-script routing** | **this PR** |
| | PR 10: exact script project registration | follow-up |
| | PRs 11-12: CodeLens and bulk setup UX | follow-up |
| | PRs 17-19: Pylance/Python debugger integration | cross-repository follow-up |

### Why this PR

The inline manager can already create/reuse environments, persist a script association, rediscover cache entries after restart, clear state safely, and report lifecycle telemetry. Nothing automatically selects that manager for a Python file, however. A script continues to use its normal project/default environment unless another component directly asks the inline manager.

Automatic routing must be stricter than checking whether a file contains PEP 723 metadata or whether Memento contains an executable path. It must prove both:

```text
current saved PEP 723 metadata exists
                  +
a persisted environment association is valid for that metadata and cache generation
                  =
route this file to InlineScriptEnvManager
```

If either proof is absent, dirty, stale, transiently unavailable, or superseded by an explicit choice, routing falls through to the existing project/default manager.

### What this PR does

**Adds an activation-scoped routing registry**

- Latches the hidden feature flag once per extension-host activation.
- Creates one registry shared by the detector, inline manager, and central environment-manager router.
- Tracks saved metadata, normalized metadata identity, metadata revision, and validated-association state independently per script.
- Emits metadata and routeability transitions.
- Requires both current metadata and a validated association before `shouldRoute(uri)` becomes true.

**Turns the existing lazy detector into a routing input only when enabled**

- Reads saved PEP 723 metadata on open and save.
- Replays documents that were already open when activation occurred.
- Tracks loose local `.py` files for routing while retaining the existing workspace-only telemetry population.
- Withholds routing for restored dirty documents.
- Invalidates routing immediately when an edit can touch the metadata block, while body-only edits retain routing.
- Uses source-compatible offsets so BOM- and CRLF-prefixed metadata edits are classified correctly.
- Uses per-URI read generations so an older open read cannot overwrite newer saved metadata.
- Clears metadata and associations for deleted or renamed paths.

When no routing registry exists, the detector retains its existing telemetry-only listeners, coalescing, workspace filter, and event behavior.

**Defines a stable metadata identity**

- Uses normalized dependencies plus trimmed `requires-python`.
- Ignores dependency ordering, equivalent package-name spelling, and the unrelated `[tool]` table.
- Does not include script paths, so equivalent scripts can share cache entries.

**Adds durable cache provenance**

- Sidecars store bounded SHA-256 hashes of metadata identities that were proven for the cache entry.
- Same-key coalesced creators and cache-reuse callers merge their identities under the existing cache-entry lock.
- The list is deduplicated and capped at 128 entries.
- No package, requirement, script path, URI, or metadata content is stored in provenance.
- Older sidecars without provenance remain usable through conservative cache-key and Python-constraint proof; additional-package environments remain non-routeable until explicitly proven.

**Makes sidecar replacement recoverable**

- Serializes same-sidecar writes in-process; production callers remain protected by the cross-process cache-entry lock.
- Uses native rename with a unique backup on Windows replacement contention.
- Restores the previous sidecar after failed replacement and retains the only known-good backup if restoration is uncertain.
- Lock-held cache inspection can recover a strict, regular, size-bounded, schema-valid backup compatible with the selected base interpreter.
- Invalid, temporary, symlinked, oversized, unsupported, or incompatible-only artifacts retain normal stale/uncertain behavior.

**Upgrades persisted associations from path-only values**

Each current record contains:

```text
schemaVersion
environmentPath
metadataBinding: legacy | pending(sourceIdentity) | matched(sourceIdentity)
```

- `legacy`: old string association; remains retrievable but is not automatically routeable.
- `pending`: the environment is proven, but saved metadata has not yet been durably matched (for example, selection while the document is dirty).
- `matched`: saved metadata identity and cache provenance agree.
- Future schema records are preserved rather than destructively rewritten.
- Malformed requested entries can be repaired without discarding unrelated valid/future records.

**Validates associations before routing**

Validation requires:

1. current saved metadata;
2. an absolute persisted executable path;
3. no active cache build/lock;
4. a resolvable environment;
5. physical ownership by the expected extension cache entry;
6. compatible Python/requires-python state;
7. a matching persisted metadata binding; and
8. sidecar proof for the current metadata identity.

Definitively stale associations are conditionally removed. Locked, transient, uncertain, or future-schema states are preserved but remain non-routeable.

**Protects asynchronous validation from stale results**

- Metadata revisions remain monotonic across cleared-state tombstones, preventing an old validator from winning after clear/restore of the same identity.
- A changed saved metadata identity invalidates routeability synchronously before provenance validation begins.
- Association revisions ensure unset/replacement wins over old rehydration or validation.
- Same-script rehydration and metadata refreshes are coalesced.
- Pending bindings are promoted to matched only if both metadata and association revisions are still current.
- Failed persistence never publishes successful routeability.

**Adds central manager routing with explicit precedence**

| Priority | Manager source |
|---|---|
| 1 | Exact per-script project setting |
| 2 | Explicit in-session non-inline override |
| 3 | Validated inline-script association |
| 4 | Existing project/default manager setting |
| 5 | Existing cached project/global manager |

- A user explicitly selecting a non-inline manager is not immediately overridden by automatic routing.
- Clearing that override re-resolves and publishes through the newly effective manager, with reserved operation ordering so a newer selection still wins.
- Inline selections are keyed by normalized script path, not containing project, so multiple scripts in one workspace can remain independent.
- Inline selections are not published as active until routeability is validated.
- Routeability changes refresh the appropriate manager and emit the existing active-environment transition.
- Invalidated metadata falls back to the existing project/default environment and removes only the inline active-selection entry.
- Existing revision checks continue preventing slow refreshes from overwriting newer selections.

### User flows

**Previously configured script after restart**

```text
open saved script
→ detector publishes metadata identity
→ manager loads persisted association
→ validate executable + ownership + sidecar provenance
→ registry marks association routeable
→ central environment lookup switches this file to the inline manager
```

**Metadata edit**

```text
edit inside metadata block
→ routeability clears immediately
→ normal project/default manager remains active
→ save reads new metadata and revalidates
→ inline routing returns only if the existing association is still proven
```

**New unassociated script**

Opening a new PEP 723 script does **not** silently select a matching cache entry. It remains on normal routing until the future explicit setup action (PR 11/12) creates or reuses an environment and persists the association. After that, this PR provides automatic routing.

### Routing and failure semantics

| State | Behavior |
|---|---|
| Saved metadata + matched/proven association | Route inline |
| Metadata but no association | Existing project/default routing |
| Association but metadata unknown/dirty | Existing project/default routing |
| Body-only edit | Preserve current routing |
| Metadata edit or invalid saved block | Clear inline routing and fall back |
| Explicit non-inline selection | Explicit override wins |
| Legacy string association | Preserve/retrieve, but do not auto-route after restart |
| Pending binding with matching saved proof | Atomically promote to matched |
| Locked/transient cache state | Preserve association; remain non-routeable; retry later |
| Same cache path rebuilt for different provenance | Reject old routing proof |
| Rename/delete | Clear old path's association and routeability |

### Review guide

The production changes are easiest to review in this order:

1. **Routing state and detection**
   - `routingRegistry.ts`
   - `activation.ts`
   - `metadata.ts`
   - `lazyDetector.ts`
2. **Cache provenance and durability**
   - `cacheLayout.ts`
   - creation/reuse sections of `inlineScript/envManager.ts`
3. **Persisted binding and validation state machine**
   - persistence/rehydration/metadata-refresh sections of `inlineScript/envManager.ts`
4. **Behavior cutover**
   - `envManagers.ts`
   - `extension.ts`
   - `inlineScript/main.ts`

More than half of the diff is deterministic unit coverage for dirty/save/restart and async race behavior.

### Tests

Coverage includes:

- activation flag latching and default-off registration;
- telemetry-only detector equivalence when routing is absent;
- open/save ordering, restored dirty editors, loose files, CRLF edits, body-only edits, rename, and delete;
- metadata-only, association-only, exact-setting, explicit-override, route-on, and fallback behavior;
- independent same-workspace script selections and batch operations;
- legacy, pending, matched, malformed, and future persisted records;
- restart rehydration and delayed routeability;
- same-path cache rebuilds and additional-package provenance;
- same-key coalesced creation and one-event telemetry semantics;
- stale metadata/association races against save, unset, and replacement;
- sidecar write contention, restoration, backup recovery, and invalid artifact handling;
- cache-clear partial failure with versioned associations; and
- public API last-known fallback/event ordering.

Validation on the final rebased tree:

- `npm run compile-tests`
- `npm run compile`
- `npm run lint`
- `npm run unittest`: **1,817 passing, 6 pending**

### Performance
- The feature remains hidden and default-off.
- With the flag off, no routing registry, manager, discovery timer, routing file listeners, Memento read, sidecar/cache work, or routing telemetry is added.
- The existing telemetry detector keeps its prior listener/coalescing behavior; routing checks are optional branches.
- With the flag on, metadata reads remain bounded to the first 8 KiB and occur only for opened/saved local Python files.
- Same-script validation and same-key creation are coalesced.
- Warm association validation is throttled.
- Discovery retries remain bounded and cache maintenance remains serialized only within the inline manager.

### Privacy and safety

- Telemetry remains count/boolean/category based and sends no URI, path, dependency, requirement, Python version, cache key, or error text.
- Provenance uses local SHA-256 metadata identities rather than raw metadata.
- Physical cache ownership and executable checks remain fail-closed.
- Symlinked/unowned cache entries are never made routeable.
- Destructive cache behavior remains exclusively in the merged confirmation-gated cleanup lifecycle.

### User impact

**No default-path user impact.** `python-envs.inlineScripts.enabled` remains undeclared and defaults to false:

- no visible setting or autocomplete;
- no new command, menu, CodeLens, picker, or status-bar surface;
- no inline manager registration;
- no routing registry, discovery, persistence, or cache work; and
- existing project/default routing remains unchanged.

For developers manually enabling the hidden flag, existing proven script associations route automatically and fall back conservatively when proof is absent. New scripts still require the future explicit setup UX.

### Scope and follow-up

- automatic setup for a newly encountered script;
- exact generated script project registration (PR 10);
- CodeLens or bulk setup UX (PRs 11-12);
- Pylance per-file interpreter support (PRs 17-18); or
- the Python debugger per-file resolver fix (PR 19).

Those later PRs can build on this guarded routing layer without changing its validation contract.

### DIFF
--- a/src/common/inlineScript/cacheLayout.ts
+++ b/src/common/inlineScript/cacheLayout.ts
@@ -444,7 +444,11 @@ function validateMeta(value: unknown): InlineScriptEnvMeta | 'unsupported' | und
         return undefined;
     }
     const obj = value as Record<string, unknown>;
-    if (typeof obj.schemaVersion !== 'number') {
+    if (
+        typeof obj.schemaVersion !== 'number' ||
+        !Number.isSafeInteger(obj.schemaVersion) ||
+        obj.schemaVersion <= 0
+    ) {
         return undefined;
     }
     if (obj.schemaVersion > META_SCHEMA_VERSION) {
--- a/src/common/inlineScript/cacheLayout.ts
+++ b/src/common/inlineScript/cacheLayout.ts
@@ -22,8 +22,12 @@ export const META_JSON_FILENAME = '.meta.json';
  * Schema version embedded in every {@link InlineScriptEnvMeta}.
  */
 export const META_SCHEMA_VERSION = 1 as const;
+export const SOURCE_METADATA_IDENTITY_HASH_HEX_LENGTH = 64;
+export const MAX_SOURCE_METADATA_IDENTITY_HASHES = 128;
 
 const MAX_META_JSON_BYTES = 1024 * 1024;
+const META_JSON_BACKUP_FILENAME_RE = /^\.meta\.json\.backup-[0-9a-f]{12}$/;
+const pendingMetaJsonWrites = new Map<string, Promise<void>>();
 
 /**
  * Validated on-disk schema for a cached inline-script environment's
@@ -38,11 +42,13 @@ export interface InlineScriptEnvMeta {
     readonly baseInterpreterVersion: string;
     /** Last successful use as a canonical UTC string produced by `Date.toISOString()`. */
     readonly lastUsedAt: string;
+    /** Bounded SHA-256 hashes of metadata identities proven for this cache entry. */
+    readonly sourceMetadataIdentityHashes?: readonly string[];
 }
 
 export type InlineScriptMetaReadResult =
     | { readonly kind: 'valid'; readonly metadata: InlineScriptEnvMeta }
-    | { readonly kind: 'missing' | 'invalid' | 'unavailable' };
+    | { readonly kind: 'missing' | 'invalid' | 'unsupported' | 'unavailable' };
 
 export type BaseInterpreterStatus = 'available' | 'missing' | 'unavailable';
 export type CacheEnvironmentInspection = 'expected' | 'stale' | 'uncertain';
@@ -119,8 +125,73 @@ export async function readMetaJson(envDir: Uri): Promise<InlineScriptEnvMeta | u
 
 /** Classify sidecar state; only `unavailable` denotes transient I/O. */
 export async function inspectMetaJson(envDir: Uri): Promise<InlineScriptMetaReadResult> {
-    const metaPath = getMetaJsonPath(envDir).fsPath;
+    return inspectMetaJsonFile(getMetaJsonPath(envDir).fsPath);
+}
 
+/**
+ * Restore the most recently-used compatible sidecar backup while the caller
+ * owns the cache-entry lock. General readers must use {@link inspectMetaJson}.
+ */
+export async function restoreMetaJsonBackupUnderLock(
+    envDir: Uri,
+    isCompatible: (metadata: InlineScriptEnvMeta) => boolean = () => true,
+): Promise<InlineScriptMetaReadResult> {
+    const finalPath = getMetaJsonPath(envDir).fsPath;
+    const initial = await inspectMetaJsonFile(finalPath);
+    if (initial.kind !== 'missing') {
+        return initial;
+    }
+
+    let entries: string[];
+    try {
+        entries = await fsapi.readdir(envDir.fsPath);
+    } catch (error) {
+        traceWarn(`inline-script meta: failed to scan backup sidecars in ${envDir.fsPath}:`, error);
+        return { kind: 'unavailable' };
+    }
+
+    const validBackups: Array<{ readonly path: string; readonly metadata: InlineScriptEnvMeta }> = [];
+    for (const entry of entries.filter((name) => META_JSON_BACKUP_FILENAME_RE.test(name))) {
+        const result = await inspectMetaJsonFile(path.join(envDir.fsPath, entry));
+        if (result.kind === 'valid' && isCompatible(result.metadata)) {
+            validBackups.push({ path: path.join(envDir.fsPath, entry), metadata: result.metadata });
+        } else if (result.kind === 'unavailable' || result.kind === 'missing') {
+            // A listed candidate changing or becoming unreadable is an
+            // uncertain scan; preserve the entry rather than rebuilding it.
+            return { kind: 'unavailable' };
+        }
+    }
+
+    if (validBackups.length === 0) {
+        return { kind: 'missing' };
+    }
+
+    // `lastUsedAt` is schema-validated canonical ISO text. Prefer the newest
+    // compatible backup; use the path as a stable tie-breaker.
+    validBackups.sort((a, b) => {
+        if (a.metadata.lastUsedAt !== b.metadata.lastUsedAt) {
+            return a.metadata.lastUsedAt < b.metadata.lastUsedAt ? 1 : -1;
+        }
+        return a.path < b.path ? -1 : a.path > b.path ? 1 : 0;
+    });
+    const selected = validBackups[0];
+
+    // Native rename may replace an existing destination on some platforms,
+    // so recheck under the caller's entry lock before restoring.
+    const current = await inspectMetaJsonFile(finalPath);
+    if (current.kind !== 'missing') {
+        return current;
+    }
+    try {
+        await fsapi.rename(selected.path, finalPath);
+    } catch (error) {
+        traceWarn(`inline-script meta: failed to restore backup ${selected.path}:`, error);
+        return { kind: 'unavailable' };
+    }
+    return { kind: 'valid', metadata: selected.metadata };
+}
+
+async function inspectMetaJsonFile(metaPath: string): Promise<InlineScriptMetaReadResult> {
     try {
         const stat = await fsapi.lstat(metaPath);
         if (!stat.isFile()) {
@@ -160,6 +231,10 @@ export async function inspectMetaJson(envDir: Uri): Promise<InlineScriptMetaRead
     }
 
     const validated = validateMeta(parsed);
+    if (validated === 'unsupported') {
+        traceWarn(`inline-script meta: unsupported schema in ${metaPath}`);
+        return { kind: 'unsupported' };
+    }
     if (!validated) {
         traceWarn(`inline-script meta: invalid shape in ${metaPath}`);
         return { kind: 'invalid' };
@@ -168,21 +243,95 @@ export async function inspectMetaJson(envDir: Uri): Promise<InlineScriptMetaRead
 }
 
 /**
- * Atomically write the `.meta.json` sidecar via temp-file + rename.
+ * Queue writes for a single sidecar in this process. Production callers also
+ * hold the cache-entry file lock, which serializes this operation across
+ * extension-host processes.
  */
-export async function writeMetaJson(envDir: Uri, meta: InlineScriptEnvMeta): Promise<void> {
-    await fsapi.ensureDir(envDir.fsPath);
+export function writeMetaJson(envDir: Uri, meta: InlineScriptEnvMeta): Promise<void> {
     const finalPath = getMetaJsonPath(envDir).fsPath;
+    const key = normalizePath(path.resolve(finalPath));
+    const previous = pendingMetaJsonWrites.get(key) ?? Promise.resolve();
+    const operation = previous.catch(() => undefined).then(() => writeMetaJsonOnce(envDir, meta, finalPath));
+    let queued: Promise<void>;
+    queued = operation.finally(() => {
+        if (pendingMetaJsonWrites.get(key) === queued) {
+            pendingMetaJsonWrites.delete(key);
+        }
+    });
+    pendingMetaJsonWrites.set(key, queued);
+    return queued;
+}
+
+async function writeMetaJsonOnce(envDir: Uri, meta: InlineScriptEnvMeta, finalPath: string): Promise<void> {
+    await fsapi.ensureDir(envDir.fsPath);
     const tmpSuffix = crypto.randomBytes(6).toString('hex');
     const tmpPath = `${finalPath}.tmp-${tmpSuffix}`;
+    const backupPath = `${finalPath}.backup-${tmpSuffix}`;
     const payload = JSON.stringify(meta, undefined, 2);
+    let hasBackup = false;
+    let finalKnownToExist = false;
+
     try {
         await fsapi.writeFile(tmpPath, payload, 'utf8');
-        await fsapi.rename(tmpPath, finalPath);
-    } catch (err) {
+        try {
+            await fsapi.rename(tmpPath, finalPath);
+            finalKnownToExist = true;
+            return;
+        } catch (err) {
+            const code = (err as NodeJS.ErrnoException | undefined)?.code;
+            if (!['EPERM', 'EEXIST', 'EBUSY'].includes(code ?? '')) {
+                throw err;
+            }
+        }
+
+        try {
+            await fsapi.rename(finalPath, backupPath);
+            hasBackup = true;
+        } catch (err) {
+            if (!isFileNotFoundError(err)) {
+                throw err;
+            }
+        }
+
+        try {
+            await fsapi.rename(tmpPath, finalPath);
+            finalKnownToExist = true;
+        } catch (replaceError) {
+            if (hasBackup) {
+                try {
+                    await fsapi.rename(backupPath, finalPath);
+                    finalKnownToExist = true;
+                } catch {
+                    // Keep the backup: it is the only known copy when
+                    // restoration cannot prove the final sidecar exists.
+                }
+            }
+            throw replaceError;
+        }
+    } finally {
         await fsapi.remove(tmpPath).catch(() => undefined);
-        throw err;
+        if (hasBackup && finalKnownToExist) {
+            await fsapi.remove(backupPath).catch(() => undefined);
+        }
     }
+}
+
+export function hashSourceMetadataIdentity(identity: string): string {
+    return crypto.createHash('sha256').update(identity, 'utf8').digest('hex');
+}
+
+export function mergeSourceMetadataIdentityHashes(
+    existing: readonly string[] | undefined,
+    current: string | undefined,
+): readonly string[] | undefined {
+    const ordered = [...(existing ?? [])];
+    if (current && !ordered.includes(current)) {
+        ordered.push(current);
+    }
+    if (ordered.length === 0) {
+        return undefined;
+    }
+    return Object.freeze(ordered.slice(-MAX_SOURCE_METADATA_IDENTITY_HASHES));
 }
 
 /**
@@ -290,11 +439,17 @@ function isNonEmptyTrimmedString(value: unknown): value is string {
     return typeof value === 'string' && value.length > 0 && value.trim() === value;
 }
 
-function validateMeta(value: unknown): InlineScriptEnvMeta | undefined {
+function validateMeta(value: unknown): InlineScriptEnvMeta | 'unsupported' | undefined {
     if (typeof value !== 'object' || value === null || Array.isArray(value)) {
         return undefined;
     }
     const obj = value as Record<string, unknown>;
+    if (typeof obj.schemaVersion !== 'number') {
+        return undefined;
+    }
+    if (obj.schemaVersion > META_SCHEMA_VERSION) {
+        return 'unsupported';
+    }
     if (obj.schemaVersion !== META_SCHEMA_VERSION) {
         return undefined;
     }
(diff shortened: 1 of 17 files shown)


heejaechang:

src/features/inlineScript/lazyDetector.ts
```diff
@@ -209,6 +243,18 @@ export class InlineScriptLazyDetector implements Disposable {
...
+        if (this.routingRegistry) {
+            const metadata = this.routingRegistry.getMetadata(e.document.uri);
+            if (
+                metadata &&
+                this.contentChangesMayAffectMetadata(
+                    e.contentChanges,
+                    metadata.sourceRange?.end ?? metadata.range.end,
+                )
+            ) {
+                this.clearRouteability(e.document.uri);
```
heejaechang, line 257: > **Issue** · Please address or respond

A metadata-affecting edit only clears routing when metadata has already been registered. If an open-file read is still in flight, this leaves its generation current, so it can publish the saved metadata after the document became dirty and route the dirty document. Advance the read generation for every metadata-affecting edit.

src/common/inlineScript/routingRegistry.ts
```diff
@@ -0,0 +1,195 @@
...
+        updater: (state: ScriptRoutingState) => ScriptRoutingState,
+        fireMetadataChange: boolean = false,
+    ): void {
+        const previous = this.states.get(scriptPath) ?? { metadataRevision: 0, validatedAssociation: false };
+        const previousRouteable = this.isRouteable(previous);
+        const next = updater(previous);
+
+        if (!next.metadata && !next.validatedAssociation) {
+            this.states.delete(scriptPath);
+        } else {
```
heejaechang, line 150: > **Issue** · Please address or respond

Deleting an empty state also discards its revision. A clear followed by restoring the same metadata identity can reuse an earlier revision, allowing an old validation task to satisfy its stale-result checks and re-enable routing. Preserve revisions independently or retain a tombstone.

src/common/inlineScript/metadata.ts
```diff
@@ -79,7 +88,8 @@ export function readInlineScriptMetadata(scriptText: string): InlineScriptMetada
     // "UTF-8 with BOM" on Windows have this; without stripping it the
```
heejaechang, line 88: > **Issue** · Please address or respond

`sourceRange` is computed after removing a BOM, but document change offsets include that BOM. Consequently, an edit at the end of a BOM-prefixed metadata block can be treated as body-only and leave stale routing active. Include the BOM displacement in `sourceRange` or translate the change offset.

src/managers/builtin/inlineScript/envManager.ts
```diff
@@ -1117,57 +1399,477 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
...
+    private cacheEntryProvesSourceMetadataIdentity(
+        sidecar: InlineScriptEnvMeta,
+        environment: PythonEnvironment,
+        metadataIdentity: string,
+        metadata: InlineScriptMetadata,
+    ): boolean {
+        return (
+            this.sidecarProvesSourceMetadataIdentity(sidecar, metadataIdentity) ||
+            this.isMetadataOnlyCacheEntryForMetadata(sidecar, environment, metadata)
+        );
```
heejaechang, line 1635: > **Info** · Optional note

The manager now combines cache lifecycle, sidecar recovery, persistence migration, provenance validation, and routing publication, increasing its change blast radius. Keep this from becoming permanent manager-level complexity by extracting the persisted-association validation state machine into a focused collaborator.

[verified]

heejaechang:

src/features/inlineScript/lazyDetector.ts
StellaHuang95: Addressed in cb4928c. Routing invalidation now advances the generation whenever an active read can no longer be trusted. If metadata is not registered yet but an open/save read is in flight, any edit fails closed; known metadata still preserves body-only edits and invalidates only metadata-affecting changes. `clearRouteability` also covers delete/rename/dirty-open races. Generation entries are retained only while a tracked read needs them and are removed when the final read completes. Added the in-flight open/edit regression.

src/common/inlineScript/routingRegistry.ts
StellaHuang95: Addressed in cb4928c. Metadata revisions now live in a separate per-path map, so deleting an empty routing state no longer resets the revision. `setMetadata` and `clearMetadata` always advance that monotonic revision, and newly created state inherits the current tombstone revision. Added direct registry coverage plus a manager regression where an old proof completes after clear + restore of the same metadata identity and remains unable to re-enable routing.

src/common/inlineScript/metadata.ts
StellaHuang95: Addressed in cb4928c. `sourceRange` now refers to the original source text: it adds the leading BOM displacement while retaining CRLF-aware offset translation. Normalized parser `range` remains unchanged. Updated parser coverage and added a detector regression for an edit at the end of a BOM-prefixed CRLF metadata block.

src/managers/builtin/inlineScript/envManager.ts
StellaHuang95: Agreed that the manager has accumulated several responsibilities. I am deferring collaborator extraction rather than mixing a large structural refactor into this routing correctness PR: the persistence/provenance state machine has extensive race coverage, and moving it now would materially enlarge an already large review without changing behavior. A focused extraction can follow once this routing contract is accepted, preserving the existing tests as the boundary specification.

src/managers/builtin/inlineScript/envManager.ts
```diff
@@ -1182,54 +1884,218 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
...
+            return { kind: 'invalid' };
+        }
+        if (!metadataBinding || typeof metadataBinding !== 'object' || Array.isArray(metadataBinding)) {
+            return { kind: 'invalid' };
+        }
+        const binding = metadataBinding as Record<string, unknown>;
+        if (binding.kind === 'pending') {
+            if (typeof binding.sourceIdentity === 'string' && binding.sourceIdentity.trim().length > 0) {
+                return {
+                    kind: 'valid',
```
heejaechang, line 2078: > **Warning** · Non-blocking recommendation

Only preserve integer schema versions newer than the current version. This currently treats `0`, negative, and fractional numeric values as future records, so malformed entries are retained indefinitely rather than repaired; classify those values as invalid.

heejaechang:

heejaechang (dismissed):
Approved via Review Center.

eleanorjboyd (changes requested):
Thanks for the thorough work here. I found three correctness issues that need addressing before approval:

1. `InlineScriptRoutingRegistry.setMetadata()` preserves `validatedAssociation` when the metadata routing identity changes. This leaves the old association routeable while async provenance validation for the newly saved metadata is pending, so callers can temporarily receive the previous inline environment. Please invalidate association validation synchronously whenever the metadata identity changes.

2. When a non-inline script override is cleared, `setEnvironment()` removes the override but continues resolving and publishing through the old manager/key. It also deletes the now-relevant inline last-known selection, so timeout-based API callers can receive `undefined` and no later manager event is guaranteed to correct it. Please re-resolve and publish through the manager that becomes effective after override removal.

3. The cache creation/reuse path rejects entries when the selected base interpreter version has changed or the environment no longer satisfies `requires-python`, but cached and rehydrated persisted-association validation does not apply those checks before restoring routeability. A same-path interpreter upgrade can therefore keep an association routeable that fresh cache validation would reject. Please apply the same base-version and compatibility invariants to persisted associations.

Test note: `triggerSavedMetadataChange()` both publishes through the registry and directly invokes the private handler. It would be valuable to exercise only the public event path so broken production wiring cannot be masked.

src/common/inlineScript/routingRegistry.ts
```diff
@@ -0,0 +1,206 @@
...
+                return {
+                    ...state,
+                    uri,
+                    metadata,
+                    metadataIdentity,
+                    metadataRevision,
+                };
+            },
+            true,
+        );
```
heejaechang, line 63: > **Issue** · Please address or respond

Changing `metadataIdentity` retains the prior `validatedAssociation`, so a script can remain routeable using proof for its previous metadata. Clear validation atomically when the identity changes, while retaining it for equivalent metadata refreshes.

src/features/envManagers.ts
```diff
@@ -540,6 +558,7 @@ export class PythonEnvironmentManagers implements EnvironmentManagers {
                     await manager.set(uris);
```
heejaechang, line 580: > **Issue** · Please address or respond

This clears an explicit non-inline override during an ordinary manager refresh, allowing a routeable script to switch back to the inline manager without a new user selection. Preserve the override until the user explicitly resets or replaces it.

src/managers/builtin/inlineScript/envManager.ts
```diff
@@ -1182,54 +1884,218 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
...
+                    record: {
+                        environmentPath,
+                        metadataBinding: { kind: 'pending', sourceIdentity: binding.sourceIdentity },
+                    },
```
heejaechang, line 2082: > **Warning** · Non-blocking recommendation

Only integer schema versions greater than the current version should be preserved as future records. Values such as `0`, negative, fractional, or non-finite numbers are malformed but currently remain preserved as future records, so they cannot be repaired.

heejaechang:

StellaHuang95:
Addressed the three correctness issues from review 4985022926 in aed8aee:

1. `InlineScriptRoutingRegistry.setMetadata()` now preserves association validation only when the normalized routing identity is unchanged. A changed identity synchronously emits the routeability-off transition before asynchronous provenance validation starts; equivalent metadata spelling/order remains routeable.

2. Clearing an explicit non-inline override now reserves the inline selection operation before awaiting the old manager, removes the override, re-resolves the effective manager, and publishes its environment instead of `undefined` through the old manager. The reserved operation is abandoned if any newer inline selection begins, so the handoff cannot overwrite newer user intent. Tests cover both the normal override-clear transition and the concurrent newer-selection race.

3. Warm and restart association validation now require a valid sidecar's recorded base release to match the resolved runtime and the current `requires-python` constraint to accept it before direct retrieval or routeability. When strict proof fails during selection, the known current identity is retained as non-routeable `pending` rather than degraded to broad `legacy` behavior. Existing future/unavailable sidecar preservation remains unchanged.

I also added a focused test that drives validation solely through `routingRegistry.setMetadata()` and the manager's constructor subscription. The existing private-handler tests remain for deterministic race control, but production event wiring is now independently covered.

Validation: compile-tests, bundle, lint, and full unit suite — 1,806 passing, 6 pending.

src/common/inlineScript/cacheLayout.ts
```diff
@@ -290,11 +439,17 @@ function isNonEmptyTrimmedString(value: unknown): value is string {
...
-function validateMeta(value: unknown): InlineScriptEnvMeta | undefined {
+function validateMeta(value: unknown): InlineScriptEnvMeta | 'unsupported' | undefined {
     if (typeof value !== 'object' || value === null || Array.isArray(value)) {
         return undefined;
     }
     const obj = value as Record<string, unknown>;
+    if (typeof obj.schemaVersion !== 'number') {
+        return undefined;
+    }
+    if (obj.schemaVersion > META_SCHEMA_VERSION) {
```
heejaechang, line 454: > **Issue** · Please address or respond

`schemaVersion: 1.5` is classified as unsupported rather than malformed, so cache inspection preserves it as uncertain and cannot rebuild the corrupted entry. Require an integer schema version before treating a version greater than the current schema as unsupported.

src/common/inlineScript/routingRegistry.ts
StellaHuang95: Addressed in aed8aee. `setMetadata()` now atomically clears association validation whenever the normalized metadata identity changes, while preserving it for equivalent normalized metadata. Focused registry tests verify the synchronous routeability-off transition and equivalent-identity retention.

src/features/envManagers.ts
StellaHuang95: Addressed in c1964df. Batch manager refresh/unset no longer clears `_inlineRoutingOverrides`; only the explicit single-scope reset/replacement path changes that user override. Added coverage showing a routeable script remains on its explicit non-inline manager through a batch refresh. The explicit override-clear handoff from aed8aee remains covered separately.

src/managers/builtin/inlineScript/envManager.ts
StellaHuang95: Addressed in c1964df. Persisted association schema versions are now accepted only when they are positive safe integers. Values greater than the current version remain preserved as future records; zero, negative, fractional, `NaN`, and infinite values are classified as malformed and can be repaired. Added parameterized repair coverage for each invalid numeric form.

src/common/inlineScript/cacheLayout.ts
StellaHuang95: Addressed in ba72a6e. Sidecar schema versions must now be positive safe integers before a value greater than the current version is classified as future/unsupported. Zero, negative, and fractional values are malformed and return `invalid`, allowing cache inspection to rebuild the corrupted entry. Added focused coverage for `0`, `-1`, and `1.5`; integer future version `99` remains `unsupported`.

src/features/envManagers.ts
```diff
@@ -475,7 +511,11 @@ export class PythonEnvironmentManagers implements EnvironmentManagers {
                     await setAllManagerSettings(settings);
```
heejaechang, line 511: > **Issue** · Please address or respond

When a batch refresh passes `environment === undefined`, this deletes the explicit non-inline override. That is not an explicit user reset, so a subsequently routeable inline association can regain precedence over the user's selection. Preserve the override during refreshes and clear it only for an explicit reset or replacement.

src/managers/builtin/inlineScript/envManager.ts
```diff
@@ -1274,6 +2162,37 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
...
+                .map((uri) => ({
+                    uri,
+                    scriptPath: normalizePath(uri.fsPath),
+                }))
+                .filter((script, index, all) => all.findIndex((candidate) => candidate.scriptPath === script.scriptPath) === index)
+                .filter(
+                    (script) =>
+                        this.fsPathToEnv.has(script.scriptPath) ||
+                        this.fsPathToPersistedAssociation.has(script.scriptPath),
+                );
```
heejaechang, line 2183: > **Warning** · Non-blocking recommendation

Delete/rename cleanup only considers associations already loaded into memory. If the event arrives while persisted-association initialization is still in flight, `changes` is empty and the durable association is loaded afterwards. Serialize cleanup with initialization or consult persisted state before returning.

heejaechang:

src/managers/builtin/inlineScript/envManager.ts
StellaHuang95: Addressed in ec1ddc4. The manager now starts and retains a dedicated persisted-association load promise at construction. Delete/rename cleanup waits for that load before filtering in-memory associations, while post-load route validation remains separate to avoid a selection-queue dependency cycle. A deterministic regression blocks the initial state read, fires deletion, and proves the durable association is removed after initialization completes.

src/features/envManagers.ts
StellaHuang95: Already addressed in c1964df; no additional code change is needed for this scenario. The `environment === undefined` batch path is the `else` branch and does not call `updateInlineRoutingOverride`; only the truthy explicit-selection branch updates or replaces an override. The focused `preserves an explicit non-inline override during a batch manager refresh` regression exercises `setEnvironments([script], undefined, false)` and passes at ec1ddc4.

heejaechang, line 511: > **Issue** · Please address or respond

`setEnvironments` updates the inline-routing override even when `environment` is `undefined`, which deletes an explicit non-inline override. A routeable script can then switch back to inline routing during a batch reset/refresh without an explicit user choice. Preserve the override for batch operations unless that operation is intended to explicitly replace or clear it.

heejaechang:
GitHub cannot anchor PR review comments to unchanged lines in the diff. Falling back to a general PR comment for [src/managers/builtin/inlineScript/envManager.ts:L227](https://github.com/microsoft/vscode-python-environments/blob/ec1ddc446270a374df4b561bc49bcbf2aba6c592/src/managers/builtin/inlineScript/envManager.ts#L227).

`routingRegistry` defaults to a new isolated registry, although routing requires the activation-scoped registry shared with the detector and `PythonEnvironmentManagers`. Make this dependency required (and provide it explicitly in tests) so a composition mistake cannot silently disconnect association validation from central routing.

heejaechang:

src/features/envManagers.ts
StellaHuang95: duplicate of last feedback

heejaechang:

heejaechang (approved):

bschnurr (approved):

eleanorjboyd (approved):

PR #1733, opened 2026-08-25
Harden uv Python compatibility for inline scripts

StellaHuang95:
## Context

Inline-script environment creation selects an installed Python when possible and falls back to installing a compatible Python with uv. The fallback has to use the same PEP 440 semantics before and after installation; otherwise setup can download Python successfully and then reject it.

This remains behind the undeclared, default-off inline-script feature flag.

## Why this change is needed

Two compatibility paths were inconsistent:

- A directly resolved Python was checked by the release-segment-only helper before the manager's strict PEP 440 check. This could reject a compatible final release, such as Python 3.15.0 for `!=3.15.0rc2`.
- A short exact requirement such as `==3.13` could be passed to uv as the broad selector `3.13`. uv may install a later 3.13 patch, while strict PEP 440 equality requires 3.13.0.

The default uv catalog can also omit older patch releases. A strict short-exact lookup therefore needs an all-versions catalog to find 3.13.0 after newer 3.13 releases exist.

## What changed

- Direct post-install resolution now keeps the existing Python 3, error, executable, and canonical-path guards while using strict PEP 440 compatibility.
- Short exact equality resolves through a concrete uv catalog result instead of forwarding a broad minor selector.
- Only the short-exact path requests `uv python list --all-versions`; existing catalog callers retain their previous command arguments.
- Fully specified exact versions, simple lower bounds, bounded ranges, exclusions, prerelease/dev requirements, no-requirement behavior, and quick-create prompt suppression remain unchanged.

## Behavior and compatibility

- No setting, command, menu, view, or status-bar contribution is added.
- The inline manager remains unregistered while the feature flag is off.
- Non-inline uv and package-management flows keep their existing behavior.
- Failures remain fail-closed: no incompatible cache environment is created.

## Reviewer guide

1. Review the strict direct-resolution guard in `inlineScript/envManager.ts`.
2. Review short-exact catalog selection and the `allVersions` option.
3. Review uv command-argument tests, then the manager compatibility regressions.

## Validation

- Targeted inline manager, uv installer, and interpreter suites: 311 passing
- `git diff --check`

### DIFF
--- a/src/managers/builtin/uvPythonInstaller.ts
+++ b/src/managers/builtin/uvPythonInstaller.ts
@@ -76,6 +76,10 @@ export interface UvPythonVersion {
     arch: string;
 }
 
+export interface GetAvailablePythonVersionsOptions {
+    readonly allVersions?: boolean;
+}
+
 /**
  * Checks if a command is available on the system.
  */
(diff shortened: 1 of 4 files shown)


heejaechang:

heejaechang:

heejaechang (approved):
Approved via Review Center.

src/managers/builtin/uvPythonInstaller.ts
```diff
@@ -276,12 +280,20 @@ export async function getUvPythonPath(version?: string): Promise<string | undefi
...
+    options?: GetAvailablePythonVersionsOptions,
+): Promise<UvPythonVersion[]> {
     return new Promise((resolve) => {
         const chunks: string[] = [];
-        const proc = spawnProcess('uv', ['python', 'list', '--output-format', 'json']);
+        const args = ['python', 'list'];
+        if (options?.allVersions) {
+            args.push('--all-versions');
+        }
+        args.push('--output-format', 'json');
```
edvilme, line 295: Why not add --output-format json to the args list?

StellaHuang95: what do you mean? it is added to the args list

PR #1734, opened 2026-08-25
Harden PEP 723 per-script routing against stale asynchronous selections

StellaHuang95:
## Summary

This PR narrows the earlier inline-script hardening to two focused, PEP 723-specific changes. Everything here is scoped to the manually enabled (default-off) inline-script routing feature.

### 1. Prevent stale asynchronous selections from mutating PEP 723 per-script routing overrides

When environment selections for a script resolve out of order, a superseded (older) operation could still install, clear, or overwrite that script's per-script inline routing override. The central manager now guards inline routing mutations with a per-script revision check (`commitInlineRoutingOperation`), gated strictly to:

- `inlineScriptRouting` being enabled, and
- a file `.py` scope recognized by `getInlineScriptRoutingKey`.

Only a current (non-superseded) operation may change a script's routing override. Each script commits on its **own per-file inline key**, so two routeable scripts under the same containing project no longer contend on a shared project revision — batch-selecting one non-inline environment for both now correctly routes **both** scripts.

The ordinary project/global active-selection and settings lanes commit independently, so a stale inline operation can no longer suppress a legitimate containing-project selection.

### 2. Preserve future-schema inline cache backups

`cacheLayout` now classifies a backup sidecar written by a newer cache schema as `unsupported` rather than rebuildable, so a newer-schema cache isn't rebuilt (and clobbered) by an older client.

## Explicitly deferred / out of scope

The generic cross-manager batch **settings write race is not addressed here**. A correct fix requires shared cross-manager settings serialization. This PR intentionally does **not** claim to make batch settings persistence revision-ordered, and it does **not** change ordinary (non-inline) project selection behavior. That work is deferred to a follow-up.

## Tests

- Adapted the stale-non-inline-vs-newer-inline regression: it asserts the script still routes to inline (its inline active selection is preserved) and no longer asserts the containing-project selection is suppressed.
- Kept the older-batch-inline-vs-newer-override regression.
- Added a same-project batch regression: two routeable scripts under one project, batch-selected to one non-inline environment, both resolve to the non-inline manager.
- Added a feature-off / no-routing-registry batch regression across distinct projects (manager.set, active-selection/events, and settings unchanged).
- Kept the `cacheLayout` future-schema backup tests.

Unit suite: 1838 passing, 6 pending, 0 failing.


heejaechang:

src/features/envManagers.ts
```diff
@@ -511,6 +525,9 @@ export class PythonEnvironmentManagers implements EnvironmentManagers {
                     await setAllManagerSettings(settings);
```
heejaechang, line 522: > **Issue** · Please address or respond

`setAllManagerSettings(settings)` runs before the new `commitPendingSelection` guard. An older batch can therefore persist its manager selection after a newer operation has won, even though its later routing mutation is rejected. Make the settings write subject to the same operation ordering guarantee.

src/managers/builtin/inlineScript/envManager.ts
```diff
@@ -1955,7 +2288,13 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
                     delete rawEntries[change.scriptPath];
```
heejaechang, line 1955: > **Issue** · Please address or respond

The generation check immediately before `await state.set(...)` cannot prevent disposal after the write has begun. The post-write check only prevents the in-memory update, so a disposed retry can still persist a stale association mutation. Serialize teardown with in-flight retry persistence, or otherwise make the storage write cancellation-aware.

heejaechang:

heejaechang:

heejaechang (dismissed):
Approved via Review Center.

src/features/envManagers.ts
StellaHuang95: Valid catch, thank you. This is deliberately deferred rather than fixed here: the correct fix is shared cross-manager settings serialization, not a per-selection revision filter. This revision removes the attempted generic selection-finalization machinery entirely (`commitPendingSelection`/`committedSelections` are gone and the batch settings write order is restored to `main`). The PR is now inline-routing-only — a PEP 723 per-script override guard (`commitInlineRoutingOperation`) that touches only the per-file inline routing key. It no longer claims to make batch settings persistence revision-ordered. The generic settings write race is tracked as a follow-up.

src/managers/builtin/inlineScript/envManager.ts
StellaHuang95: The retry / association-validation persistence subsystem was removed from this PR in this revision — `src/managers/builtin/inlineScript/envManager.ts` is reverted to `main`, so the generation-check-before-`await state.set(...)` path this comment flags no longer exists here. This PR is now inline-routing-only. The disposal-during-write concern will be revisited in the follow-up that reintroduces that subsystem.

src/common/inlineScript/cacheLayout.ts
```diff
@@ -163,7 +166,7 @@ export async function restoreMetaJsonBackupUnderLock(
     }
 
     if (validBackups.length === 0) {
-        return { kind: 'missing' };
+        return { kind: hasUnsupportedBackup ? 'unsupported' : 'missing' };
```
heejaechang, line 169: > **Warning** · Non-blocking recommendation

When compatible and future-schema backups coexist, `validBackups` takes precedence over `hasUnsupportedBackup`, restoring older metadata despite the newer-schema sidecar. Make an unsupported backup dominate restoration and add coverage for the mixed-backup case.

heejaechang:

heejaechang (approved):

eleanorjboyd (approved):

PR #1735, opened 2026-08-25
Harden inline script cache lifecycle coordination

StellaHuang95:
## Context

Inline-script environments live in a shared extension cache that may be accessed by more than one extension host. Clearing that cache is destructive and must coordinate with lock recovery, persistent-state clears, and inline association persistence.

The dedicated inline clear command remains hidden behind the undeclared, default-off inline-script flag and is not contributed to the Command Palette.

## What changed

This PR is scoped to three coordinated hardening changes:

1. **Atomic canonical lock retirement on release** (`src/common/lockfile.apis.ts`) — `release()` renames the owner marker to a transient `.release-*` marker and atomically renames the lock directory to a `.retired-*` sibling before cleaning it up out of band. Transient `EPERM`/`EBUSY`/`EACCES` retirement failures retry while ownership is retained; a terminal failure restores ownership and throws `ELOCKRELEASEFAILED`. Lock inspection recognizes the transient `.release-*` marker and tolerates the directory being retired between `lstat` and `readdir`.

2. **Serialized persistent-state clears** (`src/common/persistentState.ts`) — every clear request is serialized through a failure-isolated queue tail, each request's key options are preserved, and `clearPersistentState` gains `preserveWorkspaceKeys`/`preserveGlobalKeys` (`ClearPersistentStateOptions`).

3. **Generic Clear Cache preserves the inline association key** (`src/common/constants.ts`, `src/features/envCommands.ts`, `src/extension.ts`, and an import-only move of `INLINE_SCRIPT_ENVS_KEY` in the inline-script manager) — `python-envs.clearCache` now routes through `clearEnvironmentCachesCommand`, which preserves the inline association workspace-state key (built on the serialized-clear `preserveKeys` plumbing from change 2) while retaining existing behavior for all other state and managers.

## Behavior and compatibility

- No setting, command, menu, view, or status-bar contribution is added.
- Default-off users do not register the inline manager or dedicated command.
- Generic clear continues clearing existing non-inline managers and state. The only inline-specific difference is that dormant associations are no longer forgotten while their cache directories remain.
- Production lock API callers remain confined to the hidden inline manager.

## Reviewer guide

1. Review release-side lock retirement, ownership restoration, and inspection tolerance in `lockfile.apis.ts`.
2. Review persistent-clear serialization and preserved keys in `persistentState.ts`.
3. Review generic clear routing and inline-association preservation (`envCommands.ts`, `extension.ts`, `constants.ts`).

## Validation

- `npm run compile-tests`
- `npm run compile`
- `npm run lint`
- Unit suite: 1844 passing

> Note: This PR was reduced in scope from an earlier revision. Root-generation/admission changes, reclaim-side lock retirement, workspace-root settings protection, create-counting order, and the typed generic-inline unsupported error were reverted to the base; the net diff now covers only the three changes above.


heejaechang:

src/common/lockfile.apis.ts
```diff
@@ -80,16 +87,36 @@ export async function acquireFileLock(filePath: string, options: AcquireFileLock
...
+                            .then(
+                                () => true,
+                                () => false,
+                            );
+                        if (restored) {
+                            throw createLockError(
+                                'Failed to retire the lock directory; ownership was restored',
+                                'ELOCKRELEASEFAILED',
+                                lockPath,
+                                error,
```
heejaechang: > **Warning** · Non-blocking recommendation

If retiring the canonical directory fails and restoring the `.release-*` marker back to the owner marker also fails, this handle remains in `held` state but neither `release()` nor `retain()` can operate on its expected owner marker. Preserve a retryable transition state or retry restoration while ownership is retained, and cover the double-rename failure path.

src/managers/builtin/inlineScript/envManager.ts
```diff
@@ -2966,7 +3150,13 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
             return entryPath;
         } finally {
             if (lock) {
-                await lock.release();
+                const releaseError = await this.releaseCacheLockOrRetain(
+                    lock,
+                    'inline-script cache entry during cleanup',
+                );
+                if (releaseError) {
```
heejaechang: > **Warning** · Non-blocking recommendation

A generation-mismatch error can be replaced by a subsequent entry-lock release error when root replacement has already removed the old lock directory. Preserve the primary root-replacement failure and report the release failure secondarily so callers receive the actual reason cleanup was refused.

heejaechang:

heejaechang (dismissed):
Approved via Review Center.

src/common/lockfile.apis.ts
```diff
@@ -80,16 +86,36 @@ export async function acquireFileLock(filePath: string, options: AcquireFileLock
...
+                            );
+                        }
+                        throw error;
```
heejaechang: > **Issue** · Please address or respond

If both canonical-directory retirement and restoration of the owner marker fail, `state` remains `held` but the handle's owner marker no longer exists. A later `release()` or `retain()` on this handle cannot recover the live `.release-*` transition. Preserve a retryable transition state (or retry restoration while ownership is retained) and add coverage for both renames failing.

src/common/lockfile.apis.ts
```diff
@@ -269,12 +329,80 @@ function parseMarkerPid(entry: string, prefix: string): number | undefined {
...
+        pid > 0 &&
+        parseMarkerPid(generationMarker, generationPrefix) !== undefined
+        ? { pid, generationMarker }
+        : undefined;
+}
+
+function getRetiredLockPath(lockPath: string): string {
+    return `${lockPath}${FILE_LOCK_RETIRED_DIR_INFIX}${process.pid}-${crypto.randomBytes(16).toString('hex')}`;
+}
+
```
heejaechang, line 387: > **Warning** · Non-blocking recommendation

When both cleanup operations fail, this suppresses the errors and leaves a uniquely named `.retired-*` directory with no later cleanup path. Repeated sharing or permission failures can therefore accumulate retired lock artifacts indefinitely; add bounded, safe scavenging and fault-injection coverage.

heejaechang:

src/common/lockfile.apis.ts
StellaHuang95: Fixed in `3df425a`. `release()` now uses a resumable `releasing` state: once the owner marker is renamed to `.release-*`, if canonical-directory retirement fails **and** restoring the owner marker also fails, the handle stays `releasing` (rather than `held` with a missing owner marker) and throws a retryable `ELOCKRELEASEFAILED`. A later `release()` skips the owner→`.release-*` rename and resumes retirement, so the handle can always make progress while the process is alive; if restoration succeeds it returns to `held` as before. Added test *"release stays resumable when both retirement and ownership restoration fail"* covering the double-rename failure + same-handle resume.

src/common/lockfile.apis.ts
StellaHuang95: Deferring proactive scavenging here by design, to keep this PR's scope minimal. A leaked `.retired-*` directory only occurs when both `rmdir` **and** fs-extra `remove` (rimraf) fail on a freshly-created, uniquely-named directory — a degraded-filesystem state. The retired dir is inert: it's a sibling of the canonical `.lock`, is never mistaken for a live lock, and doesn't block acquisition (covered by *"an interrupted retired artifact neither blocks acquisition nor gets mistaken for a live canonical lock"*). Acquire-time scavenging would reverse that intentional test for a case that shouldn't occur in practice, so we're leaving retired artifacts inert.

src/managers/builtin/inlineScript/envManager.ts
StellaHuang95: No longer applicable — the PR scope was trimmed and the `releaseCacheLockOrRetain` cleanup path (and the rest of the inline-script envManager hardening) was reverted out. `envManager.ts` now only re-exports the `INLINE_SCRIPT_ENVS_KEY` constant, so the generation-mismatch vs. release-error precedence this refers to isn't part of the PR anymore.

src/common/persistentState.ts
```diff
@@ -32,14 +36,15 @@ class PersistentStateImpl implements PersistentState {
...
-        }
-        return this.clearing.promise;
+    async clear(keys?: string[], options?: { readonly preserveKeys?: readonly string[] }): Promise<void> {
+        const requestedKeys = keys ? [...keys] : undefined;
+        const preservedKeys = new Set(options?.preserveKeys ?? []);
+        const operation = this.clearQueue.then(async () => {
+            const keysToClear = (requestedKeys ?? this.momento.keys()).filter((key) => !preservedKeys.has(key));
+            await Promise.all(keysToClear.map((key) => this.momento.update(key, undefined)));
+        });
+        this.clearQueue = operation.catch(() => undefined);
```
heejaechang: > **Issue** · Please address or respond

`Promise.all` rejects on the first failed update, so `operation.catch()` can advance `clearQueue` while another `Memento.update` from the same clear is still pending. A later `set` or `clear` can then race that older deletion. Wait for all updates to settle before releasing the queue, then propagate the failure.

src/common/lockfile.apis.ts
```diff
@@ -77,19 +83,53 @@ export async function acquireFileLock(filePath: string, options: AcquireFileLock
...
+                    if (state === 'held') {
+                        try {
+                            await fsapi.rename(ownerMarker, releaseMarkerPath);
+                        } catch (error) {
+                            if (hasErrorCode(error, 'ENOENT')) {
+                                throw createLockError('Lock ownership was compromised', 'ECOMPROMISED', lockPath);
+                            }
+                            throw error;
+                        }
+                        state = 'releasing';
```
heejaechang: > **Warning** · Non-blocking recommendation

Two concurrent `release()` calls can both observe `held` before either awaited rename changes `state`; one then sees `ENOENT` and reports `ECOMPROMISED` during a valid release. Serialize the transition before the first filesystem await, with rollback if the rename fails.

heejaechang:

src/common/persistentState.ts
StellaHuang95: Fixed in `c3544ce`. `clear()` now awaits `Promise.allSettled(...)` so every `Memento.update` finishes before `clearQueue` advances, then re-throws the first rejection so the failure still propagates to the caller (`clearQueue = operation.catch(...)` continues to keep the chain from being poisoned). This closes the window where a later `set`/`clear` could race a still-pending deletion from a partially-failed clear. Added a regression test — `holds the queue until every deletion settles when a clear partially fails` — that gates a slow deletion behind a failing one and asserts a subsequently-queued write is applied strictly after the deletion settles.

src/common/lockfile.apis.ts
```diff
@@ -77,19 +83,53 @@ export async function acquireFileLock(filePath: string, options: AcquireFileLock
...
+                    }
+                    // state === 'releasing': the release marker exists; retire the canonical
```
heejaechang: > **Issue** · Please address or respond

Two concurrent `release()` calls can both observe `state === 'held'` before either owner-marker rename completes. The losing call then sees `ENOENT` and reports `ECOMPROMISED`, although the other call validly released the lock. Serialize concurrent releases through a shared in-flight promise and add a regression test.

src/common/persistentState.ts
```diff
@@ -1,28 +1,32 @@
...
+        await this.clearQueue;
         if (defaultValue === undefined) {
             return this.momento.get<T>(key);
         }
         return this.momento.get<T>(key, defaultValue);
     }
     async set<T>(key: string, value: T): Promise<void> {
-        await this.clearing.promise;
+        await this.clearQueue;
         await this.momento.update(key, value);
```
heejaechang, line 26: > **Issue** · Please address or respond

`set()` waits for the current clear tail but does not join the mutation queue. A clear requested after a gated write can delete the key before that older write finishes, allowing the write to restore data after the clear resolves. Queue writes on the same tail and cover a gated `set()` followed by `clear()` case.

src/common/lockfile.apis.ts
```diff
@@ -269,12 +343,80 @@ function parseMarkerPid(entry: string, prefix: string): number | undefined {
...
+        pid > 0 &&
+        parseMarkerPid(generationMarker, generationPrefix) !== undefined
+        ? { pid, generationMarker }
+        : undefined;
+}
+
+function getRetiredLockPath(lockPath: string): string {
+    return `${lockPath}${FILE_LOCK_RETIRED_DIR_INFIX}${process.pid}-${crypto.randomBytes(16).toString('hex')}`;
+}
+
```
heejaechang, line 387: > **Warning** · Non-blocking recommendation

If both cleanup mechanisms fail, the uniquely named retired directory is abandoned with no retry or scavenging path. Add bounded, ownership-safe cleanup of stale retired directories so repeated cleanup failures cannot accumulate artifacts indefinitely.

[verified]

heejaechang:

src/common/persistentState.ts
```diff
@@ -32,14 +36,21 @@ class PersistentStateImpl implements PersistentState {
             traceError('Error while updating state for key:', key);
         }
     }
```
heejaechang, line 34: > **Issue** · Please address or respond

`set()` waits for the current queue but does not join it, so a clear requested after a write begins can run while that write is still in flight. The write may then complete after the clear and restore state that was expected to be removed. Enqueue `set()` on the same failure-isolated tail as `clear()` and add a gated set-before-clear regression test.

[verified]

heejaechang:

heejaechang:

heejaechang (dismissed):

src/common/persistentState.ts
StellaHuang95: Fixed in `5e043b1`. `set()` now joins the same failure-isolated queue as `clear()` — it chains its update + read-back verification on `clearQueue` (`const operation = this.clearQueue.then(...); this.clearQueue = operation.catch(() => undefined); return operation;`), so a write becomes the new tail. A `clear()` requested after a gated `set()` now runs strictly after that write settles and can no longer be resurrected by a lagging update. Added regression test *"serializes a gated write before a later clear so the clear is not resurrected"*.

src/common/persistentState.ts
StellaHuang95: Fixed in `5e043b1`. `set()` now joins the same failure-isolated queue as `clear()` — it chains its update + read-back verification on `clearQueue` (`const operation = this.clearQueue.then(...); this.clearQueue = operation.catch(() => undefined); return operation;`), so a write becomes the new tail. A `clear()` requested after a gated `set()` now runs strictly after that write settles and can no longer be resurrected by a lagging update. Added regression test *"serializes a gated write before a later clear so the clear is not resurrected"*.

src/common/lockfile.apis.ts
StellaHuang95: Fixed in `5e043b1`. Concurrent `release()` calls now share one in-flight promise: the body was hoisted into `performRelease()` and `release()` memoizes it — `if (!releaseInFlight) { releaseInFlight = performRelease().finally(() => { releaseInFlight = undefined; }); } return releaseInFlight;`. Overlapping calls de-dupe onto the same operation, so the losing caller no longer sees `ENOENT`/`ECOMPROMISED` on a valid release; clearing the handle on settle preserves the resumable-retry semantics. Added regression test *"serializes concurrent release() calls through a shared in-flight promise"* (gates the owner→`.release-*` rename, starts a second `release()` while the first is in-flight, asserts both resolve, the transition runs once, lock ends `missing`).

src/common/lockfile.apis.ts
StellaHuang95: Fixed in `5e043b1`. Concurrent `release()` calls now share one in-flight promise: the body was hoisted into `performRelease()` and `release()` memoizes it — `if (!releaseInFlight) { releaseInFlight = performRelease().finally(() => { releaseInFlight = undefined; }); } return releaseInFlight;`. Overlapping calls de-dupe onto the same operation, so the losing caller no longer sees `ENOENT`/`ECOMPROMISED` on a valid release; clearing the handle on settle preserves the resumable-retry semantics. Added regression test *"serializes concurrent release() calls through a shared in-flight promise"* (gates the owner→`.release-*` rename, starts a second `release()` while the first is in-flight, asserts both resolve, the transition runs once, lock ends `missing`).

src/common/lockfile.apis.ts
StellaHuang95: Addressed by the resumable `release()` (`5e043b1`, originally `3df425a`): if retirement and owner-marker restoration both fail, the handle stays in a retryable `releasing` state and a later `release()` resumes retirement. Covered by *"release stays resumable when both retirement and ownership restoration fail"*.

src/common/lockfile.apis.ts
StellaHuang95: Deferring proactive scavenging by design to keep this PR's scope minimal. A leaked `.retired-*` dir only arises when both `rmdir` **and** fs-extra `remove` fail on a freshly-created unique directory. The retired dir is inert — a sibling of the canonical `.lock`, never mistaken for a live lock, and doesn't block acquisition (covered by an existing test). Acquire-time scavenging would reverse that intentional test for a case that shouldn't occur in practice. Happy to add bounded, ownership-safe scavenging as a follow-up if you'd prefer.

src/common/persistentState.ts
```diff
@@ -32,10 +37,11 @@ class PersistentStateImpl implements PersistentState {
...
-    async clear(keys?: string[]): Promise<void> {
+    async clear(keys?: string[], options?: { readonly preserveKeys?: readonly string[] }): Promise<void> {
         if (this.clearing.completed) {
             this.clearing = createDeferred<void>();
-            const _keys = keys ?? this.momento.keys();
+            const preservedKeys = new Set(options?.preserveKeys ?? []);
+            const _keys = (keys ?? this.momento.keys()).filter((key) => !preservedKeys.has(key));
             await Promise.all(_keys.map((key) => this.momento.update(key, undefined)));
             this.clearing.resolve();
         }
```
heejaechang, line 41: > **Issue** · Please address or respond

This is still a coalescing gate rather than a failure-isolated queue: concurrent callers await the first request and their own keys/preservation options are never applied. More importantly, if any `momento.update` rejects, `this.clearing.resolve()` is skipped, leaving the deferred pending so every later `clear()` waits forever. Queue each captured request independently and ensure a failed request cannot block later clears.

heejaechang:

**Result:** `could-not-verify`

**Summary:** No meaningful test execution was possible because the verification container could not start and local execution was not authorized. The PR adds 14 targeted unit tests covering lock retirement, persistent-state clearing, generic cache clearing, and inline persistence coordination. Their pass/fail status remains unverified.

**Test runs:** 2 not run

- :warning: **Not run** | <code>Lockfile and persistent-state unit suites</code> | <code>npm run compile-tests &amp;&amp; npm run unittest -- --grep &quot;(lockfile APIs|persistent state clearing)&quot;</code>
- :warning: **Not run** | <code>Environment cache command and inline manager unit suites</code> | <code>npm run compile-tests &amp;&amp; npm run unittest -- --grep &quot;(Clear Environment Caches Command Tests|InlineScriptEnvManager)&quot;</code>

<details><summary>:warning: <code>Lockfile and persistent-state unit suites</code> diagnostic output</summary>

<pre>Container verification could not start: no trusted sandbox image is configured for microsoft/vscode-python-environments; local execution was not authorized.</pre>
</details>

<details><summary>:warning: <code>Environment cache command and inline manager unit suites</code> diagnostic output</summary>

<pre>Container verification could not start: no trusted sandbox image is configured for microsoft/vscode-python-environments; local execution was not authorized.</pre>
</details>

src/common/persistentState.ts
StellaHuang95: You're right on both counts: the shared `PersistentState.clear()` is a coalescing gate, and a rejected `Memento.update` skips `this.clearing.resolve()`, leaving the deferred pending so every later `clear()` waits forever. Rather than reshape that shared implementation in this PR, I've taken the PR out of it entirely.

- `src/common/persistentState.ts` is now reverted to base — this PR has **no production diff** in that file. Its general coalescing/wedge defect is unchanged and explicitly out of scope for this inline-only PR (happy to track it separately).
- Inline PEP 723 associations moved to a new inline-owned, `Memento`-backed `InlineScriptAssociationStore` that owns only `INLINE_SCRIPT_ENVS_KEY`. It has its own failure-isolated FIFO queue: each caller awaits its **own** enqueued operation and receives that operation's success/failure, and a rejected op still advances the tail (`this.tail = run.then(ok, ignore)`) so a failed read/write/delete can't block later inline operations. The store never awaits the shared `clear()` deferred, so a wedged generic clear cannot stall inline work.
- Generic "Clear Cache" no longer needs any preserve option in shared code: at the command it reads the current workspace keys from the `Memento`, excludes `INLINE_SCRIPT_ENVS_KEY`, and calls the existing `PersistentState.clear(explicitKeys)` plus the existing global `clear()`. Generic clear and the inline store are key-disjoint (generic clears every key except the inline key; the store touches only the inline key), so neither can drop the other, and the dedicated inline delete is a queued direct update of that one key.

Net: this PR leaves the shared `clear()` behavior untouched, and inline association reliability no longer depends on it.

heejaechang:

**Result:** `could-not-verify`

**Summary:** No tests could execute because the verification container failed to start\: no trusted sandbox image is configured for this repository, and local execution was not authorized. The PR adds 19 targeted unit tests covering lock retirement, cache preservation, and inline association serialization. Because no meaningful test command ran, the verdict is \*\*could-not-verify\*\*.

**Test runs:** 1 not run

- :warning: **Not run** | <code>Targeted test discovery and dependency check</code> | <code>git status --short &amp;&amp; git branch --show-current &amp;&amp; git --no-pager diff --name-status upstream/main...HEAD &amp;&amp; node -e &quot;const p=require(&#x27;./package.json&#x27;); console.log(JSON.stringify({scripts:p.scripts,devDependencies:p.devDependencies&amp;&amp;{mocha:p.devDependencies.mocha,tsnode:p.devDependencies[&#x27;ts-node&#x27;]}},null,2))&quot; &amp;&amp; if [ -f .github/instructions/testing-workflow.instructions.md ]; then sed -n &#x27;1,220p&#x27; .github/instructions/testing-workflow.instructions.md; fi &amp;&amp; printf &#x27;\nNODE_MODULES=&#x27; &amp;&amp; if [ -d node_modules ]; then echo present; else echo missing; fi &amp;&amp; printf &#x27;PROFILE=%s\n&#x27; &quot;${AUTOMATION_SANDBOX_PROFILE:-}&quot; &amp;&amp; find src/test -type f \( -name &#x27;*lockfile*test.ts&#x27; -o -name &#x27;*envCommands*test.ts&#x27; -o -name &#x27;*associationStore*test.ts&#x27; -o -name &#x27;*inlineScript*test.ts&#x27; \) -print</code>

<details><summary>:warning: <code>Targeted test discovery and dependency check</code> diagnostic output</summary>

<pre>Container verification could not start and local execution was not authorized for this PR HEAD: No trusted sandbox image is configured for microsoft/vscode-python-environments.</pre>
</details>

heejaechang (approved):

PR #1736, opened 2026-08-25
Make standalone script template creation reliable

StellaHuang95:
## Context

The existing Script project template is presented as a PEP 723 standalone script, but its placeholder metadata is not valid TOML. The quick-create path also returns without creating a file, even though external callers provide a script project name and destination.

This change fixes the existing user-facing template and file-creation workflow. It does not create, select, or expose an inline-script environment.

## Why this change is needed

- The shipped metadata block cannot be parsed by the extension's own PEP 723 parser.
- Quick create skips the copy/register/open path.
- External quick callers provide a base project name such as `hello_world`, not necessarily a complete `.py` filename.
- Filename/path handling needs to reject traversal, reserved Windows device names, remote-workspace identity loss, and physical symlink escapes.
- Project-registration failure can otherwise leave a generated file or in-memory ghost project.
- Copilot instruction paths can be redirected outside the workspace through existing symlinks or junctions.

## What changed

- The template now contains valid `requires-python = ">=3.9"` metadata and an empty dependency list.
- Quick base names are normalized to `.py`; interactive creation continues requiring an explicit `.py` filename.
- Interactive and quick flows share validation for extension, characters, separators, traversal, containment, and Windows reserved names.
- Workspace matching preserves remote URI scheme and authority.
- Destination and Copilot-instruction paths are physically checked against the containing workspace before any side effect.
- Quick and interactive creation share the file copy, substitution, project registration, instruction, and open flow.
- Project registration is awaited. Insert-then-fail behavior removes the in-memory project and copied script while preserving the original error.
- Out-of-workspace destinations are rejected because they cannot be durably registered.
- Related documentation and Copilot instructions now describe the actual script behavior and metadata fields.

## Behavior and compatibility

- Package project creation is unchanged.
- Script creation still does not provision an environment.
- The hidden inline manager remains undeclared, default-off, and unregistered.
- Users already saw a PEP 723 block in generated scripts; this change makes that existing block valid.
- Opening the now-valid generated file may emit the existing anonymized telemetry-only detection event. It does not create or select an environment.

## Reviewer guide

1. Review shared filename normalization and validation.
2. Review workspace URI matching and physical containment preflight.
3. Review side-effect ordering and registration rollback.
4. Review the actual template parser test and quick-create tests.

## Validation

- Targeted metadata parser and script creator suites: 71 passing
- `git diff --check`


heejaechang:

src/features/creators/newScriptProject.ts
```diff
@@ -67,64 +145,152 @@ export class NewScriptProject implements PythonProjectCreator {
...
+            ]);
+        }
 
-            await showTextDocument(createdScript.uri);
+        await showTextDocument(createdScript.uri);
 
-            return createdScript;
-        }
-        return undefined;
+        return createdScript;
```
heejaechang, line 224: > **Warning** · Non-blocking recommendation

If `replaceInFilesAndNames` rejects after the template copy succeeds, the partially created script remains and prevents a clean retry. Remove the copied destination when substitution fails, and add a rejecting-stub regression test.

src/features/creators/newScriptProject.ts
```diff
@@ -67,64 +145,152 @@ export class NewScriptProject implements PythonProjectCreator {
...
+        if (createCopilotInstructions) {
+            const packageInstructionsPath = path.join(
+                NEW_PROJECT_TEMPLATES_FOLDER,
+                'copilot-instructions-text',
+                'script-copilot-instructions.md',
+            );
+            await manageCopilotInstructionsFile(destRoot, packageInstructionsPath, [
+                { searchValue: '<script_name>', replaceValue: scriptFileName },
+            ]);
+        }
```
heejaechang, line 221: > **Warning** · Non-blocking recommendation

The physical containment check and the later write are separate operations, so a symlink or junction could be swapped after validation and before the copy. Revalidate at the write boundary or use an operation anchored to a safely opened directory to preserve the intended containment guarantee.

heejaechang:
GitHub cannot anchor PR review comments to unchanged lines in the diff. Falling back to a general PR comment for [src/features/creators/newScriptProject.ts:L336](https://github.com/microsoft/vscode-python-environments/blob/08ca7ee87926fe993955b48d7f1bc65e686279f1/src/features/creators/newScriptProject.ts#L336).

After project registration succeeds, failures from Copilot-instruction management or `showTextDocument` reject `create` while retaining the generated file and project. Define the registration commit boundary explicitly: either roll back these resources on such failures or make post-registration editor opening best-effort, with focused failure coverage.

src/test/features/creators/newScriptProject.unit.test.ts
```diff
@@ -20,25 +32,693 @@ const TEMPLATE_PATH = path.join(
...
+            uri: Uri.file(tmpDir),
+        };
+        getWorkspaceFolderStub = sinon.stub(workspaceApis, 'getWorkspaceFolder').returns(workspaceFolder);
+        getWorkspaceFoldersStub = sinon.stub(workspaceApis, 'getWorkspaceFolders').returns([workspaceFolder]);
     });
 
     teardown(async () => {
+        sinon.restore();
         await fs.remove(tmpDir);
     });
```
heejaechang, line 64: > **Info** · Optional note

`stubSuccessfulFileCreation` relies on the order of `pathExists` calls, so harmless internal reordering will break unrelated tests. Use a path-aware fake that returns based on the requested path instead.

heejaechang:

heejaechang (dismissed):
Approved via Review Center.

src/features/creators/newScriptProject.ts
```diff
@@ -67,64 +86,121 @@ export class NewScriptProject implements PythonProjectCreator {
...
+        // Check if the destination file already exists
+        if (await fs.pathExists(scriptDestination)) {
+            window.showErrorMessage(
+                l10n.t(
+                    'A script file by that name already exists, aborting creation. Please retry with a unique script name given your workspace.',
+                ),
+            );
+            return undefined;
+        }
+        await fs.copy(newScriptTemplateFile, scriptDestination);
```
heejaechang: > **Issue** · Please address or respond

The workspace containment check is lexical only. A destination root that is syntactically inside the workspace but traverses an existing symlink or junction can cause `fs.copy` to write the script outside the workspace. Resolve and validate physical paths before writing.

heejaechang:

src/features/creators/newScriptProject.ts
StellaHuang95: Good catch — restored physical-path validation before the write. The destination root and workspace root are now resolved with fs.realpath and checked with isSameOrParentPath before fs.copy, so a destination that lexically sits inside the workspace but traverses a symlink/junction to an outside target is rejected. Added a unit test for the symlink-escape rejection. Fixed in 5b24cbe.

src/features/creators/newScriptProject.ts
```diff
@@ -67,64 +86,138 @@ export class NewScriptProject implements PythonProjectCreator {
...
+        if (createCopilotInstructions) {
+            const packageInstructionsPath = path.join(
+                NEW_PROJECT_TEMPLATES_FOLDER,
+                'copilot-instructions-text',
+                'script-copilot-instructions.md',
+            );
+            await manageCopilotInstructionsFile(destRoot, packageInstructionsPath, [
+                { searchValue: '<script_name>', replaceValue: scriptFileName },
+            ]);
+        }
```
heejaechang, line 221: > **Issue** · Please address or respond

`realpath` validates only the current filesystem state. An ancestor can be replaced with a symlink or junction between this preflight and `fs.copy`, allowing the write to escape the workspace. Use a write mechanism anchored to the validated directory, or otherwise prevent a post-validation symlink swap from being followed.

src/features/creators/newScriptProject.ts
```diff
@@ -67,64 +86,138 @@ export class NewScriptProject implements PythonProjectCreator {
...
 
-            await showTextDocument(createdScript.uri);
+        await showTextDocument(createdScript.uri);
 
-            return createdScript;
-        }
-        return undefined;
+        return createdScript;
```
heejaechang, line 224: > **Issue** · Please address or respond

If `replaceInFilesAndNames` rejects after `fs.copy` succeeds, execution never reaches the registration rollback and the copied script remains, blocking a clean retry. Put copying and substitution in the same cleanup boundary and remove the destination when substitution fails.

src/test/features/creators/newScriptProject.unit.test.ts
```diff
@@ -20,25 +32,530 @@ const TEMPLATE_PATH = path.join(
...
```
heejaechang, line 64: > **Issue** · Please address or respond

This fixture assigns `pathExists` results by invocation order. A harmless reordering of existence probes can make tests fail or simulate the wrong filesystem state. Make the stub inspect the requested path and return template/destination results accordingly.

heejaechang:

src/features/creators/newScriptProject.ts
StellaHuang95: Fixed in 1bf156c. Copy, name substitution, and project registration now share one try/catch cleanup boundary — if substitution (or registration) fails after `fs.copy`, the partially created script is removed with `fs.remove` before the error propagates, so a retry starts from a clean state. Added a regression test that rejects `replaceInFilesAndNames` after the copy and asserts the destination is deleted (ENOENT) and the project is never registered.

src/test/features/creators/newScriptProject.unit.test.ts
StellaHuang95: Fixed in 1bf156c. `stubSuccessfulFileCreation` no longer depends on probe call order — the `pathExists` fake now resolves by the requested path (true only for the template file, false for the destination), so reordering the existence checks can no longer simulate the wrong filesystem state.

src/features/creators/newScriptProject.ts
StellaHuang95: You're right that the `realpath` preflight is check-then-use and can't, on its own, stop an ancestor from being swapped for a symlink/junction between validation and `fs.copy`. Before adding more machinery I wanted to surface the tradeoff:

- **Threat model:** the destination is a folder inside the user's own open workspace in a local, single-user desktop extension. Anyone able to swap a directory for a symlink in that window already has write access to the workspace and can compromise it more directly (e.g. edit code that runs on the next build/test run). No privilege boundary is being crossed here.
- **Feasibility:** closing the race for real needs anchored I/O (an `openat`/`O_NOFOLLOW`-style operation against a validated directory handle). Node's `fs`/`fs-extra` don't expose that portably, and `fs.copy` follows symlinks by design. Re-checking `realpath` right before the copy only shrinks the window, it doesn't close it.

So I've kept the `realpath` check as a best-effort guard rather than a hard guarantee for this reliability-focused PR. If you feel the residual race warrants a true guarantee, I'm happy to take it on as a separate hardening task — it would likely need a small anchored-write helper. Let me know which you'd prefer.

heejaechang:

src/features/creators/newScriptProject.ts
```diff
@@ -1,33 +1,68 @@
...
+        return l10n.t('Script name must be a file name without path separators or traversal.');
+    }
+    if (!value.endsWith('.py')) {
+        return l10n.t('Script name must end with ".py".');
+    }
+    const baseName = value.replace(/\.py$/, '');
+    const deviceBaseName = baseName.split('.')[0];
+    if (isWindows() && /^(con|prn|aux|nul|com[1-9]|lpt[1-9])$/i.test(deviceBaseName)) {
+        return l10n.t('Script name uses a reserved Windows device name.');
+    }
```
edvilme: I would put this as a function in pathUtils

src/features/creators/newScriptProject.ts
```diff
@@ -1,33 +1,68 @@
...
+    // following PyPI (PEP 508) rules for package names
+    if (!/^([a-z_]|[a-z0-9_][a-z0-9._-]*[a-z0-9_])$/i.test(baseName)) {
+        return l10n.t(
+            'Invalid script name. Use only letters, numbers, underscores, hyphens, or periods. Must start and end with a letter or number.',
+        );
+    }
```
edvilme, line 30: I think we already have a function to validate valid PEP508 package names
Though I might be wrong

src/features/creators/newScriptProject.ts
```diff
@@ -67,64 +86,143 @@ export class NewScriptProject implements PythonProjectCreator {
...
+            return undefined;
+        }
 
-            // 1. Copy template folder
-            const newScriptTemplateFile = path.join(NEW_PROJECT_TEMPLATES_FOLDER, 'new723ScriptTemplate', 'script.py');
-            if (!(await fs.pathExists(newScriptTemplateFile))) {
-                window.showErrorMessage(l10n.t('Template file does not exist, aborting creation.'));
-                traceError(`Template file not found at: ${newScriptTemplateFile}`);
+        // 1. Copy template file
+        const newScriptTemplateFile = path.join(NEW_PROJECT_TEMPLATES_FOLDER, 'new723ScriptTemplate', 'script.py');
```
edvilme: Maybe we could call it `"newInlineScriptTemplate"`?

src/features/creators/newScriptProject.ts
StellaHuang95: Moved the Windows reserved device-name check into the shared path utilities as `isWindowsReservedDeviceName` and reused it from the script filename validator, so `newScriptProject` no longer needs its own `isWindows` import. Added focused unit coverage for the helper in `pathUtils.unit.test.ts`. Fixed in 7b39699.

src/features/creators/newScriptProject.ts
StellaHuang95: Renamed the internal template folder to `newInlineScriptTemplate` and updated all creator and test references (no stale `new723` references remain). The template content and the user-facing creator/class/command are unchanged. Fixed in 7b39699.

StellaHuang95: done

src/features/creators/newScriptProject.ts
StellaHuang95: done

src/features/creators/newScriptProject.ts
StellaHuang95: Didn't find so keeping it here

heejaechang (approved):

PR #1737, opened 2026-08-25
Fix script terminal commands returning errors after dispatch

StellaHuang95:
## Context

The existing normal and dedicated script-terminal command handlers dispatch the script successfully, then fall through to the invalid-context error because the successful branches do not return.

This affects any caller using these existing commands; it is not specific to PEP 723.

## Why this change is needed

A script can start running while the command promise rejects, causing callers to report failure after successful dispatch. The false error also makes command completion inconsistent with the underlying terminal operation.

## What changed

- Return after successful normal terminal dispatch.
- Return after successful dedicated terminal dispatch.
- Preserve invalid-context errors for non-URI inputs and URI inputs without a project or environment.

## Behavior and compatibility

- No command, menu, setting, view, or status-bar contribution is added.
- Environment resolution and fallback are unchanged.
- Terminal selection/reuse, cwd, arguments, visibility, and activation behavior are unchanged.
- The only success-path difference is that the command now resolves instead of throwing after dispatch.

## Reviewer guide

1. Review the two production return statements.
2. Review success coverage for resolved and fallback environments.
3. Review invalid and missing-context regressions.

## Validation

- Targeted command and terminal manager suites: 33 passing
- `git diff --check`

### DIFF
--- a/src/features/envCommands.ts
+++ b/src/features/envCommands.ts
@@ -831,6 +831,7 @@ export async function runInTerminalCommand(
                 args: [item.fsPath],
                 show: true,
             });
+            return;
         }
     }
     throw new Error(`Invalid context for run-in-terminal: ${item}`);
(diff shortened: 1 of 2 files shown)


heejaechang:

heejaechang:

heejaechang (approved):
Approved via Review Center.

PR #1772, opened 2026-09-09
Fix five PEP 723 inline-script defects: cancellation, rename, shared builds, terminal scope, and package drift

StellaHuang95:
## Summary

Five fixes to the PEP 723 inline-script feature, found while working through reported issues against the testbed. Each is independently reviewable and has unit coverage that was checked to **fail when the corresponding guard is removed**, not merely to pass.

The feature remains behind the undeclared `python-envs.inlineScripts.enabled` flag (default `false`), so nothing here is user-visible until that is enabled.

## Commits and the issue each addresses

| Commit | Issue addressed |
|---|---|
| `0a21a2c1` | Cancelling setup retained the cache lock with no cleanup, so the retry failed with `Lock was retained after an interrupted operation` |
| `887548d8` | Renaming a script dropped its environment association even though nothing about the environment changed |
| `eaa71423` | Cancelling a build shared by two scripts reported cancellation to only one of them |
| `22628e1f` | Selecting a script environment could leak into general workspace terminals |
| `669cb8ed` | Editing packages on a shared environment silently broke other scripts |

Two earlier commits on this branch (`9acf5a21`, `cd7d214a`) predate this work and are unrelated to the above.

---

### `0a21a2c1` — Discard cancelled environments instead of quarantining them

Cancelling package installation retained the cache-entry lock and made no cleanup attempt. The cancellation surfaced as a generic *"Failed to set up the environment for this script"*, and the next attempt failed with `ELOCKRETAINED`. Recovery required clearing the entire cache.

Cancellation now cleans up automatically. `discardCacheEntry` deletes `.meta.json` and any `.meta.json.backup-*` **first**, then removes the directory with bounded retry.

**The key invariant for reviewers:** the sidecar is the correctness guarantee, not the directory. `inspectCacheEntry` treats a missing sidecar as `stale`, and `writeMetaJson` is the only writer of that file — four call sites, all under the cache-entry lock (see the comment above `withCacheEntryLock`). A surviving installer writes into `site-packages` and cannot recreate a sidecar. So an entry whose directory survives deletion is inert, and gets rebuilt or TTL-swept rather than reused.

Retry exists because a just-stopped installer can hold file handles briefly, most visibly on Windows.

Also fixes two bulk-setup bugs: `setUpInlineScriptEnvironmentsInWorkspace` only counted successes and never surfaced outcomes, so a mid-run cancellation was invisible and the next install started immediately. It now stops the run on cancellation and reports failures distinctly.

### `887548d8` — Follow renames instead of dropping associations

A cache entry is keyed by normalized dependencies and base interpreter — **the script path is not an input** — so a rename cannot invalidate it. Neither the metadata block nor the environment changes.

This also left two subsystems disagreeing. `PythonProjectManagerImpl` already follows renames via `updatePythonProjectSettingPath`, which rewrites the `pythonProjects` entry and preserves its `_inlineScriptRegistration` marker. Clearing the association therefore produced a managed inline-script project entry pointing at a file with no environment behind it.

The record is transferred in one persistence transaction and then **re-validated rather than trusted**: its metadata binding is content-derived, so if the file at the new path no longer matches, ordinary validation clears it and the CodeLens returns.

Guards: destination must still be a routable local `.py`; renaming onto an already-associated script replaces it; deletes are unchanged.

**Known gap:** directory renames are not covered — VS Code reports one event per folder rather than one per contained file. Moving an individual *file* already works, since VS Code reports a move as a rename.

### `eaa71423` — Cancellation reaches every script sharing a build

Scripts whose dependencies normalize to the same list resolve to the same cache key, so a second request joins the in-flight build via `pendingCreations` instead of starting its own. Cancelling recorded an outcome only against the initiating URI. The joined script fell through to a generic failure, and in a bulk run its outcome was not a cancellation — so the run kept installing.

The failure is now recorded on the shared `PendingCreationContext` and translated per caller after awaiting the shared promise.

**Note for reviewers:** same-file coalescing (`pendingSetups`) and different-file shared builds (`pendingCreations`) are separate mechanisms. Only the second was broken. Bulk setup is sequential, so selecting both files in one bulk run does not reproduce it — the second request must overlap the first build.

### `22628e1f` — Shell startup variables resolve at folder scope

Shell-startup activation keeps a single activation command per workspace folder (for example `VSCODE_PYTHON_PWSH_ACTIVATE`), injected into every terminal opened there. `handleEnvironmentChange` wrote whichever environment the event carried into that slot after collapsing the event's URI to its containing folder — so a per-file selection became the folder default.

The handler now resolves the folder's own environment via `getEnvironment(workspaceFolder.uri)`. This matches what `initializeInternal` already did, so the two paths no longer disagree, and it fixes the whole class rather than special-casing the inline-script manager: any file-scoped environment is excluded. A folder already polluted by an earlier session self-repairs on the next environment change.

Only reachable with `python-envs.terminal.autoActivationType: "shellStartup"`; the default is `"command"`. Adds the first unit coverage for this manager.

### `669cb8ed` — Invalidate an environment when its packages are edited

Cache entries are shared by design. Editing packages there through the package UI silently affected every bound script: nothing validated installed versions, so the entry stayed valid, the CodeLens stayed hidden, and a script the user never opened would run the wrong version while its own header still declared the original pin. Re-running setup reused the modified entry rather than repairing it.

A package change on an owned entry now records `manuallyModified` in the sidecar and un-routes every associated script, so the CodeLens returns for each. `inspectCacheEntry` treats a marked entry as stale, so the next setup rebuilds from declared metadata.

**Please look closely at the in-flight guard.** Setup installs through `managePackages`, which fires this same event. Without the guard, every setup would mark the environment it had just built and rebuild endlessly. The guard relies on VS Code's `EventEmitter` being synchronous, so a build still registered in `pendingCreations` is recognised before any `await`; the check is repeated once the lock is held.

**Known gaps, deliberately not addressed:** sharing is still not surfaced before an edit; a deliberate ad-hoc install is discarded by the next setup without explanation; package changes made outside VS Code fire no event and remain undetected.

---

### DIFF
--- a/src/features/inlineScript/setupEnvironment.ts
+++ b/src/features/inlineScript/setupEnvironment.ts
@@ -124,10 +124,14 @@ function setupInlineScriptEnvironmentHandler(
     };
 }
 
-function notifyInlineScriptSetupOutcome(uri: Uri, routing: InlineScriptRoutingRegistry): void {
-    const outcome = routing.takeSetupOutcome(uri);
+export function notifyInlineScriptSetupOutcome(uri: Uri, routing: InlineScriptRoutingRegistry): void {
+    const outcome = routing.getSetupOutcome(uri);
     if (outcome?.kind === 'skipped') {
         // Env built but intentionally not associated (metadata changed mid-setup); stay silent.
+        return;
+    }
+    if (outcome?.kind === 'cancelled') {
+        showInformationMessage(l10n.t('Environment setup was canceled.'));
         return;
     }
     if (outcome?.kind === 'failed') {
@@ -223,16 +227,57 @@ export async function setUpInlineScriptEnvironmentsInWorkspace(
         return;
     }
     let succeeded = 0;
+    let attempted = 0;
+    let failed = 0;
+    let cancelled = false;
     for (const pick of picks) {
+        attempted += 1;
         try {
             if (await setUpInlineScriptEnvironment(pick.uri, em, routing)) {
                 succeeded += 1;
+                continue;
+            }
+            const outcome = routing.getSetupOutcome(pick.uri);
+            if (outcome?.kind === 'cancelled') {
+                // Cancelling one script's installer stops the whole run rather than immediately
+                // starting the next script's install.
+                cancelled = true;
+                break;
+            }
+            if (outcome?.kind !== 'skipped') {
+                failed += 1;
             }
         } catch (error) {
+            failed += 1;
             traceError(`Failed to set up the inline-script environment for ${pick.uri.fsPath}:`, error);
         }
     }
-    traceInfo(`Inline-script bulk setup: created or reused ${succeeded} of ${picks.length} environment(s).`);
+    traceInfo(
+        `Inline-script bulk setup: created or reused ${succeeded} of ${picks.length} environment(s)` +
+            `${cancelled ? ' (canceled)' : ''}.`,
+    );
+    if (cancelled) {
+        showWarningMessage(
+            l10n.t(
+                'Environment setup was canceled. Set up {0} of {1} selected inline script environment(s); the remaining {2} were not started.',
+                succeeded,
+                picks.length,
+                picks.length - attempted,
+            ),
+        );
+        return;
+    }
+    if (failed > 0) {
+        showWarningMessage(
+            l10n.t(
+                'Set up {0} of {1} selected inline script environment(s). {2} failed — see the Python Environments output for details.',
+                succeeded,
+                picks.length,
+                failed,
+            ),
+        );
+        return;
+    }
     showInformationMessage(l10n.t('Set up {0} of {1} selected inline script environment(s).', succeeded, picks.length));
 }
 
--- a/src/features/inlineScript/setupEnvironment.ts
+++ b/src/features/inlineScript/setupEnvironment.ts
@@ -75,6 +75,9 @@ export async function setUpInlineScriptEnvironment(
         return undefined;
     }
     await em.setEnvironment(scriptUri, environment);
+    // Outcomes are read non-consumingly, so a success must retire the previous attempt's outcome
+    // rather than relying on the next `create` to clear it on entry.
+    routing.clearSetupOutcome(scriptUri);
     return environment;
 }
 
--- a/src/managers/builtin/inlineScript/envManager.ts
+++ b/src/managers/builtin/inlineScript/envManager.ts
@@ -2282,19 +2282,6 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
     }
 
     /**
-     * Follow a rename instead of dropping the association.
-     *
-     * A cache entry is keyed by the script's dependencies and base interpreter, never by its path,
-     * so renaming a script does not invalidate its environment. `PythonProjectManagerImpl` already
-     * rewrites the matching `python-envs.pythonProjects` entry to the new path, so clearing the
-     * association here would leave a managed inline-script project entry with no environment behind
-     * it.
-     *
-     * The moved record is re-validated afterwards rather than trusted: its metadata binding is
-     * content-derived, so if the file at the new path no longer matches, ordinary validation clears
-     * the association and the setup CodeLens returns.
-     */
-    /**
      * Editing packages outside setup silently affects every script sharing the entry, so mark the
      * entry non-reusable and un-route each of them; the next setup rebuilds from declared metadata.
      */
@@ -2367,6 +2354,19 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
         }
     }
 
+    /**
+     * Follow a rename instead of dropping the association.
+     *
+     * A cache entry is keyed by the script's dependencies and base interpreter, never by its path,
+     * so renaming a script does not invalidate its environment. `PythonProjectManagerImpl` already
+     * rewrites the matching `python-envs.pythonProjects` entry to the new path, so clearing the
+     * association here would leave a managed inline-script project entry with no environment behind
+     * it.
+     *
+     * The moved record is re-validated afterwards rather than trusted: its metadata binding is
+     * content-derived, so if the file at the new path no longer matches, ordinary validation clears
+     * the association and the setup CodeLens returns.
+     */
     private async handleRenamedScripts(
         files: readonly { readonly oldUri: Uri; readonly newUri: Uri }[],
     ): Promise<void> {
--- a/src/managers/builtin/inlineScript/envManager.ts
+++ b/src/managers/builtin/inlineScript/envManager.ts
@@ -62,6 +62,7 @@ import { extractLowerBoundVersion, pickCompatibleInterpreter } from '../../../co
 import { InlineScriptMetadata, readInlineScriptMetadataFromFile } from '../../../common/inlineScript/metadata';
 import {
     getInlineScriptMetadataRoutingIdentity,
+    getInlineScriptRoutingKey,
     InlineScriptMetadataChangeEvent,
     InlineScriptRoutingRegistry,
 } from '../../../common/inlineScript/routingRegistry';
@@ -260,9 +261,9 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
                 });
             }),
             onDidRenameFiles((event) => {
-                void this.clearAssociationsForScripts(event.files.map((file) => file.oldUri)).catch((error) => {
+                void this.handleRenamedScripts(event.files).catch((error) => {
                     this.log.warn(
-                        `Failed to clear inline-script associations for renamed files: ${getErrorMessage(error)}`,
+                        `Failed to update inline-script associations for renamed files: ${getErrorMessage(error)}`,
                     );
                 });
             }),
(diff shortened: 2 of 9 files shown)


rchiodo:

src/test/features/inlineScript/setupEnvironment.unit.test.ts
```diff
@@ -209,4 +210,93 @@ suite('setUpInlineScriptEnvironmentsInWorkspace', () => {
...
+suite('notifyInlineScriptSetupOutcome', () => {
+    const scriptUri = Uri.file('/workspace/script.py');
+    let routing: InlineScriptRoutingRegistry;
+
+    setup(() => {
+        routing = new InlineScriptRoutingRegistry();
+    });
+
+    teardown(() => {
+        routing.dispose();
```
rchiodo, line 285: > **Issue** · Please address or respond

These `assert.match` and `assert.doesNotMatch` checks violate `docs/pylancewiki/review/tests-no-partial-asserts.md`, which requires exact, complete assertions for static user-facing strings. Replace them with `assert.strictEqual` checks against the complete messages; this documented rule violation must be fixed.

[verified]

src/features/inlineScript/setupEnvironment.ts
```diff
@@ -124,12 +124,16 @@ function setupInlineScriptEnvironmentHandler(
     };
 }
 
-function notifyInlineScriptSetupOutcome(uri: Uri, routing: InlineScriptRoutingRegistry): void {
-    const outcome = routing.takeSetupOutcome(uri);
+export function notifyInlineScriptSetupOutcome(uri: Uri, routing: InlineScriptRoutingRegistry): void {
```
rchiodo, line 130: > **Warning** · Non-blocking recommendation

Could a cancel-then-success retry still report cancellation here? This now reads without consuming the stored outcome, while a successful setup does not appear to replace it. Clear or generation-scope outcomes when a new non-coalesced setup begins, and add a cancellation-then-success test.

src/managers/builtin/inlineScript/envManager.ts
```diff
@@ -2228,6 +2281,194 @@ export class InlineScriptEnvManager implements EnvironmentManager, Disposable {
...
+     *
+     * A cache entry is keyed by the script's dependencies and base interpreter, never by its path,
+     * so renaming a script does not invalidate its environment. `PythonProjectManagerImpl` already
+     * rewrites the matching `python-envs.pythonProjects` entry to the new path, so clearing the
+     * association here would leave a managed inline-script project entry with no environment behind
+     * it.
+     *
+     * The moved record is re-validated afterwards rather than trusted: its metadata binding is
+     * content-derived, so if the file at the new path no longer matches, ordinary validation clears
+     * the association and the setup CodeLens returns.
```
rchiodo: > **Info** · Optional note

The rename documentation is attached to `handlePackagesChanged` rather than `handleRenamedScripts`. Move that block directly above the rename handler so each method documents its own invariant.

[verified]

rchiodo:

**Verification:** Isolated verification observed failures that were not classified as caused by this PR\: Node dependency and test discovery preflight. The relevant tests could not be fully run in the isolated environment; this review is not fully verified.

**Summary:** Verification could not proceed because the container lacks Node.js \(\`node\: not found\`\). Consequently, compilation and all three targeted Mocha suites were not run. I identified 19 tests added or substantially rewritten by the PR. Confidence is low until they run in a Node-capable environment.

**Test runs:** 1 failed, 3 not run

- :warning: **Not run** | <code>Inline-script setup UI unit tests</code> | <code>node ./node_modules/mocha/bin/mocha.js --config=./build/.mocha.unittests.json out/test/features/inlineScript/setupEnvironment.unit.test.js</code>
- :warning: **Not run** | <code>Shell startup activation variables unit tests</code> | <code>node ./node_modules/mocha/bin/mocha.js --config=./build/.mocha.unittests.json out/test/features/terminal/shellStartupActivationVariablesManager.unit.test.js</code>
- :warning: **Not run** | <code>Inline-script environment manager unit tests</code> | <code>node ./node_modules/mocha/bin/mocha.js --config=./build/.mocha.unittests.json out/test/managers/builtin/inlineScript/envManager.unit.test.js</code>
- :x: **Failed** | unrelated to this PR | <code>Node dependency and test discovery preflight</code> | <code>printf &#x27;%s\n&#x27; &#x27;== toolchain ==&#x27; &amp;&amp; node --version &amp;&amp; npm --version &amp;&amp; printf &#x27;%s\n&#x27; &#x27;== dependencies ==&#x27; &amp;&amp; if [ -x node_modules/.bin/mocha ] &amp;&amp; [ -x node_modules/.bin/tsc ]; then echo &#x27;node_modules ready&#x27;; else echo &#x27;node_modules missing or incomplete&#x27;; fi &amp;&amp; printf &#x27;%s\n&#x27; &#x27;== unit runner ==&#x27; &amp;&amp; cat build/.mocha.unittests.json &amp;&amp; printf &#x27;%s\n&#x27; &#x27;== targeted source tests ==&#x27; &amp;&amp; find src/test/features/inlineScript src/test/features/terminal src/test/managers/builtin/inlineScript -maxdepth 1 -type f -name &#x27;*.unit.test.ts&#x27; -print | sort</code>

<details><summary>:warning: <code>Inline-script setup UI unit tests</code> diagnostic output</summary>

<pre>Blocked by preflight: /bin/sh: 1: node: not found</pre>
</details>

<details><summary>:warning: <code>Shell startup activation variables unit tests</code> diagnostic output</summary>

<pre>Blocked by preflight: /bin/sh: 1: node: not found</pre>
</details>

<details><summary>:warning: <code>Inline-script environment manager unit tests</code> diagnostic output</summary>

<pre>Blocked by preflight: /bin/sh: 1: node: not found</pre>
</details>

<details><summary>:x: <code>Node dependency and test discovery preflight</code> diagnostic output</summary>

<pre>== toolchain ==
/bin/sh: 1: node: not found</pre>
</details>
</details>

src/test/features/inlineScript/setupEnvironment.unit.test.ts
StellaHuang95: Addressed in 9631d32c.

Replaced all four partial assertions in this file with `assert.strictEqual` against the complete messages, including the parameterized bulk ones (the `l10n.t` test mock performs `{0}` substitution, so the final strings are assertable):

- `'Environment setup was canceled.'`
- `'Environment setup was canceled. Set up 0 of 2 selected inline script environment(s); the remaining 1 were not started.'`
- `'Set up 0 of 1 selected inline script environment(s). 1 failed — see the Python Environments output for details.'`

The separate `assert.doesNotMatch(... /clean up|quarantin|retry|lock/i)` test is now redundant, since exact equality already proves the message says none of those things. I merged it into the cancellation test and kept the intent as a comment rather than leaving a weaker duplicate assertion.

One note for the record: the cited rule lives at `docs/pylancewiki/review/tests-no-partial-asserts.md` in the **pyrx** repository and is not present in this one, so the citation does not apply here as written. The underlying point is right regardless — exact assertions on static user-facing strings catch accidental wording changes, and a partial match on `/failed/i` would have passed against almost any message — so I adopted it.

src/features/inlineScript/setupEnvironment.ts
StellaHuang95: Good catch — this was a real hole, and it reproduces. Fixed in 9631d32c.

You are right that nothing replaced the outcome on success: only the next `create` cleared it on entry, so the invariant depended on a later attempt starting rather than on the successful one finishing. `setUpInlineScriptEnvironment` now clears the outcome once the environment is associated.

Added `retires a previous cancellation once setup succeeds`, and I verified it actually catches the bug by reverting just the new `clearSetupOutcome` call:

```
AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
+ actual   { kind: 'cancelled' }
- expected undefined
```

On generation-scoping: I kept the simpler clear-on-success rather than adding generations, because the outcome map is already scoped to one attempt from both ends — `create` clears on entry and success now clears on exit — and the non-consuming read is what lets coalesced callers observe the same result. Happy to revisit if you would rather see explicit generations.

src/managers/builtin/inlineScript/envManager.ts
StellaHuang95: Correct, and thanks — this was my mistake. Fixed in 9631d32c.

The rename doc block was written above `handleRenamedScripts`, and when `handlePackagesChanged` was later inserted immediately before that method the comment was left behind, so the file ended up with two stacked doc comments and the rename invariant documenting the package-change handler. Moved it back so each method documents its own invariant.

rchiodo:

**Verification:** Isolated verification observed failures that were not classified as caused by this PR\: Runtime, dependency, test, and change discovery. The relevant tests could not be fully run in the isolated environment; this review is not fully verified.

**Summary:** Verification could not proceed because the disposable container has no Node.js executable \(\`node\: not found\`\). Consequently, none of the four targeted unit-test files could be compiled or executed. The PR adds or rewrites 19 focused tests across cancellation, rename, shared builds, terminal scoping, and package drift. Confidence is limited to test discovery and source inspection.

**Test runs:** 1 failed, 4 not run

- :warning: **Not run** | <code>Inline-script setup environment unit tests</code> | <code>node ./node_modules/mocha/bin/mocha.js --require source-map-support/register --require out/test/unittests.js --ui tdd --timeout 180000 out/test/features/inlineScript/setupEnvironment.unit.test.js</code>
- :warning: **Not run** | <code>Shell startup activation variables manager unit tests</code> | <code>node ./node_modules/mocha/bin/mocha.js --require source-map-support/register --require out/test/unittests.js --ui tdd --timeout 180000 out/test/features/terminal/shellStartupActivationVariablesManager.unit.test.js</code>
- :warning: **Not run** | <code>Inline-script environment manager unit tests</code> | <code>node ./node_modules/mocha/bin/mocha.js --require source-map-support/register --require out/test/unittests.js --ui tdd --timeout 180000 out/test/managers/builtin/inlineScript/envManager.unit.test.js</code>
- :warning: **Not run** | <code>Inline-script cache layout unit tests</code> | <code>node ./node_modules/mocha/bin/mocha.js --require source-map-support/register --require out/test/unittests.js --ui tdd --timeout 180000 out/test/common/inlineScript/cacheLayout.unit.test.js</code>
- :x: **Failed** | unrelated to this PR | <code>Runtime, dependency, test, and change discovery</code> | <code>printf &#x27;%s\n&#x27; &#x27;== runtime ==&#x27; &amp;&amp; node --version &amp;&amp; npm --version &amp;&amp; printf &#x27;%s\n&#x27; &#x27;== dependency state ==&#x27; &amp;&amp; if [ -x node_modules/.bin/mocha ] &amp;&amp; [ -x node_modules/.bin/tsc ]; then echo &#x27;node_modules ready&#x27;; else echo &#x27;node_modules missing or incomplete&#x27;; fi &amp;&amp; printf &#x27;%s\n&#x27; &#x27;== test configuration ==&#x27; &amp;&amp; node -e &quot;const p=require(&#x27;./package.json&#x27;); console.log(JSON.stringify({compileTests:p.scripts[&#x27;compile-tests&#x27;],unittest:p.scripts.unittest},null,2))&quot; &amp;&amp; cat build/.mocha.unittests.json &amp;&amp; printf &#x27;%s\n&#x27; &#x27;== refs ==&#x27; &amp;&amp; git status --short &amp;&amp; git log -8 --oneline &amp;&amp; printf &#x27;%s\n&#x27; &#x27;== changed files against merge base ==&#x27; &amp;&amp; base=$(git merge-base HEAD origin/main 2&gt;/dev/null || git merge-base HEAD main 2&gt;/dev/null) &amp;&amp; echo &quot;base=$base&quot; &amp;&amp; git diff --name-status &quot;$base&quot;..HEAD</code>

<details><summary>:warning: <code>Inline-script setup environment unit tests</code> diagnostic output</summary>

<pre>Preflight failed: /bin/sh: 1: node: not found</pre>
</details>

<details><summary>:warning: <code>Shell startup activation variables manager unit tests</code> diagnostic output</summary>

<pre>Preflight failed: /bin/sh: 1: node: not found</pre>
</details>

<details><summary>:warning: <code>Inline-script environment manager unit tests</code> diagnostic output</summary>

<pre>Preflight failed: /bin/sh: 1: node: not found</pre>
</details>

<details><summary>:warning: <code>Inline-script cache layout unit tests</code> diagnostic output</summary>

<pre>Preflight failed: /bin/sh: 1: node: not found</pre>
</details>

<details><summary>:x: <code>Runtime, dependency, test, and change discovery</code> diagnostic output</summary>

<pre>== runtime ==
/bin/sh: 1: node: not found</pre>
</details>
</details>

rchiodo (approved):
Approved via Review Center.

PR #1777, opened 2026-09-10
Warn when companion extensions are too old for inline-script envs

StellaHuang95:
## Problem

A PEP 723 inline-script environment is built correctly by this extension on its own, but it only reaches the language server once two companion changes ship:

- **Python extension** resolves interpreters per file (`exactResource`, microsoft/vscode-python#26129)
- **Pylance** handles the file-scoped `python/didChangeFilePythonPath` notification (PR #9302)

Until both are present, setup reports success and the user is left without the full inline script experience, with nothing explaining why.

## Change

Warn once, **after a successful setup** rather than before one. That is the moment the gap becomes visible, and the creation progress notification is already gone — prompting up front would be noise stacked on top of it.

Both entry points prompt at the handler level (`setupInlineScriptEnvironmentHandler` and `setUpInlineScriptEnvironmentsInWorkspace`), so a bulk run over N scripts shows **one** notification rather than N.

The prompt offers **Update Extension** and **Don't Show Again**; the latter persists in global state. Dismissing without choosing an action suppresses it for the session only.

## Version handling

| | stable | pre-release |
|---|---|---|
| Python | `2026.4.0` | `2026.7.2026082601` |
| Pylance | `2026.3.1` | `2026.3.101` |

Values are the newest version on each channel that still **lacks** the change; anything newer passes.

Notes on why this is more involved than a single `>=`:

- **Per channel, because the two lines interleave numerically.** Pre-release `2026.5.x`/`2026.7.x` sort *above* stable `2026.4.0`, so one threshold cannot express both — a future stable such as `2026.6.0` would be wrongly rejected.
- **Channel is inferred from the patch component.** VS Code does not populate `__metadata.preRelease` for installed extensions (verified empty on every installed extension locally, including a known pre-release build). Python pre-releases carry a build-date patch, Pylance pre-releases use a 100+ counter. Both rules were checked against every published version above the relevant floors.
- **PEP 440 ordering**, since these extensions are not semver.
- **Dev builds are skipped, not compared.** `2026.7.0-dev` is valid PEP 440 and normalizes to `2026.7.0.dev0`, which sorts below every real build of the same minor — comparing it would flag anyone running a local build of either extension.
- **A missing extension is not outdated.** Pylance is optional.
- **Pylance is skipped unless `python.languageServer` selects it**, so Jedi and None users are not asked to update an extension they do not use.

Each outdated combination has its own complete l10n string rather than a joined extension list, since conjunctions and word order are locale-specific.

## Follow-up

The thresholds are a snapshot of the newest version lacking the change. **If either extension ships another release before the change lands, the corresponding constant must be bumped**, or the warning silently stops firing for users on that release.

Once the real shipping versions are known, consider inverting to "first version *with* the change" (`<` instead of `<=`): that framing fails toward an over-eager prompt, which is visible and reported, rather than toward silence.

## Testing

36 new unit tests in `src/test/features/inlineScript/extensionVersionCheck.unit.test.ts`, covering both channels for both extensions, the dev-build case, the language-server gate, and all dismissal paths.

Full unit suite: **2047 passing, 0 failing** (clean `out/` rebuild). `tsc --noEmit`, `eslint src`, and `prettier --check` all clean.

### DIFF
--- a/src/features/inlineScript/extensionVersionCheck.ts
+++ b/src/features/inlineScript/extensionVersionCheck.ts
@@ -1,0 +1,123 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import { compare as pep440Compare } from '@renovatebot/pep440';
+import { PYLANCE_EXTENSION_ID, PYTHON_EXTENSION_ID } from '../../common/constants';
+import { getComparableExtensionVersion } from '../../common/extVersion';
+import { Common, InlineScriptStrings } from '../../common/localize';
+import { traceInfo, traceVerbose } from '../../common/logging';
+import { getGlobalPersistentState } from '../../common/persistentState';
+import { showWarningMessage } from '../../common/window.apis';
+import { openExtension } from '../../common/workbenchCommands';
+import { getConfiguration } from '../../common/workspace.apis';
+
+export const INLINE_SCRIPT_UPDATE_EXTENSIONS_DONT_SHOW_KEY = 'python-envs:inline-script:UPDATE_EXTENSIONS_DONT_SHOW';
+
+interface CompanionExtension {
+    readonly id: string;
+    /** Newest version on each channel that still LACKS the required change; anything newer is fine. */
+    readonly lastUnsupportedStable: string;
+    readonly lastUnsupportedPreRelease: string;
+    /** Patch component at or above which a version is a pre-release build. */
+    readonly preReleasePatchFloor: number;
+}
+
+/**
+ * Python needs per-file interpreter resolution (`exactResource`, PR #26129, merged 2026-08-31);
+ * Pylance needs the `python/didChangeFilePythonPath` notification (PR #9302, merged 2026-09-01).
+ */
+const PYTHON_COMPANION: CompanionExtension = {
+    id: PYTHON_EXTENSION_ID,
+    lastUnsupportedStable: '2026.4.0',
+    lastUnsupportedPreRelease: '2026.7.2026082601',
+    preReleasePatchFloor: 1_000_000,
+};
+
+const PYLANCE_COMPANION: CompanionExtension = {
+    id: PYLANCE_EXTENSION_ID,
+    lastUnsupportedStable: '2026.3.1',
+    lastUnsupportedPreRelease: '2026.3.101',
+    preReleasePatchFloor: 100,
+};
+
+let promptShownThisSession = false;
+
+export function resetInlineScriptExtensionPromptForTests(): void {
+    promptShownThisSession = false;
+}
+
+function isPylanceInUse(): boolean {
+    const languageServer = getConfiguration('python').get<string>('languageServer', 'Default');
+    return languageServer === 'Default' || languageServer === 'Pylance';
+}
+
+function isPreReleaseBuild(version: string, preReleasePatchFloor: number): boolean {
+    const patch = Number((version.split('.')[2] ?? '').replace(/\D.*$/, ''));
+    return Number.isFinite(patch) && patch >= preReleasePatchFloor;
+}
+
+function isOutdated(extension: CompanionExtension): boolean {
+    const resolved = getComparableExtensionVersion(extension.id);
+    if (resolved.kind !== 'version') {
+        traceVerbose(`inline-script companion check: ${extension.id} -> ${resolved.kind}`);
+        return false;
+    }
+    const lastUnsupported = isPreReleaseBuild(resolved.version, extension.preReleasePatchFloor)
+        ? extension.lastUnsupportedPreRelease
+        : extension.lastUnsupportedStable;
+    const outdated = pep440Compare(resolved.version, lastUnsupported) <= 0;
+    if (outdated) {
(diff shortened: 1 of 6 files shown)


rchiodo:

src/features/inlineScript/setupEnvironment.ts
```diff
@@ -115,7 +116,9 @@ function setupInlineScriptEnvironmentHandler(
             const environment = await setUpInlineScriptEnvironment(uri, em, routing);
             if (!environment) {
                 notifyInlineScriptSetupOutcome(uri, routing);
+                return;
             }
+            await promptUpdateExtensionsForInlineScripts();
```
rchiodo, line 121: > **Warning** · Non-blocking recommendation

If the advisory prompt rejects after environment creation, this handler reports setup as failed. Isolate prompt failures from the setup `try` block, matching the bulk path, and add a rejection test.

src/features/inlineScript/extensionVersionCheck.ts
```diff
@@ -0,0 +1,123 @@
...
+    const names = outdated.map((extension) => extension.id).join(', ');
+    traceInfo(`Inline-script companion extensions out of date: ${names}`);
+
+    const result = await showWarningMessage(
+        getOutdatedMessage(outdated),
+        InlineScriptStrings.updateExtension,
+        Common.dontShowAgain,
+    );
+
+    if (result === InlineScriptStrings.updateExtension) {
```
rchiodo, line 117: > **Warning** · Non-blocking recommendation

When both extensions are outdated, the combined prompt names both but `outdated[0]` opens only Python, and the session latch prevents another prompt. Provide separate actions or make every named update reachable.

rchiodo:

**Verification:** Isolated verification observed failures that were not classified as caused by this PR\: Offline npm dependency bootstrap. The relevant tests could not be fully run in the isolated environment; this review is not fully verified.

**Summary:** No meaningful tests could run because the container lacked Node.js, npm, and \`node\_modules\`; \`npm ci --offline\` failed with \`npm\: not found\`. The two targeted inline-script unit suites were therefore not run. I found 37 newly added tests covering version classification, thresholds, language-server gating, and prompt actions. Explicit coverage is still missing for invoking the prompt from the single-file and bulk setup entry points.

**Test runs:** 1 passed, 1 failed, 2 not run

- :warning: **Not run** | <code>Inline-script extension version check unit suite</code> | <code>npm run compile-tests &amp;&amp; node ./node_modules/mocha/bin/mocha.js --require source-map-support/register --require out/test/unittests.js --ui tdd --timeout 180000 out/test/features/inlineScript/extensionVersionCheck.unit.test.js</code>
- :warning: **Not run** | <code>Inline-script setup environment regression suite</code> | <code>npm run compile-tests &amp;&amp; node ./node_modules/mocha/bin/mocha.js --require source-map-support/register --require out/test/unittests.js --ui tdd --timeout 180000 out/test/features/inlineScript/setupEnvironment.unit.test.js</code>
- :x: **Failed** | unrelated to this PR | <code>Offline npm dependency bootstrap</code> | <code>npm ci --offline</code>
- :white_check_mark: **Passed** | <code>Dependency and targeted-test discovery</code> | <code>printf &#x27;%s\n&#x27; &#x27;=== changed files ===&#x27;; git diff --name-status upstream/main...HEAD; printf &#x27;%s\n&#x27; &#x27;=== relevant package scripts ===&#x27;; node -e &quot;const p=require(&#x27;./package.json&#x27;); for (const [k,v] of Object.entries(p.scripts||{})) if (/test|compile|lint|prettier/.test(k)) console.log(k+&#x27;=&#x27;+v)&quot;; printf &#x27;%s\n&#x27; &#x27;=== dependency and build state ===&#x27;; if [ -d node_modules ]; then echo &#x27;node_modules=present&#x27;; else echo &#x27;node_modules=missing&#x27;; fi; if [ -x node_modules/.bin/jest ]; then echo &#x27;jest=present&#x27;; else echo &#x27;jest=missing&#x27;; fi; if [ -x node_modules/.bin/mocha ]; then echo &#x27;mocha=present&#x27;; else echo &#x27;mocha=missing&#x27;; fi; if [ -d out ]; then echo &#x27;out=present&#x27;; else echo &#x27;out=missing&#x27;; fi; printf &#x27;%s\n&#x27; &#x27;=== targeted test files ===&#x27;; find src/test/features/inlineScript -maxdepth 1 -type f -name &#x27;*test.ts&#x27; -print | sort</code>

<details><summary>:warning: <code>Inline-script extension version check unit suite</code> diagnostic output</summary>

<pre>Not run because the offline dependency bootstrap failed: /bin/sh: npm: not found.</pre>
</details>

<details><summary>:warning: <code>Inline-script setup environment regression suite</code> diagnostic output</summary>

<pre>Not run because the offline dependency bootstrap failed: /bin/sh: npm: not found.</pre>
</details>

<details><summary>:x: <code>Offline npm dependency bootstrap</code> diagnostic output</summary>

<pre>/bin/sh: 1: npm: not found</pre>
</details>
</details>

rchiodo (approved):
Approved via Review Center.

PR #1785, opened 2026-09-14
feat: report diagnostics for malformed PEP 723 inline script metadata

StellaHuang95:
Part of #1602 (PEP 723 inline script env support). Does not close it — this covers only diagnostics for malformed metadata.

## Problem

Malformed PEP 723 metadata is currently **invisible**. The parser collapses every failure to `undefined`, so a block with a typo — a missing closing `# ///`, a content line without the required space after `#`, a TOML syntax error — produces no CodeLens, no warning, nothing. The file just looks like ordinary Python and the user has no idea why the "Setup environment" action never appeared.

Worse, a malformed block and a file with *no* block were indistinguishable, so we couldn't have told the user even if we wanted to.

## What this does

The parser now returns a discriminated result — `valid` / `invalid` / `none` — carrying structured problems with source ranges, and a new `InlineScriptDiagnosticsPublisher` surfaces them as squiggles.

Gated behind the existing `python-envs.inlineScripts.enabled` setting (default `false`, latched at activation). No Pylance or Python extension changes; no cross-team dependencies.

---

## Spec conformance

Normative source: [PyPA — Inline script metadata](https://packaging.python.org/en/latest/specifications/inline-script-metadata/). Permalinks below are pinned to commit [`9dd85f2`](https://github.com/pypa/packaging.python.org/blob/9dd85f27266c88c4eca58828c918c337b665d331/source/specifications/inline-script-metadata.rst), the latest commit touching that file, so line numbers can't drift.

I extracted every RFC 2119 keyword sentence from the spec source programmatically: **20 total — 11 MUST, 2 MUST NOT, 1 SHOULD, 7 MAY.** The mapping below is against that complete set, not a sampling.

### Diagnostics backed by an explicit spec MUST

| Code | Severity | Spec text | Strength | Source |
|---|---|---|---|---|
| `invalid-block-marker` | Error | "blocks that MUST start with the line `# /// TYPE`… Block MUST end with the line `# ///`… The `TYPE` MUST only consist of ASCII letters, numbers and hyphens" | MUST x3 | [L20-L26](https://github.com/pypa/packaging.python.org/blob/9dd85f27266c88c4eca58828c918c337b665d331/source/specifications/inline-script-metadata.rst#L20-L26) |
| `invalid-content-line` | Error | "Every line between these two lines MUST be a comment starting with `#`. If there are characters after the `#` then the first character MUST be a space" | MUST x2 | [L28-L30](https://github.com/pypa/packaging.python.org/blob/9dd85f27266c88c4eca58828c918c337b665d331/source/specifications/inline-script-metadata.rst#L28-L30) |
| `multiple-blocks` | Error | "When there are multiple comment blocks of the same `TYPE` defined, tools MUST produce an error" | MUST | [L54-L55](https://github.com/pypa/packaging.python.org/blob/9dd85f27266c88c4eca58828c918c337b665d331/source/specifications/inline-script-metadata.rst#L54-L55) |
| `unterminated-block` | **Warning** | "Unclosed blocks MUST be ignored" | MUST | [L51-L52](https://github.com/pypa/packaging.python.org/blob/9dd85f27266c88c4eca58828c918c337b665d331/source/specifications/inline-script-metadata.rst#L51-L52) |

`unterminated-block` is a **warning, not an error**, precisely because the spec says such blocks MUST be *ignored* — the file is valid Python and valid per spec. We only hint that the author probably meant to close it. See the scoping rule below.

Also enforced without needing a diagnostic:

| Spec requirement | Strength | How we conform | Source |
|---|---|---|---|
| "Tools MUST NOT read from metadata blocks with types that have not been standardized" | MUST NOT | `type === 'script'` filter in both the block regex and the opener scan | [L70-L71](https://github.com/pypa/packaging.python.org/blob/9dd85f27266c88c4eca58828c918c337b665d331/source/specifications/inline-script-metadata.rst#L70-L71) |
| "If they choose not to [respect the encoding declaration], they MUST process the file as UTF-8" | MAY / MUST | We do not honour the encoding declaration; we process as UTF-8 and strip BOM | [L57-L58](https://github.com/pypa/packaging.python.org/blob/9dd85f27266c88c4eca58828c918c337b665d331/source/specifications/inline-script-metadata.rst#L57-L58) |
| "Tools MAY choose to do a simple textual scan, rather than a full Python parse" | MAY | Textual scan, per the canonical regex | [L77](https://github.com/pypa/packaging.python.org/blob/9dd85f27266c88c4eca58828c918c337b665d331/source/specifications/inline-script-metadata.rst#L77) |
| "the text specification takes precedence" over the canonical regex | tiebreak | Applied deliberately — see empty blocks below | [L67-L68](https://github.com/pypa/packaging.python.org/blob/9dd85f27266c88c4eca58828c918c337b665d331/source/specifications/inline-script-metadata.rst#L67-L68) |

### Deliberately NOT reported

These are conformance decisions, not oversights. Each is measured by a test.

| Case | Why we stay silent |
|---|---|
| A start line nested inside another block | Spec says tools **MAY** produce an error — MAY, not MUST. We take the least-aggressive legal option. [L51](https://github.com/pypa/packaging.python.org/blob/9dd85f27266c88c4eca58828c918c337b665d331/source/specifications/inline-script-metadata.rst#L51) |
| **Empty block** (`# /// script` immediately followed by `# ///`) | The canonical regex requires >=1 content line (`+`), but the prose doesn't. Spec says prose wins, so this is **valid, empty metadata**. `BLOCK_RE`'s content group is `*` with a comment marking the deliberate deviation. |
| Unclosed `# /// script` appearing *below* real code | Almost always a PEP 723 example inside a docstring or tutorial text, not a broken header. New `headerRegionEnd()` suppresses the warning at/after the first non-blank, non-`#` line. The spec explicitly says behaviour inside multi-line strings "is tool-dependent and should not be relied on". |
| `# ///` used as block *content* | Spec's precedence rule plus its own embedded-C# example. Handled by regex backtracking; verified against the spec's literal example. |
| Divider comments (`# /////////`), `## ///`, non-`script` types | Not blocks. |

### Checks that are **ours**, not the spec's

Worth calling out explicitly so reviewers can push back:

**`invalid-field-type`** (Error) fires when `dependencies` isn't an array of strings, `requires-python` isn't a string, or `tool` isn't a table. The spec describes these types in prose — "`dependencies`: A list of strings…", "`requires-python`: A string…" ([L99-L104](https://github.com/pypa/packaging.python.org/blob/9dd85f27266c88c4eca58828c918c337b665d331/source/specifications/inline-script-metadata.rst#L99-L104)) — **but attaches no MUST to the container type.** The only MUST in those bullets governs entry *validity*.

I kept it as an error because a non-array `dependencies` has no valid interpretation and the alternative is a confusing downstream resolver failure. But it is inference, not mandate. Happy to downgrade to a warning if reviewers prefer strict spec-only conformance.

**`invalid-toml`** (Error) — the spec never literally states the content is TOML; it's implied by the `[tool]` table semantics and the reference implementation's `tomllib.loads(content)`.

### Known gaps (not addressed here)

Two genuine spec MUSTs are **not** enforced. Verified by measurement — both parse as valid with no diagnostic today:

| Spec text | Source | Status |
|---|---|---|
| "Each entry MUST be a valid [dependency specifier](https://packaging.python.org/en/latest/specifications/dependency-specifiers/)" (PEP 508) | [L99-L101](https://github.com/pypa/packaging.python.org/blob/9dd85f27266c88c4eca58828c918c337b665d331/source/specifications/inline-script-metadata.rst#L99-L101) | Not validated — `dependencies = ["!!! not a specifier !!!"]` accepted silently |
| "The value of this field MUST be a valid [version specifier](https://packaging.python.org/en/latest/specifications/version-specifiers/)" (PEP 440) | [L102-L104](https://github.com/pypa/packaging.python.org/blob/9dd85f27266c88c4eca58828c918c337b665d331/source/specifications/inline-script-metadata.rst#L102-L104) | Not validated — `requires-python = "totally-not-a-version"` accepted silently |

Both need real grammar implementations and would meaningfully expand scope. Deferred intentionally; happy to file a follow-up issue.

---

## Behaviour change reviewers must notice

This PR is **not purely additive.** One existing behaviour changes:

```python
# /// script
# ///
```

**Before:** parsed as no metadata -> no CodeLens.
**After:** parses as valid, empty metadata -> **the setup CodeLens now appears.**

That follows from the spec's "text takes precedence" tiebreak, and I believe it's correct, but it does affect provisioning and not just diagnostics. Everything else in this PR is diagnostics-only and leaves the provisioning path byte-identical — I verified that with a harness running the old and new parsers side by side over 20 scenarios.

---

## Implementation notes

**`src/common/inlineScript/metadata.ts`** — the parser rework.
- Result is now `{ kind: 'valid' | 'invalid' | 'none', problems, metadata? }`.
- New independent opener scan (`findScriptOpeners`) — this is what makes "malformed block" distinguishable from "no block". The canonical regex alone can't tell them apart.
- `toSourceRange` is the **single** offset-to-line/character translation point. Reviewers should focus here for range correctness.
- `tomlErrorSourceRange` recovers `line`/`col` from `@iarna/toml` errors that were previously discarded, so the squiggle underlines the actual offending text instead of the whole block.

**`src/features/inlineScript/diagnostics.ts`** — the publisher.
- Publishes on open, save, and debounced change (**300 ms**, via the existing `createSimpleDebounce` helper). Debouncing matters here: without it we'd flash "missing closing marker" at someone who is still mid-edit.
- `createSimpleDebounce(ms, cb).trigger()` takes no arguments and owns one timer, so a shared instance would let documents clobber each other — hence the per-URI `pending` map.
- Clears on close, delete, rename, and dispose, so no stale squiggles survive.
- Scope: `file:` scheme plus `.py` extension only.

**Validation reads the live editor buffer**, so squiggles match what's on screen; provisioning keeps reading the saved file. Both call the same parser, so the two views can't disagree about what "valid" means.

### Edge cases handled

- **BOM and CRLF** — BOM stripped, `\r\n?` normalized to `\n` before matching. Note JS regex `.` doesn't match `\r` while Python's does, so the canonical regex can't be used verbatim.
- **Ranges under BOM/CRLF** — the bulk of the test suite. Every range assertion runs across four variants: LF, CRLF, BOM+LF, BOM+CRLF.
- **8 KiB header slice** — diagnostics apply `sliceHeaderBytes` exactly as the provisioning path does, so the two can't diverge on large files.
- **Adjacent blocks merge** — `# ///` is itself a legal content line, so back-to-back blocks are consumed as one. `multiple-blocks` correctly requires a non-content line between them.

### Known limitation

Indented markers (`    # /// script`) are silent. The spec anchors markers at column 0, so an indented one isn't a block at all. Documented in the parser.

---

## Testing

- **52 new unit tests** (31 parser/range, 21 publisher).
- Full suite: **2280 passing, 0 failing, 6 pending**, on a fresh `npm run compile-tests` build.
- `npm run lint` clean, `tsc` clean, Prettier clean.

> Note for anyone re-running: `npm run unittest` executes `./out/test/**/*.unit.test.js` — **compiled output**. Run `npm run compile-tests` first or you'll silently validate stale code.

### Manual verification

Set `python-envs.inlineScripts.enabled: true` and **reload the window** (the flag is latched at activation). Then try:

```python
# /// script
# requires-python = ">=3.11"
# dependencies = ["requests"]
```
-> warning: missing closing `# ///`

```python
# /// script
#requires-python = ">=3.11"
# ///
```
-> error: content line must be exactly `#` or start with `# `

```python
# /// script
# dependencies = [
# ///
```
-> error: TOML error, underlining the offending token

## Review focus

1. **`toSourceRange` correctness** — every range flows through it; BOM/CRLF offsets are the likeliest bug.
2. **Is `invalid-field-type` too aggressive?** It's the one error not backed by a spec MUST.
3. **The empty-block behaviour change** — it alters provisioning, not just diagnostics.
4. **Debounce lifecycle** — per-URI map, disposal on close/delete/rename.


rchiodo:

src/common/inlineScript/metadata.ts
```diff
@@ -269,16 +386,176 @@ function countLines(text: string, offset: number): number {
...
+        if (lineEnd >= text.length) {
+            break;
+        }
+        offset = lineEnd + 1;
+    }
+
+    traceWarn(
+        `inline script metadata${where}: the \`# /// script\` block on line ${countLines(text, opener.offset)} is missing its closing \`# ///\` marker`,
+    );
+    return { code: 'unterminated-block', severity: 'warning', sourceRange: openerRange };
```
rchiodo: > **Issue** · Please address or respond

A non-comment line followed by `# ///` is reported as `unterminated-block` even though the closer exists. Search ahead for the closer and report the offending line as `invalid-content-line`, then add this regression case.

src/features/inlineScript/diagnostics.ts
```diff
@@ -0,0 +1,223 @@
...
+            return;
+        }
+        for (const published of Array.from(this.published.values())) {
+            if (published.scheme === 'file' && isSameOrParentPath(uri.fsPath, published.fsPath)) {
+                this.clear(published);
+            }
+        }
+    }
+
+    private cancelPending(uri: Uri): void {
```
rchiodo, line 174: > **Warning** · Non-blocking recommendation

`clearTree` cancels only the exact URI and iterates descendants only in `published`, so deleting or renaming a folder during the 300 ms debounce can leave pending child validations that later publish stale diagnostics. Cancel descendant entries in `pending` as well and add a fake-timer folder-operation test.

src/common/inlineScript/metadata.ts
```diff
@@ -48,6 +40,30 @@ export interface InlineScriptMetadata {
...
+    readonly severity: 'error' | 'warning';
+    readonly sourceRange: { readonly start: number; readonly end: number };
+    readonly detail?: string;
+}
+
+export type InlineScriptMetadataParseResult =
+    | {
+          readonly kind: 'valid';
+          readonly metadata: InlineScriptMetadata;
+          readonly problems: readonly InlineScriptMetadataProblem[];
```
rchiodo, line 68: > **Warning** · Non-blocking recommendation

`kind: 'valid'` can include error or warning problems, so the discriminant means metadata is usable rather than input is valid. Rename it to reflect metadata availability or model `metadata` and `problems` independently before more consumers adopt this contract.

src/test/common/inlineScript/metadataDiagnostics.unit.test.ts
```diff
@@ -0,0 +1,349 @@
...
+
+        test('detail carries the raw parser message without coordinates', () => {
+            const source = ['# /// script', '# dependencies = ["requests', '# ///'].join('\n');
+            const problem = onlyProblem(parseInlineScriptMetadata(source));
+            assert.ok(problem.detail && problem.detail.length > 0, 'expected a detail message');
+            assert.ok(
+                !/row \d+, col \d+/.test(problem.detail!),
+                `detail should not leak payload coordinates: ${problem.detail}`,
+            );
+        });
```
rchiodo, line 234: > **Issue** · Please address or respond

This violates `docs/pylancewiki/review/tests-no-partial-asserts.md`: the test checks only that `detail` is non-empty and omits coordinates. Assert the complete expected detail string, normalizing any genuinely variable portion first.

src/test/features/inlineScript/diagnostics.unit.test.ts
```diff
@@ -0,0 +1,318 @@
...
+    suite('publishing', () => {
+        test('a malformed block produces a diagnostic on open', () => {
+            openListener?.(makeDoc(scriptUri, BROKEN_SCRIPT));
+
+            const published = sink.for(scriptUri);
+            assert.ok(published, 'expected diagnostics to be published');
+            assert.strictEqual(published!.length, 1);
+            assert.strictEqual(published![0].code, 'unterminated-block');
+            assert.strictEqual(published![0].severity, DiagnosticSeverity.Warning);
+            assert.ok(published![0].message.length > 0, 'expected a localized message');
```
rchiodo: > **Issue** · Please address or respond

This violates `docs/pylancewiki/review/tests-no-partial-asserts.md` by checking only that the user-facing diagnostic message is non-empty. Assert the full localized message exactly.

rchiodo:

**Summary:** Targeted parser regression and diagnostics publisher tests were identified, including all 52 tests added by the PR. Execution could not begin because the container has neither \`node\` nor \`npm\`; the targeted compile-and-test command exited 127. Consequently, no meaningful test passed or failed, and confidence is low.

**Test runs:** 2 not run

- :warning: **Not run** | <code>Inline script parser and diagnostics targeted unit tests</code> | <code>npm run compile-tests &amp;&amp; node ./node_modules/mocha/bin/mocha.js --config=./build/.mocha.unittests.json out/test/common/inlineScript/metadata.unit.test.js out/test/common/inlineScript/metadataDiagnostics.unit.test.js out/test/features/inlineScript/diagnostics.unit.test.js</code>
- :warning: **Not run** | <code>Node dependency and test-discovery preflight</code> | <code>printf &#x27;%s\n&#x27; &#x27;== toolchain ==&#x27; &amp;&amp; node --version &amp;&amp; npm --version &amp;&amp; printf &#x27;%s\n&#x27; &#x27;== dependency state ==&#x27; &amp;&amp; if [ -d node_modules ]; then echo &#x27;node_modules=present&#x27;; else echo &#x27;node_modules=missing&#x27;; fi &amp;&amp; if [ -f package-lock.json ]; then echo &#x27;lockfile=package-lock.json&#x27;; elif [ -f pnpm-lock.yaml ]; then echo &#x27;lockfile=pnpm-lock.yaml&#x27;; else echo &#x27;lockfile=missing&#x27;; fi &amp;&amp; printf &#x27;%s\n&#x27; &#x27;== relevant scripts ==&#x27; &amp;&amp; node -e &quot;const p=require(&#x27;./package.json&#x27;); for (const k of [&#x27;compile-tests&#x27;,&#x27;unittest&#x27;,&#x27;lint&#x27;]) console.log(k+&#x27;=&#x27;+p.scripts[k])&quot; &amp;&amp; printf &#x27;%s\n&#x27; &#x27;== mocha config ==&#x27; &amp;&amp; cat build/.mocha.unittests.json &amp;&amp; printf &#x27;%s\n&#x27; &#x27;== targeted source/compiled tests ==&#x27; &amp;&amp; find src/test/common/inlineScript src/test/features/inlineScript -maxdepth 1 -type f \( -name &#x27;metadata*.unit.test.ts&#x27; -o -name &#x27;diagnostics.unit.test.ts&#x27; \) -print &amp;&amp; find out/test/common/inlineScript out/test/features/inlineScript -maxdepth 1 -type f \( -name &#x27;metadata*.unit.test.js&#x27; -o -name &#x27;diagnostics.unit.test.js&#x27; \) -print 2&gt;/dev/null || true</code>

<details><summary>:warning: <code>Inline script parser and diagnostics targeted unit tests</code> diagnostic output</summary>

<pre>/bin/sh: 1: npm: not found
[container exit=127]</pre>
</details>

<details><summary>:warning: <code>Node dependency and test-discovery preflight</code> diagnostic output</summary>

<pre>== toolchain ==
/bin/sh: 1: node: not found</pre>
</details>
</details>

src/common/inlineScript/metadata.ts
StellaHuang95: Good catch - reproduced and fixed in f5027d86.

The root cause was in `diagnoseMalformedBlock`: a line that is not a valid content line but does *not* start with `#` fell straight to `break`, which then reported `unterminated-block` regardless of whether a closer existed. So `#requires-python` (bad spacing) correctly produced `invalid-content-line`, while `not_a_comment = 1` produced a factually wrong "missing closing marker" - an inconsistency with no visible justification to the user.

The most common trigger turned out not to be stray code but a **blank line between fields**, which breaks block candidacy the same way:

```python
# /// script
# requires-python = ">=3.11"

# dependencies = ["requests"]
# ///
```

`hasCloserAhead` now scans forward for the closer, stopping at the next `# /// ` opener so a later block's closer cannot absolve an earlier unclosed one. When a closer is found we report `invalid-content-line` (error) on the offending line; otherwise `unterminated-block` (warning) is unchanged. Blank lines get a dedicated message rather than an empty `Found: `, and the range is widened by one character so a zero-width squiggle stays visible.

One deliberate limit: the reclassification is suppressed below the leading comment region, matching the existing rule for unclosed blocks. Without that, a malformed PEP 723 example inside a docstring below real code would start squiggling - measured and confirmed before tightening it.

Regression cases added: non-comment line with a closer, blank line with a closer, non-comment line with *no* closer (still a warning), a closer belonging to a later block, and the docstring case staying silent - each across the LF / CRLF / BOM+LF / BOM+CRLF variants.

src/features/inlineScript/diagnostics.ts
StellaHuang95: Agreed, and fixed in f5027d86.

Confirmed by reading the code path: `clearTree` called `clear(uri)` for the exact URI and then iterated only `this.published`. A child whose debounce was still pending and had never published was invisible to that loop, so the sequence

1. edit `folder/script.py` (300 ms debounce starts, nothing published yet),
2. delete or rename `folder/` inside that window,
3. timer fires

published diagnostics for a file that no longer exists. Narrow window, but branch switches and bulk refactors routinely delete files right after edits.

`clearTree` now sweeps both maps, taking `entry.document.uri` for pending entries; `clear` already cancels the debounce and is idempotent, so no other change was needed.

Added two fake-timer tests covering the folder delete and the folder rename: each queues a change for a nested file, fires the folder event, advances the clock past the debounce, and asserts nothing was published.

src/common/inlineScript/metadata.ts
StellaHuang95: You are right, and this one corrected me. I had assumed only warnings could accompany `valid` because every `severity: 'error'` push is followed by `return { kind: 'invalid', problems }`. That is true of the matched-block path but not of the opener scan, so I measured it instead of trusting the assumption:

| input | result |
| --- | --- |
| valid block + later opener with a bad content line | `kind=valid`, `metadata` present, `error:invalid-content-line` |
| valid block + later opener with a trailing-space closer | `kind=valid`, `metadata` present, `error:invalid-block-marker` |

So a consumer writing `if (result.kind === 'valid')` and assuming clean input would silently swallow errors - exactly the trap you describe, and cheapest to fix now while this is still new surface with no external consumers.

Renamed the discriminant to `parsed` in f5027d86, which states what it actually guarantees (metadata is usable), and documented on the type that `problems` may carry `error` severities even then. Added a test that pins the contract so it cannot regress quietly.

src/test/common/inlineScript/metadataDiagnostics.unit.test.ts
StellaHuang95: Partly agreed - tightened in f5027d86.

On the substance you have a point I had missed: the test is named "detail carries the raw parser message", but nothing actually verified that it carries *the parser message*, only that it was non-empty. It now asserts the exact value (`Unterminated string`). I kept the negative coordinate assertion alongside it, because omitting `row N, col N` is the invariant the test exists to protect - the coordinates are recovered separately and turned into the squiggle range, and leaking them into the human-readable text would be the regression.

One note for calibration: the convention referenced here, `docs/pylancewiki/review/tests-no-partial-asserts.md`, does not exist in this repository - `docs/` contains only `design.md`, `managing-python-projects.md`, `projects-api-reference.md`, and `startup-flow.md`. I took the advice on its merits, but the citation may be carried over from another repo.

src/test/features/inlineScript/diagnostics.unit.test.ts
StellaHuang95: Agreed, and the gap was wider than the single assertion - fixed in f5027d86.

Grepping all of `src/test`, **no test anywhere asserted an exact diagnostic message**; neither new test file even imported `InlineScriptStrings`. So if `messageFor` mapped `unterminated-block` to the multiple-blocks text, the whole suite would still have gone green.

This test now asserts the exact `InlineScriptStrings.unterminatedBlock` and the `source` field, and a new test walks additional codes (`invalid-content-line`, `invalid-block-marker`) through the publisher and pins each to its own localized message, so the code-to-message mapping is covered rather than just its non-emptiness.

This also matches existing practice in the repo - `extensionVersionCheck.unit.test.ts` already asserts `strictEqual(message, InlineScriptStrings.updatePythonExtension)`. Worth flagging though that the convention cited here, `docs/pylancewiki/review/tests-no-partial-asserts.md`, does not exist in this repository; the underlying point stands on its own.

rchiodo:

**Verification:** Isolated verification observed failures that were not classified as caused by this PR\: Environment and test discovery. The relevant tests could not be fully run in the isolated environment; this review is not fully verified.

**Summary:** Verification could not proceed because the disposable container has no Node.js; the discovery command exited 127 with \`node\: not found\`. Consequently, TypeScript compilation and the targeted parser and diagnostics suites were not run. Source inspection found 61 newly added tests across two files, covering parser ranges, malformed metadata, diagnostics publishing, debouncing, and cleanup. Confidence is limited without executable results.

**Test runs:** 1 failed, 3 not run

- :x: **Failed** | unrelated to this PR | <code>Environment and test discovery</code> | <code>node -e &quot;const fs=require(&#x27;fs&#x27;),p=require(&#x27;./package.json&#x27;); console.log(&#x27;node=&#x27;+process.version); console.log(&#x27;scripts=&#x27;+JSON.stringify(p.scripts,null,2)); console.log(&#x27;node_modules=&#x27;+fs.existsSync(&#x27;node_modules&#x27;)); console.log(&#x27;compiled_parser_test=&#x27;+fs.existsSync(&#x27;out/test/common/inlineScript/metadataDiagnostics.unit.test.js&#x27;)); console.log(&#x27;compiled_publisher_test=&#x27;+fs.existsSync(&#x27;out/test/features/inlineScript/diagnostics.unit.test.js&#x27;))&quot; &amp;&amp; printf &#x27;\nChanged files:\n&#x27; &amp;&amp; git diff --name-status HEAD~1...HEAD &amp;&amp; printf &#x27;\nNew test counts:\n&#x27; &amp;&amp; grep -cE &#x27;^[[:space:]]*test\(&#x27; src/test/common/inlineScript/metadataDiagnostics.unit.test.ts src/test/features/inlineScript/diagnostics.unit.test.ts</code>
- :warning: **Not run** | <code>TypeScript test compilation</code> | <code>npm run compile-tests</code>
- :warning: **Not run** | <code>Inline script metadata parser suites</code> | <code>npm run unittest -- --grep &quot;inlineScriptMetadata&quot;</code>
- :warning: **Not run** | <code>Inline script diagnostics publisher suite</code> | <code>npm run unittest -- --grep &quot;^InlineScriptDiagnosticsPublisher&quot;</code>

<details><summary>:x: <code>Environment and test discovery</code> diagnostic output</summary>

<pre>/bin/sh: 1: node: not found
[container exit=127]</pre>
</details>

<details><summary>:warning: <code>TypeScript test compilation</code> diagnostic output</summary>

<pre>Not run because the container lacks Node.js.</pre>
</details>

<details><summary>:warning: <code>Inline script metadata parser suites</code> diagnostic output</summary>

<pre>Not run because Node.js was unavailable and tests could not be compiled.</pre>
</details>

<details><summary>:warning: <code>Inline script diagnostics publisher suite</code> diagnostic output</summary>

<pre>Not run because Node.js was unavailable and tests could not be compiled.</pre>
</details>
</details>

rchiodo (approved):
Approved via Review Center.

PR #1797, opened 2026-09-17
Add API Docs

edvilme:
Added docs for public facing API


bschnurr:

docs/README.md
```diff
@@ -0,0 +1,1780 @@
...
+- Pass URIs so the API can route project and environment operations.
+- Feature-detect optional provider methods before calling them.
+- Dispose event subscriptions and provider registrations.
+- Preserve operational errors and handle documented `undefined` results.
+- Use `isPackageVersionLookupNotSupportedError()` across bundle boundaries.
+- Recompile after updating the npm package so TypeScript detects API changes.
+
+The public API is intended to avoid breaking changes. Check
+[`api/CHANGELOG.md`](../api/CHANGELOG.md) when updating the package.
+
```
bschnurr, line 1807: > **Issue** · Please address or respond

📍 docs/README.md:1770
The comprehensive reference links outward to `api/README.md`, but the npm-facing README has no reciprocal link, leaving package consumers unable to discover this PR’s primary documentation. Add a prominent, publish-safe “Full API reference” link to `api/README.md`.

[verified]

docs/README.md
```diff
@@ -0,0 +1,1780 @@
...
+> [!IMPORTANT]
+> `ResolveEnvironmentContext` is `Uri` only. Even though nearby source comments
+> mention environments, the exported signature does not accept a
+> `PythonEnvironment`. Pass `env.environmentPath` if you have an environment
+> and want it re-resolved.
+
+#### `getEnvironment`
+
+Reads the environment currently selected for a scope.
```
bschnurr, line 428: > **Warning** · Non-blocking recommendation

📍 docs/README.md:423
`getEnvironment()` can return its last-known value after a one-second resolution timeout, then publish the resolved selection through `onDidChangeEnvironment`. Document this fallback and advise consumers that require freshness to observe the event before acting on the result.

[verified]

docs/README.md
```diff
@@ -0,0 +1,1780 @@
...
+const forProject = await api.getEnvironments(projectUri);
+```
+
+#### `refreshEnvironments`
+
+Asks the relevant environment managers to re-discover environments.
+
+```typescript
+refreshEnvironments(scope: RefreshEnvironmentsScope): Promise<void>;
+```
```
bschnurr, line 376: > **Warning** · Non-blocking recommendation

📍 docs/README.md:375
Results are not guaranteed to arrive through `onDidChangeEnvironments` because that provider event is optional and `refreshEnvironments()` does not synthesize one. Tell consumers to call `getEnvironments()` after awaiting refresh, and describe event delivery as provider-dependent.

[verified]

docs/README.md
```diff
@@ -0,0 +1,1780 @@
...
+| `project` | [`PythonProject`](#pythonproject) | No | Project the task belongs to. |
+| `cwd` | `string` | No | Working directory. Defaults to the project directory of the script being run. |
+| `env` | `{ [key: string]: string }` | No | Additional environment variables for the task. |
+
+#### `PythonBackgroundRunOptions`
+
+| Field | Type | Required | Description |
+| --- | --- | --- | --- |
+| `args` | `string[]` | Yes | Arguments passed to the Python executable. |
+| `cwd` | `string` | No | Working directory. Defaults to the project directory of the script being run. |
```
bschnurr: > **Warning** · Non-blocking recommendation

📍 docs/README.md:1247
`runInBackground()` forwards an omitted `cwd` unchanged and has no script or project context from which to infer the documented default. Either implement that default or update both this text and the public TSDoc to explain the inherited process working directory and recommend supplying `cwd`.

[verified]

docs/README.md
edvilme: Added in `507bee2`. `api/README.md` now has a **Full API reference** section linking to `docs/README.md` plus each of the six domain anchors, using absolute `github.com/.../blob/main/` URLs so the links resolve on npmjs.com, not just in-repo. Also linked `api/CHANGELOG.md` from there.

docs/README.md
edvilme: Addressed in `507bee2`. `refreshEnvironments` now tells consumers to call `getEnvironments` once the promise settles, and states explicitly that `onDidChangeEnvironments` is optional on `EnvironmentManager`, is not synthesized by `refreshEnvironments`, and is therefore provider-dependent. The example comments the `getEnvironments` call as the authoritative read.

docs/README.md
edvilme: Good catch - addressed in `507bee2`. `getEnvironment` now carries an IMPORTANT callout explaining that it races resolution against the 1s `GET_ENVIRONMENT_TIMEOUT_MS` and can return the last-known value (possibly `undefined` on a first call) while resolution continues in the background, and that the authoritative value arrives via `onDidChangeEnvironment`. Added a snippet showing the treat-it-as-a-fast-first-guess pattern.

docs/README.md
edvilme: Addressed in `507bee2` - took the `update both` option since this is a docs change. `options.cwd` is forwarded straight to `spawnProcess` (`runInBackground.ts:44`), so the docs now say the process inherits the extension host's working directory when `cwd` is omitted and recommend passing it explicitly; the TSDoc on `PythonBackgroundRunOptions.cwd` in `src/types.ts` was corrected to match. While there, I also corrected `PythonTaskExecutionOptions.cwd` in the docs: `runAsTask` forwards it to `ShellExecution`, so the fallback is the task scope derived from `project`, not the script's directory.

docs/README.md
```diff
@@ -0,0 +1,1817 @@
...
+```
+
+#### `PythonTaskExecutionOptions`
+
+| Field | Type | Required | Description |
+| --- | --- | --- | --- |
+| `name` | `string` | Yes | Name of the task, shown in the task UI. |
+| `args` | `string[]` | Yes | Arguments passed to the Python executable. |
+| `project` | [`PythonProject`](#pythonproject) | No | Project the task belongs to. |
+| `cwd` | `string` | No | Working directory for the task's shell execution. When omitted, VS Code resolves it from the task scope - the workspace folder containing `project`, or the global scope when `project` is not supplied. |
```
bschnurr, line 1267: > **Warning** · Non-blocking recommendation

📍 docs/README.md:1267
This task-scope default conflicts with `PythonTaskExecutionOptions.cwd` TSDoc, which still says the default is the script's project directory. Align the public TSDoc with the documented/runtime behavior so consumers receive one consistent contract.

[verified]

bschnurr (approved):
Approved via Review Center.

bschnurr (approved):

StellaHuang95 (approved):

PR #1801, opened 2026-09-19
fix: skip terminal activation when there is no activate script

### LINKED ISSUE #1775: Terminal activation emits `source <prefix>/bin/activate` for uv-managed toolchains that have no activate script
## Environment

| | |
|---|---|
| Extension | `ms-python.vscode-python-envs` **1.36.0** (also present in **1.37.2026090401**, published 2026-09-04) |
| VS Code | 1.137.0 |
| OS | macOS 15.6 (arm64) |
| Shell | zsh |
| Environment manager | `uv` (plus `ms-python.python:system` for comparison) |

## Steps to reproduce

1. `uv python install 3.13` — installs a toolchain to `~/.local/share/uv/python/cpython-3.13.9-macos-aarch64-none/`.
2. Open a folder in VS Code whose selected interpreter is that toolchain:
   `Ctrl/Cmd+Shift+P` → **Python: Select Interpreter** → **Enter interpreter path…** →
   `~/.local/share/uv/python/cpython-3.13.9-macos-aarch64-none/bin/python`
3. Open an integrated terminal.

## Actual behavior

```
source /Users/<user>/.local/share/uv/python/cpython-3.13.9-macos-aarch64-none/bin/activate
source: no such file or directory: /Users/<user>/.local/share/uv/python/cpython-3.13.9-macos-aarch64-none/bin/activate
```

The toolchain's `bin/` contains `python3.13`, `python`, `pip`, `idle3` — there is no `activate`.
A `uv python install` toolchain is a base interpreter, not a virtual environment, so the failure
is not a broken install: the activation command itself should never have been emitted.

Once persisted, the same command is emitted on **every** new terminal in that folder and
survives restarts — this is not a one-off.

Python Environments log for the same window:

```
[interpreterSelection] deven367: cpython-3.13.9-macos-aarch64-none (3.13.9) (source: autoDiscovery)
Terminal is activated: /Users/<user>/.local/share/uv/python/cpython-3.13.9-macos-aarch64-none/bin/python
```

(The extension also auto-discovered and selected this toolchain for a folder with no
environment at all, with no user selection — same outcome.)

## Expected behavior

No activation command is emitted for an environment that cannot be activated; either the
toolchain is reported as non-activatable (like `system` environments, which correctly emit
nothing), or the activation command is gated on the activation script existing.

## Cause

`getShellActivationCommands(dir)` sets the sh/bash/zsh/gitbash (`and "unknown"`) activation
entries **unconditionally**, while fish/csh/xsh/nu are gated on `pathExists`:

```js
t.getShellActivationCommands = async function (e) {
  const t = new Map(), n = new Map();
  return isWindows()
    ? t.set("unknown", [{ executable: join(e, "activate") }])
    : t.set("unknown", [{ executable: "source", args: [join(e, "activate")] }]),   // no pathExists
  t.set(SH,      [{ executable: "source", args: [join(e, "activate")] }]),         // no pathExists
  t.set(BASH,    [{ executable: "source", args: [join(e, "activate")] }]),         // no pathExists
  t.set(GITBASH, [{ executable: "source", args: [v(join(e, "activate"))] }]),      // no pathExists
  t.set(ZSH,     [{ executable: "source", args: [join(e, "activate")] }]),         // no pathExists
  ...
  await pathExists(join(e, "activate.csh")) && t.set(CSH, ...);                    // gated
  await pathExists(join(e, "activate.fish")) && t.set(FISH, ...);                  // gated
  await pathExists(join(e, "activate.xsh"))  && t.set(XSH, ...);                   // gated
  await pathExists(join(e, "activate.nu"))   && t.set(NU, ...);                    // gated
  return { shellActivation: t, shellDeactivation: n };
};
```

It is called with `dirname(executable)`, i.e. `<prefix>/bin`:

```js
const a = dirname(e.executable),
      { shellActivation: o, shellDeactivation: s } = await getShellActivationCommands(a);
```

So for any prefix without `bin/activate`, a failing command is produced instead of an
environment with no activation. `isActivatableEnvironment()` then returns `true`, because it
only checks that `shellActivation` is non-empty:

```js
t.isActivatableEnvironment = function (e) {
  return !!e.execInfo?.activation || !!e.execInfo?.shellActivation;
};
```

The uv manager registers these toolchains through the venv path — the log shows
`Found venv environment: cpython-3.13.9-macos-aarch64-none (3.13.9)` for a directory that has
no `pyvenv.cfg` and no `activate`. `system` environments, which carry no `shellActivation`,
behave correctly and emit nothing.

## Suggested fix

- Gate the sh/bash/zsh/gitbash/unknown entries on `pathExists(join(dir, "activate"))`, matching
  the existing fish/csh/xsh/nu checks, so a prefix without an activation script yields an
  effectively non-activatable environment.
- Optionally, additionally report `uv python install` toolchains as non-activatable (they have
  no `pyvenv.cfg`), so they are not offered as activatable venv environments.

## Workaround

Pin the folder to the system manager, which emits no activation command:

```json
"python-envs.pythonProjects": [
  { "path": "<folder>", "envManager": "ms-python.python:system" }
]
```

or select a real `.venv` / system interpreter instead of the uv toolchain.

## Additional context

Reproduced by selecting the interpreter through **Python: Select Interpreter → Enter interpreter
path…** in the GUI, and independently by restoring the persisted
`ms-python.vscode-python-envs:venv:WORKSPACE_SELECTED` entry the picker writes.

Unrelated observation from the same session, in case it is useful: activating the extension
wrote `"python-envs.defaultEnvManager": "ms-python.python:venv"` into the workspace
`settings.json` (`~/.vscode/settings.json`) by itself.

Shubham-Padkonde:
Fixes #1775

## Problem

When an interpreter's `bin/` has no `activate` script, `getShellActivationCommands(binDir)` still emitted `source <bin>/activate` for sh, bash, zsh, Git Bash, ksh and the `unknown` fallback. `uv python install` toolchains are an example: they're base interpreters, not virtual environments. Every new terminal then ran a failing command:

```
source: no such file or directory: ~/.local/share/uv/python/cpython-3.13.9-macos-aarch64-none/bin/activate
```

The other shells (csh, fish, xonsh, nu, pwsh, cmd) were already gated on their script existing.

## Fix

- `getShellActivationCommands` only adds the `activate`-based entries (`unknown`, sh, bash, Git Bash, zsh, ksh) when `<binDir>/activate` exists. This matches how the other shells are handled. Real venvs, pipenv and poetry environments always ship `activate`, so they're unaffected.
- `isActivatableEnvironment` now requires a non-empty `shellActivation` map. It used to check only that the map existed, so an environment with no activation commands was still offered for activation (the terminal activate button and command-based auto-activation).

## Tests

- New tests in `utils.getShellActivationCommands.unit.test.ts`:
  - no POSIX or `unknown` activation without an `activate` script, on non-Windows and on Windows;
  - activation is still emitted when the script exists.
- New `src/test/features/common/activation.unit.test.ts` covers `isActivatableEnvironment`, including the empty-map case.
- Two existing tests ("Bash activation does not include Set-ExecutionPolicy" and "Windows unknown shell uses activate…") now create the `activate` file they rely on.
- Unit tests (`tsc` + mocha, Node 22):
  - with the change: 2411 passing;
  - without it: the 3 new tests covering the bug fail.
- `eslint` and `prettier --check` are clean on the changed files.

I did not test this manually in VS Code with a real uv toolchain. The fix is covered by the unit tests above.

This change was written with help from an AI coding assistant (Claude Code). I reviewed and tested it as described above.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


bschnurr:

bschnurr:

**Summary:** \[unavailable\] Container verification could not start and local execution was not authorized for this PR HEAD\: Could not inspect Podman machine &\#x27;pyrx-automation&\#x27;.

**Test runs:** none recorded.
</details>

bschnurr (approved):
Approved via Review Center.

Shubham-Padkonde:
The automated run came back `could-not-verify` for infrastructure reasons ("Could not inspect Podman machine"), with no tests recorded — so here are the results from running it locally, in case that unblocks the review.

**The two suites this PR touches**, on the PR head:

```
  isActivatableEnvironment
    ✔ Environment without activation is not activatable
    ✔ Environment with an empty shell activation map is not activatable
    ✔ Environment with shell activation commands is activatable
    ✔ Environment with an activation command is activatable

  getShellActivationCommands
    ...
    No activation without activation scripts
      ✔ No POSIX shell activation when bin has no activate script (e.g. uv python toolchain)
      ✔ No Windows unknown shell activation when Scripts has no activate script
      ✔ POSIX shells are activated when the activate script exists

  16 passing (97ms)
```

**Full unit suite:**

```
  2411 passing (19s)
  FAILURES=0
```

**The new tests do pin the change.** Reverting only `src/features/common/activation.ts` and `src/managers/common/utils.ts` to the parent commit, leaving the tests in place:

```
  13 passing
  3 failing

  AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
  AssertionError [ERR_ASSERTION]: unknown should not be activated
  AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:

FAILURES=3
```

and restoring them returns it to `16 passing`.

---

One aside that may be worth a separate look, unrelated to this PR: `npm run unittest` does not start at all on Node 26 —

```
node_modules/yargs/yargs:3
const {applyExtends, cjsPlatformShim, Parser, Yargs, processArgv} = require('./build/index.cjs')
ReferenceError: require is not defined in ES module scope
```

That is mocha's CLI loading yargs, before any test runs; `npm run compile-tests` is fine. I worked around it by driving Mocha programmatically rather than through its binary, which is how I produced the output above. It reproduces without my changes, so it is purely a local toolchain thing on newer Node — happy to open a separate issue if it is useful.

---

Disclosure: this change was written by Claude Code (Claude Opus 5) working as my agent, at my direction. The test output quoted above comes from real runs in my local environment; I am accountable for this PR.

StellaHuang95 (approved):
Approved via Review Center.

edvilme (approved):

edvilme:
Good catch! Thanks for working on this @Shubham-Padkonde

Shubham-Padkonde:
The latest Ubuntu E2E retry fails before running tests: vscode-test stops at "Resolving version..." with a network AggregateError [ETIMEDOUT] (job 106894373838). Could a maintainer rerun that job? The rerun API returns 403 requiring repository admin rights. The required-label checks also need maintainer attention. I have not changed the implementation for this download failure. Investigated with Codex assistance.

